PR #8, opened 2015-04-14
Update reactive-streams-tck to latest RC5

ktoso:
We think this is going to be the last RC before Final, so would be awesome to sanity check with all possible implementations (including this one).


ktoso:
Yay :-)

PR #14, opened 2015-06-18
Kafka Producer/Consumer configs

dbuschman7:
I have created Producer and Consumer Builder patterns to allow for more extensible configuration of the Kafka producer and consumer objects needed for our production environments. This code change also isolates the Kafka configuration from the rest of the code. I expect that this PR will draw some discussion on this. 

Motivations on Producer:
1. We use async mode to batch messages from Kafka. This requires the both batch.num.messages and queue.buffering.max.ms to be set to get messages correctly delivered in a timely manner. You could not set the later before. 

2. Kafka now supports snappy compression which we want to use over the default compression codec. Speed over size for us.

3. Allow any additional key/value pairs to be set by users for their custom situations.

Motivations on the Consumer:
1. We needed a way to set the consumer.timeout.ms value to something larger than 1500 which was hardcoded in the code base.

2. In Kafka 0.8.2,x, you can now migrate your offsets from ZooKeeper to Kafka so I added a explicit feature to handle that migration. See the kafkaOffsetsStorage(dualCommit: Boolean) method. 

3. Allow any additional key/value pairs to be set by users for their custom situations.


kciesielski:
How about using case classes with default values for the builder pattern? This is typical way of writing builders in Scala, should make things a bit more immutable.

dbuschman7:
How does this look? I have removed the java.util.Properties until the very end in the build method. Also a new case class is constructed for each call with the data stored in a immutable Scala Map.

I also added some tests for this code. :)

kciesielski:
I like it. Let's merge :)

PR #23, opened 2015-07-22
New api

kciesielski:
This introduces new API for creating producers / consumers.

- It allows defining custom RequestStrategy
- It allows passing additional kafka-specific settings


build.sbt
```diff
@@ -25,7 +25,6 @@ scalacOptions ++= Seq(
   "-encoding", "UTF-8",       // yes, this is 2 args
   "-feature",
   "-unchecked",
-  "-Xfatal-warnings",
```
rintcius: just curious, what's the reason for this change?

src/main/scala/com/softwaremill/react/kafka/ConsumerProps.scala
```diff
@@ -1,7 +1,9 @@
-package kafka.consumer
+package com.softwaremill.react.kafka
 
 import java.util.Properties
-import java.util.UUID
+
+import kafka.consumer.ConsumerConfig
+import kafka.serializer.Decoder
 
 object ConsumerProps {
```
rintcius: using `Props` may be a bit confusing here (because of akka `Props`). Maybe rename to `ConsumerProperties`? (similar for `ProducerProps`)

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
```diff
@@ -23,17 +22,12 @@ private[kafka] class KafkaActorSubscriber[T](
       processElement(element.asInstanceOf[T])
     case ActorSubscriberMessage.OnError(ex) =>
       handleError(ex)
-    case ActorSubscriberMessage.OnComplete =>
+    case ActorSubscriberMessage.OnComplete | "Stop" =>
```
rintcius: Why is `Stop` added?

build.sbt
kciesielski: Original intent is to guard certain level of quality. This comes from a commercial project I've been working on where such set of flags does a good job. Do you see any concerns regarding this flag?

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
kciesielski: To be honest: for tests only. I know that's pretty smelly but I couldn't find a better way to implement an important test (`ReactiveKafkaIntegrationSpec. "close producer, kill actor and log error when in trouble"`). Using `TestActorRef.underlyingActor` couldn't work there because I already had an instance of `ActorRef` which needs to be supervised. Sending `ActorSubscriberMessage.OnComplete` triggers some internal akka stuff that I didn't want to get executed.

src/main/scala/com/softwaremill/react/kafka/ConsumerProps.scala
kciesielski: Good point. I'll update that. Thanks :)

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
```diff
@@ -23,17 +22,12 @@ private[kafka] class KafkaActorSubscriber[T](
...
   }
 
   private def processElement(element: T) = {
-    try {
-      producer.send(encoder.toBytes(element), partitionizer(element))
-    }
-    catch {
-      case e: Exception => handleError(e)
-    }
+    producer.send(props.encoder.toBytes(element), props.partitionizer(element))
```
rintcius: No exception handling anymore? this means exceptions get silently lost now, I think?

Were there any negative effect of 
```
-    catch {		
-      case e: Exception => handleError(e)		
-    }		
```

Probably nicest solution is to make it configurable what to do depending on the type of exception

src/main/scala/com/softwaremill/react/kafka/ReactiveKafka.scala
```diff
@@ -1,66 +1,131 @@
...
+    props: ConsumerProps[T]
+  )(implicit actorSystem: ActorSystem): Publisher[T] = {
+    ActorPublisher[T](consumerActor(props))
   }
 
-  def consumeFromEndAsActor[T](topic: String, groupId: String, decoder: Decoder[T])(implicit actorSystem: ActorSystem): ActorRef = {
-    val props = ConsumerProps(host, zooKeeperHost, topic, groupId).readFromEndOfStream()
-    consumeAsActor(props, decoder)
+  def consumerActor[T](props: ConsumerProps[T])(implicit actorSystem: ActorSystem): ActorRef = {
+    actorSystem.actorOf(consumerActorProps(props).withDispatcher("kafka-publisher-dispatcher"))
```
rintcius: Make the name of the dispatcher also configurable?

src/test/scala/examples/examples.scala
```diff
@@ -0,0 +1,59 @@
...
+object examples {
+
+  def simple(): Unit = {
+
+    import com.softwaremill.react.kafka.ReactiveKafka
+    import com.softwaremill.react.kafka.ProducerProps
+    import com.softwaremill.react.kafka.ConsumerProps
+    import akka.actor.ActorSystem
+    import akka.stream.ActorMaterializer
+    import akka.stream.scaladsl.{Sink, Source}
```
rintcius: move imports to begin of file? (or if you prefer make `handling` consistent with respect to place of imports)

build.sbt
rintcius: Indeed that's a good intent. It's just that you are *deleting* the flag here.

rintcius:
Some nice additions @kciesielski - looking forward to see them in the next version!

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
rintcius: Doesn't need to block this PR, but maybe it's worth discussing it with the Akka guys? Seems like something that shouldn't be necessary.

build.sbt
kciesielski: Right. I wanted to keep `@deprecated` method calls in tests which wouldn't compile with this flag. Ultimately I resigned from this idea so I can bring it back, thanks for pointing this out.

src/test/scala/examples/examples.scala
kciesielski: This is just some code that I keep that way because I can copy-paste it to README.

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
kciesielski: `handleError()` in the semantics of reactive streams API is meant to be called only on fatal issues, where you cannot recover and you need to close your resources. It cannot be called twice. This case means that something bad happened when we tried to obtain an element from Kafka which may be recoverable.
I decided to leave error handling to standard mechanisms of Akka - supervision. Throwing an exception should be handled by actor's supervisor. If there's no supervisor, the default Akka's behavior will be writing to logs and restarting the actor (which has no effect on internal Kafka reference) and letting it work further.

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
kciesielski: That's not an issue of Akka - it's "by design" according to reactive streams specs. `OnComplete` is meant to be an internal message of Akka Streams. When it's sent to a subscriber, one can't send any more elements (and onError will not be called). This is part of the specs, so I needed something custom to be able to close internal resource in order to simulate sending an element when Kafka is not available.

rintcius: Sorry, yeah with Akka I meant akka-streams/the reactive stream spec. And I meant if implementations really have to add this kind of (indeed smelly) things to their production code then it is something worth thinking about/solving in akka-streams (or even the spec).

src/test/scala/examples/examples.scala
rintcius: ok

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
rintcius: Right, using supervisor sounds good. I'm only wondering if it is possible/easy to set it with the current set up. May be worth an example (doesn't need to be in this PR though)

src/main/scala/com/softwaremill/react/kafka/ConsumerProps.scala
markglh: Yeah I'd definitely second that

markglh:
Nothing more to add from me, the changes look good!!

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
kciesielski: It's included in this PR. Check README.md for the "Working with actors" section, it also describes error handling.

jamesmorgan:
:+1: looking good

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
rintcius: ok!

rintcius:
:+1:

src/main/scala/com/softwaremill/react/kafka/KafkaActorSubscriber.scala
hepin1989: the probleam is if the producer is closed,and this  change will cause the actor restart again and again and rethrow the ProducerAlreadClosed exception and restart again.until it until the poisionPill.

PR #24, opened 2015-07-24
Java friendly wrapper for creating Consumer and Producer properties introduced in 0.7.0

jamesmorgan:
I have the need to use `reactive-kafka` in a Java 8 project and would like to add a Java friendly wrapper to create Consumer and Producer properties recently added in version `0.7.0`.

The builders are basic but should provider a cleaner more Java friendly API for constructing both a Consumer and Producer configuration classes. Thoughts?

* [x] Create `Producer` builder
* [x] Create `Consumer` builder
* [x] Remove need to provide default args from Java
* [x] Create tests for both builders
* [x] Update docs
* [x] Ready for review/merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,32 @@ val subscriber = kafka.publish(ProducerProperties(
 Source(publisher).map(_.toUpperCase).to(Sink(subscriber)).run()
 ```
 
+```Java
+private ActorSystem actorSystem = ActorSystem.create("ReactiveKafka")
+private String brokerList = "localhost:9092";
+private String zooKeeperHost = "localhost:2181";
+  
+private ReactiveKafka kafka = new ReactiveKafka(brokerList, zooKeeperHost)
+
+ConsumerProperties<String> cp = new PropertiesBuilder.Consumer()
+        .withStringDecoder(Optional.empty())
+        .withGroupId(groupName)
+        .withTopic(topic)
+        .<String>build();
+
+Publisher<String> publisher = kafka.consume(cp, actorSystem);
+        
+ProducerProperties<String> pp = new PropertiesBuilder.Producer()
+        .withStringEncoder(Optional.empty())
+        .withClientId(groupName)
+        .withTopic(topic)
+        .build();
+
+Subscriber<String> subscriber = kafka.publish(pp, actorSystem);
+
+Source(publisher).map((d) -> d.toUpperCase()).to(Sink(subscriber)).run()
+```
+
 Passing configuration properties to Kafka
 ----
 In order to set your own custom Kafka parameters, you can construct `ConsumerProperties` and `ProducerProperties` using
--- a/README.md
+++ b/README.md
@@ -59,16 +59,15 @@ String brokerList = "localhost:9092";
 
 ReactiveKafka kafka = new ReactiveKafka();
 ActorSystem system = ActorSystem.create("ReactiveKafka");
+ActorMaterializer materializer = ActorMaterializer.create(system);
 
 ConsumerProperties<String> cp =
-        new PropertiesBuilder.Consumer(zooKeeperHost, brokerList, "topic", "groupId")
-                .withStringDecoder()
+        new PropertiesBuilder.Consumer(zooKeeperHost, brokerList, "topic", "groupId", new StringDecoder(null))
                 .build();
 
 Publisher<String> publisher = kafka.consume(cp, system);
 
-ProducerProperties<String> pp = new PropertiesBuilder.Producer(zooKeeperHost, brokerList, "topic", "clientId")
-        .withStringEncoder()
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -51,6 +52,37 @@ val subscriber = kafka.publish(ProducerProperties(
...
+        .withZooKeeperHost(zooKeeperHost)
+        .withBrokerList(brokerList)
+        .withClientId(groupName)
+        .withTopic(topic)
+        .withStringEncoder()
+        .build();
+
+Subscriber<String> subscriber = kafka.publish(pp, actorSystem);
+
+Source(publisher).map((d) -> d.toUpperCase()).to(Sink(subscriber)).run()
```
markglh: Can you replace `((d) -> d.toUpperCase())` with `(String::toUpercase())`?

Suppose it doesn't matter for the docs.

jamesmorgan: I will clean up :+1:

src/main/java/com/softwaremill/react/kafka/PropertiesBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+            return this;
+        }
+
+        public Consumer withDefaultDecoder() {
+            this.decoder = new DefaultDecoder(null);
+            return this;
+        }
+
+        public Consumer withParams(Map<String, String> params) {
+            this.consumerParams = new HashMap<String, String>().$plus$plus(JavaConverters.mapAsScalaMapConverter(params).asScala());
```
markglh: Tried a few different approaches to this, all uglier. This does indeed seem to be the winner.

src/main/java/com/softwaremill/react/kafka/PropertiesBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+            return this;
+        }
+
+        public <C> ConsumerProperties<C> build() {
+            if (brokerList != null && zooKeeperHost != null) {
+                return ConsumerProperties.<C>apply(brokerList, zooKeeperHost, topic, groupId, decoder);
+            }
+            return ConsumerProperties.<C>apply(consumerParams, topic, groupId, decoder);
```
markglh: personally I would add an `else {}` - or remove the curly brackets around the `if` block?

src/test/java/com/softwaremill/react/kafka/JavaConsumerPropertiesTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void HandleKafkaStorage() {
+
+        final ConsumerProperties consumerProperties = new PropertiesBuilder.Consumer()
+                .withZooKeeperHost(zooKeepHost)
+                .withBrokerList(brokerList)
+                .withGroupId(groupId)
+                .withTopic(topic)
+                .withStringDecoder()
+                .build()
+                .readFromEndOfStream()
```
markglh: It might be nicer to move these into the builder - so the builder actually sets these before returning the `ConsumerProperties`.
I'm assuming these can't be set when constructing the properties which is why they're being set afterwards?

markglh: What are your thoughts @kciesielski? 

We can leave as it, change the builder to set them as described above. Or add these params to a Scala constructor?

src/main/java/com/softwaremill/react/kafka/PropertiesBuilder.java
jamesmorgan: Ye looked and this seems the best I could find

src/main/java/com/softwaremill/react/kafka/PropertiesBuilder.java
jamesmorgan: I can change if the @kciesielski wants

kciesielski: That's a minor thing, a matter of taste. Let's keep it.

src/test/java/com/softwaremill/react/kafka/JavaConsumerPropertiesTest.java
kciesielski: These are mandatory, so it's better to reshape the API, maybe pass them to the `Consumer` constructor? I believe that it is the the typical approach in builder-like DSLs. This way you'll get something like:
```Scala
new PropertiesBuilder.Consumer(brokerList, zkHost, groupId, topic, decoder)
  .readFromEndOfStream().build()

jamesmorgan: @kciesielski I have updated with your suggestion about using a constructor

jamesmorgan:
@kciesielski PR comments addressed - in regards to changing the constructor this may be one for the future.

kciesielski:
Terrific, thanks a lot for this PR, @jamesmorgan :)

PR #40, opened 2015-09-17
Add imports to the example

andrioni:
Publisher and Subscriber should be imported for the example to be slightly more copy-and-paste-able.


kciesielski:
Thanks @andrioni !

PR #54, opened 2015-11-21
Upgrading to Akka 2.4

manub:
Akka 2.4.0 has been out for two months and it's binary compatible with 2.3.x. 

A simple change in the build.sbt allows Reactive Kafka to use Akka 2.4.0.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import scalariform.formatter.preferences.{SpacesAroundMultiImports, CompactContr
 
 name := "reactive-kafka"
 
-val akkaVersion = "2.3.14"
+val akkaVersion = "2.4.0"
 val akkaStreamVersion = "1.0"
 val curatorVersion = "2.9.0"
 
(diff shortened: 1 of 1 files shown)


kciesielski:
OK, we are good to go with 2.4.0, thx.

adamw:
I'm not sure about this one, if 2.3 is binary compatible with 2.4, why would we constrain the users of the library and require 2.4? It would make sense if we used some 2.4 features, but otherwise I think it would be better to leave 2.3, no?

manub:
Actually you've got a point. Also Akka 2.4 hasn't got support for Scala
2.10 - I'm not in front of a computer now so I can't check if
reactive-kafka is cross built for 2.10.

Could it be a good idea to have a branch that supports 2.4 instead of
having that on master or you would prefer just users to upgrade Akka at
their own will?

kciesielski:
Right, @adamw has a point. I missed that it doesn't support 2.10 too - that's also important. Akka-streams 2.0 is going to support Akka 2.3 so let's keep a single branch for now.

PR #63, opened 2016-01-18
Add support for specifying multiple topics, topic and partition, or topic/partition/offset combinations in ReactiveKafka consumers

mjkrumlauf:
This pull request should resolve issues 61 and 62.

ReactiveKafka consumers can now specify the above in the following combinations (and in the following precedence):

1.  Topic, partition, and offset for one topic only (matches Kafka API)
2.  Multiple topic and partition pairs (also matches Kafka API)
3.  One or more topics via a comma-delimited string of topics - transformed into a List of topics for Kafka API

These changes have been made so that they are Java-friendly, especially since my current use case calls for utilizing Reactive Kafka from Java, so some familiar Scala idioms have not been utilized in parameters.

Unit tests have not yet been updated, nor have additional examples yet been added to the documentation.

One failing unit test has been fixed.

A typo has been corrected.

Sorry for the unsquashed commits.


mjkrumlauf:
I just remembered that default parameters don't work with Java - I will have to do some refactoring.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
```diff
@@ -15,7 +32,18 @@ case class ReactiveKafkaConsumer[K, V](properties: ConsumerProperties[K, V]) {
       properties.keyDeserializer,
       properties.valueDeserializer
     )
-    c.subscribe(List(properties.topic)) // support multiple topics?
+
+    if (topicPartitionOffset.isInstanceOf[TopicPartitionOffset]) {
+      c.seek(new TopicPartition(topicPartitionOffset.topic, topicPartitionOffset.partition), topicPartitionOffset.offset)
```
kciesielski: Can't we call `seek` multiple times, each time for different t+p+o? This way we could only require a `List[TopicPartitionOffset]` in the API.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
```diff
@@ -3,9 +3,26 @@ package com.softwaremill.react.kafka
...
+  *
+  * @param properties
+  * @param topicsAndPartitions
+  * @param topicPartitionOffset
+  * @tparam K
+  * @tparam V
+  */
+case class ReactiveKafkaConsumer[K, V](
+    properties: ConsumerProperties[K, V],
+    topicsAndPartitions: Array[TopicPartitionPair] = Array(),
```
kciesielski: A `Seq` would be more appropriate, with `List.empty` as default.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
```diff
@@ -15,7 +32,18 @@ case class ReactiveKafkaConsumer[K, V](properties: ConsumerProperties[K, V]) {
...
+
+    if (topicPartitionOffset.isInstanceOf[TopicPartitionOffset]) {
+      c.seek(new TopicPartition(topicPartitionOffset.topic, topicPartitionOffset.partition), topicPartitionOffset.offset)
+    }
+    else if (Option(topicsAndPartitions).isDefined && topicsAndPartitions.length > 0) {
+      val topicsPartitions = topicsAndPartitions.map(tp => tp.toTopicPartition).toList
+      c.assign(topicsPartitions)
+    }
+    else {
+      val topics = properties.topic.split(",")
```
kciesielski: This requires `topic` to be renamed to `topics` for better clarity (also in the scaladoc of `ConsumerProperties`)

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
```diff
@@ -15,7 +32,18 @@ case class ReactiveKafkaConsumer[K, V](properties: ConsumerProperties[K, V]) {
       properties.keyDeserializer,
       properties.valueDeserializer
     )
-    c.subscribe(List(properties.topic)) // support multiple topics?
+
+    if (topicPartitionOffset.isInstanceOf[TopicPartitionOffset]) {
+      c.seek(new TopicPartition(topicPartitionOffset.topic, topicPartitionOffset.partition), topicPartitionOffset.offset)
+    }
+    else if (Option(topicsAndPartitions).isDefined && topicsAndPartitions.length > 0) {
```
kciesielski: null-checks are discouraged in Scala, you can omit `Option(topicsAndPartitions).isDefined`.

core/src/main/scala/com/softwaremill/react/kafka/TopicPartitionPair.scala
```diff
@@ -0,0 +1,13 @@
+package com.softwaremill.react.kafka
+
+import org.apache.kafka.common.TopicPartition
+
+/**
+ * Exists to keep {@link TopicPartition} from Kafka out of our interfaces.
+ * @param topic
+ * @param partition
+ */
+case class TopicPartitionPair(topic: String, partition: Int) {
```
kciesielski: I understand that you would like to hide direct Kafka API as much as possible, but I think that we shouldn't worry that much about stuff from the kafka-clients.jar. We already use `ConsumerRecord`, for example. This API is meant to be stable.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
mjkrumlauf: How well does `Seq` work with Java, especially since Java doesn't recognize default parameter values?

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
mjkrumlauf: That can be done.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
mjkrumlauf: I'll have to see how this works with my upcoming refactoring, but I will keep that in mind.  Thanks.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafkaConsumer.scala
mjkrumlauf: You're saying `ConsumerProperties.topic` to `ConsumerProperties.topics` - just to make sure I understand correctly.

core/src/main/scala/com/softwaremill/react/kafka/TopicPartitionPair.scala
mjkrumlauf: I will use the Kafka `TopicPartition` then.  Good point.

mjkrumlauf:
I refactored to add topic/partition/offset parameters to several of the consumer-side functions.  I specified "Scala DSL" and "Java DSL" to account for Java-to-Scala collection conversion and for the fact that Java does not allow default parameter values.  Please look these over carefully as the different parameter combinations became a blur after looking at them for awhile.  Thanks for the opportunity of contributing!

core/src/main/scala/com/softwaremill/react/kafka/TopicPartitionOffset.scala
```diff
@@ -0,0 +1,16 @@
...
+ * Used as a parameter to [[ReactiveKafkaConsumer]] when consuming from
+ * a specified offset on a designated partition of a Kafka topic.
+ *
+ * @param topic
+ * @param partition
+ * @param offset
+ */
+case class TopicPartitionOffset(topic: String, partition: Int, offset: Long)
+
+object TopicPartitionOffset {
```
jtjeferreira: this can be removed right?

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafka.scala
```diff
@@ -95,15 +98,47 @@ class ReactiveKafka {
     Sink.fromGraph(new KafkaGraphStageSink(producer))
   }
 
+  // Scala DSL
+  def graphStageSource[K, V](
+    props: ConsumerProperties[K, V],
+    topicsAndPartitions: List[TopicPartition] = List(),
```
jtjeferreira: There is no better way to encode the relation Topic, Partitions, Offset? maybe with Map[Topic, Map[Partition, Offset]]?

core/src/main/scala/com/softwaremill/react/kafka/TopicPartitionOffset.scala
kciesielski: Yup :)

mjkrumlauf: Yes, the object and its constants can be removed as the constants are no longer referenced.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafka.scala
mjkrumlauf: I was just making use of the `TopicPartitionOffset` case class suggested by @kciesielski earlier.

kciesielski: One more thing: I think it's better to keep two separate functions: One that takes a `List[TopicPartition]` and one that takes a `List[TopicPartitionOffset]`.

13h3r:
What about having having `seekToBeginning` and `seekToEnd` from Kafka API?

kciesielski:
@13h3r Do you want to control if you want streaming from start/end? This is supported by https://github.com/softwaremill/reactive-kafka#controlling-consumer-start-offset

13h3r:
@kciesielski As I can see in sources this is an API to work with `auto.offset.reset` property. This property is used only in case of offset absence or being invalid. I want to be able to go to the start or the end of a stream even in case when current offset is valid.

kciesielski:
@13h3r This should work in any case, even when there's last offset set.

13h3r:
@kciesielski could you explain why it works? All I can find in documentation is

> What to do when there is no initial offset in Kafka or if the current offset does not exist any more on the server (e.g. because that data has been deleted):

kciesielski:
@13h3r I understand it as follows: When you create your stream for the first time, there is no "current offset", so you will start reading either from the beginning or from the end, as specified by "auto.offset.reset". If there's an error and you need to reconnect, then you'll start from the last committed offset.

13h3r:
@kciesielski I understand this same as you. But what if I have committed offset and want to seek to beginning or end? I have such situation when I try to join two kafka topics and in case when they are not synchronized (by some business rules) I need to fallback to very start of stream.

kciesielski:
@13h3r I think I understand your case. Does that mean that you want to re-read some elements even if they have been previously committed?

13h3r:
@kciesielski right now I need just to move to begin and to end. But IMO library should provide access also to seek method.

It may look like this

```scala
trait SeekPosition
object End extends SeekPosition
object Start extends SeekPosition
case class Offset(position: Long) extends SeekPosition

val pos: SeekPosition = ???
val consumer: ActorRef = ???

consumer ! pos
```

mjkrumlauf:
- I modified the Java API to refine and make corrections.
- I added some unit tests
- I removed some deprecation warnings in unit tests (replaced junit.framework.Assert with org.junit.Assert)

kciesielski:
@13h3r Ok, but at which point you want to move? During the stream processing, or when you initialize the stream? If it's the first option, then it's going to be difficult, because Kafka Consumer is not thread safe. Currently it's hidden beneath the graph stage logic actor and this provides synchronization. We could work to expose this actor to the outside but I'm not 100% sure that it is exactly what you want :)

13h3r:
@kciesielski sounds reasonable. For me it looks like this:
* connect to topics and partitions
* fetch some metadata and setup offsets
* start stream processing
* graceful shutdown.

kciesielski:
@13h3r So you don't need to force moving offset during stream processing, only to be able to setup offsets on initialization? And your case is that you are reconnecting to a topic where you already committed some offsets before?

kciesielski:
@mjkrumlauf Great :) One last thing are some integration tests to verify if a stream constructed with new DSLs correctly consumes from specified topics/partitions. I can merge and write them myself if you're OK with that.

build.sbt
```diff
@@ -21,6 +21,7 @@ val coreDependencies = Seq(
   "org.slf4j" % "log4j-over-slf4j" % "1.7.12",
   "org.scalatest" %% "scalatest" % "2.2.4" % "test",
   "com.typesafe.scala-logging" %% "scala-logging-slf4j" % "2.1.2",
+  "org.scalaj" %% "scalaj-collection" % "1.6",
```
jtjeferreira: is this really necessary? there is [JavaConversions](http://www.scala-lang.org/files/archive/api/current/index.html#scala.collection.JavaConversions$) and [JavaConverters](http://www.scala-lang.org/files/archive/api/current/index.html#scala.collection.JavaConverters$) in the std library

kciesielski: Good point.

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafka.scala
```diff
@@ -95,15 +96,67 @@ class ReactiveKafka {
...
+  ) = {
+    val consumer = ReactiveKafkaConsumer(props, topicsAndPartitions)
+    Source.fromGraph(new KafkaGraphStageSource(consumer))
+  }
+
+  // Scala DSL
+  def graphStageSource[K, V](
+    props: ConsumerProperties[K, V],
+    topicsAndPartitions: List[TopicPartition],
+    topicPartitionOffsetsMap: Map[TopicPartition, Long]
```
jtjeferreira: Just to be clear with my issue with this:
What happens if I specify topicsAndPartitions = List(TopicPartition("t1",0)) and topicPartitionOffsetsMap = Map(TopicPartition("t2",0), 123). Will it consume from both topics? But maybe this was a typo from my part and I want to only consume from t1...

Other than this. thx for this nice contribution

kciesielski: I maintain that there should be two separate functions: one that takes `topicsAndParitions` and another that takes `topicPartitionOffsetsMap`, just like in the Java API.

13h3r:
@kciesielski yes, you are right

kciesielski:
@13h3r I understand now, thanks for explaining. Can you create a separate GH issue for this?

core/src/main/scala/com/softwaremill/react/kafka/ReactiveKafka.scala
jtjeferreira: I agree that they should be two functions, but one receives only topicsAndPartitions: List[TopicPartition] because we dont care about the offset, and the other should receive topicsAndPartitionsAndOffsets: Map[TopicPartition, offset]

mjkrumlauf: There should be two functions, my mistake.  I am making the correction.  By the way, where ReactiveKafka functions use a List parameter, should that be a Set instead (since we want unique topic+partition[+offset] values)?

kciesielski: @mjkrumlauf Yeah, a `Set` would be better :)

build.sbt
mjkrumlauf: According to the documentation these converters provide the following two-way conversion:
`scala.collection.mutable.Map <=> java.util.Map` but what is needed is at least this one-way conversion: `java.util.Map => scala.collection.Map`, and this does not exist in the std library.

mjkrumlauf: @kciesielski Yes, please add any tests that you think are appropriate.

PR #68, opened 2016-02-01
Fix manual commit bug

13h3r:
Two points:
* wrong state handling
* native consumer close without reason

### DIFF
--- a/core/src/main/scala/com/softwaremill/react/kafka/KafkaGraphStageSource.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka/KafkaGraphStageSource.scala
@@ -86,16 +86,13 @@ class KafkaGraphStageSource[K, V](
         if (offsetMapToFlush.nonEmpty) {
           try {
             consumer.commitSync(offsetMapToFlush.toCommitRequestInfo)
-            committedOffsetMap = OffsetMap(offsetMapToFlush.map)
-            logger.debug(s"committed offsets: $offsetMapToFlush")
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/softwaremill/react/kafka/KafkaGraphStageSource.scala
```diff
@@ -74,29 +74,25 @@ class KafkaGraphStageSource[K, V](
...
-            close()
-            fail(out, ex)
+          case Failure(ex) => fail(out, ex)
         }
       }
 
       def performManualCommit(): Unit = {
-        logger.debug(s"Flushing offsets to commit. Registered offsets: $partitionOffsetMap vs $committedOffsetMap")
-        val offsetMapToFlush = partitionOffsetMap.diff(committedOffsetMap)
+        val partitionOffsetMapCopy = partitionOffsetMap.copy()
```
13h3r: Looks dirty, but it is the simplest solution I found with current shared mutable variable design.

kciesielski:
@13h3r Thanks for the PR and for your valuable comment. 
1. I admit that I wasn't aware of the reusability+materialization idea. An important point indeed. Let's make it a priority issue.
2. The shared mutable state is intentional and I fully realize its flaws. However, that was the only "working" thing that came to my mind for the first implementation. It should work without problems but from the design point of view it certainly needs to be refactored. If you have any ideas or suggestions it would be very appreciated :)

Issue #70, opened 2016-02-03
Make the GraphStage API parts reusable


kciesielski:
The issue has been reported by @13h3r:
The idea of graphs and graph stages are to be completely reusable. You construct you graph with shapes and nothing happens. Then you run your graph and you have an access to "runtime" stuff via materialization process.

GraphStage API in reactive kafka right now is mix of graph and runtime stuff. For example - sourceWithOffsetSink. Creation of ReactiveKafkaConsumer makes graph not reusable and breaks the contract of graphs. All runtime stuff should be created at materialization step.

13h3r:
in master already

Issue #51, opened 2015-11-17
How to store processed log to Kafka or Cassandra with Manual Commit?


evans-ye:
Sorry I can't find a proper place to ask so I just open up an issue here (Close it if you think it's improper).

I'm currently doing a PoC based on this awesome reactive kafka module with the manual commit feature.
And I'm struggling adding a Sink to store log into some permanent storage system such as Kafka or Cassandra. 
From your sample code messages are being processed on-the-fly in processMessage function, if I need to store data into Kafka, then I need to replace offsetCommitSink to another Kafka Sink, but in that way I can use offsetCommitSink to stream back for commit.
Another approach is to use a saveToKafka function to store processed log into Kafka(Shown in below), which is the current implementation of my PoC. 

Source(consumerWithOffsetSink.publisher)
  .map(processMessage(_)) // your message processing
  .map(saveToKafka(_))
  .to(consumerWithOffsetSink.offsetCommitSink) // stream back for commit
  .run() 

Do you think this is the best practice to achieve my goal?

kciesielski:
@evans-ye Have you considered using `Broadcast` from Akka Streams' DSL? It allows "forking" the stream so that you could try to send your processed messages to two "branches": your topic where you save and the commit Sink.
Example: http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-M2/scala/stream-quickstart.html

evans-ye:
Yeah, thanks for your prompt response.
CMIIW, my thought is when using broadcast, there's no guarantee that commit will be executed after data has been saved to my topic, hence there's a chance to lost data.

Let's say the stream saving data to my topic is slower, and the committing stream is faster. 
When failure happened at the point that offsets get committed back, but the other stream saving data to my topic is still under processing, the data under processing is lost because when next time we start to fetch offset, the offset has been set to latter.

jasongilanfarr:
You definitely can do this (and I do) by using a broadcast and some flow stages so you'll commit only after the persist to Cassandra future completes. As a side benefit, this pseudo flow only sends the message itself to your persist method.

Source(publisher) ~> broadcast ~> Flow.map(_.message).mapAsync(1)(persist) ~ zip.in0
broadcast ~> zip.in1
zip.out ~> Flow.map(_.2) ~> commitSink

evans-ye:
OKAY! I got your point. This looks perfect to me.
Thank you so much for the great help and the great module!

I only have one more question, which I think is the benefit using your module:
[...]

PR #72, opened 2016-02-06
[WIP] Alternative graph stages implementation for consumer and producer

13h3r:
Hey!

# Status
With this commit I want to discuss ideas of new API and new graph stage based consumer and producer.

# Ideas
There are few ideas I want to discuss with this commit:
1. New shapes for consumer and producer
2. New implementation of graph stages
3. Kafka consumer loop problem
4. New API for consumer and producer setup

# New shapes
## Design
I implement minimal usable stages around `KafkaConsumer` and `KafkaProducer`. These stages work on API level of kafka client and expose objects from kafka client API. The idea here is to present very basic kafka client API in terms of streams. 

Documentation is here - https://github.com/13h3r/reactive-kafka/blob/wip/api-ideas/docs/NewAPI.md
You can find usage example in `DummyConsumer.scala`.

This shape solves #51.

# Stages implementation
Stages implemented in reusable way. They shares nothing and creates kafka consumer/producer during materialization phase. Created consumer/producer can be easy exposed. Solves #70.

# Consumer event loop problem
I spent a lot of time trying to figure out why new consumer does not call commit callbacks. Here is some details - http://mail-archives.apache.org/mod_mbox/kafka-users/201602.mbox/browser

To solve this I implement some kind of event loop polling in consumer. Check `poll` method which represents one round of polling.

# New API for kafka client setup
I am not a big fan of current API for kafka client setup. I want to separate two parts: client setup and stream stages.

My idea is to provide stages that give you possibility to work with client in stream manner. These stages require producer/consumer that already setup somehow. Yes it will be cool to have typesafe fancy API to setup kafka client, but it should be possible to use all stream stuff with plain kafka client.

That why all stages depends only on producer/consumer.


13h3r:
@kciesielski wrote:

Shutdown handler for the consumer stage. Is it going to be an output where you can handle onError() from the consumer stage? If you want, for example, reset the whole stream on consumer error, should this be coded here?
How about handling producer errors? Will they be propagated to Producer’s confirmation handler output?
Do we need the record2commit itermediary stage? Can’t we just pass the original record there?
Defining consumer properties is independent from defining flows. This means that you can, for example, define an auto-committing consumer but still have a commit input if you create the stage via Consumer.flow, right?

galarragas:
> Shutdown handler for the consumer stage. Is it going to be an output where you can handle onError() from the consumer stage? If you want, for example, reset the whole stream on consumer error, should this be coded here?

Great, I've been spending a lot of time thinking about how to handle this to enable recovering and tracing of termination and failures of AKKA streams

13h3r:
1. What is consumer stage for you? Whole flow or flow between messages and commits? I am thinking about whole flow when working with errors. And at this level you can be notified about error through the handling of `onError` for `confirmation` out. After you get the notification about failure you decide how to react. Restart, shutdown, ignore - it is all up to you. You have the way to be notified.
2. Yes they should be propagated (sure we need to write test for all of this cases and I plan to do it)
3. I don't think so. My point here is to provide very basic API in terms of kafka client. With such design you are free to implement any behavior that depends only on kafka client. For example you can implement publishing in different topic through one publisher, or any custom partition assignment and so on. I don't want to restrict user level API and want to keep it on level of kafka API.
4. Yes it is possible. I am thinking how to enforce some constraints at this level, but have no ideas right now. Also it will work both with autocommit and manual commit through commit `In`. Last commit by time will win.

kciesielski:
Thanks for the answers! :) Yes, consumer stage is whole flow. This looks really good, because error handling or manual stopping has been a painful thing to achieve in the previous API.

core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
```diff
@@ -0,0 +1,203 @@
...
+    val logic = new TimerGraphStageLogic(shape) {
+      case object Poll
+
+      var awaitingConfirmation = 0L
+      var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
+
+      def poll() = {
+        def setupConsumer() = {
+          if (isAvailable(messagesOut)) {
+            consumer.resume(consumer.assignment().toSeq: _*)
```
kciesielski: Why do we need to pause/resume for the whole assignment?

PR #79, opened 2016-02-18
Fixed some typos in New API doc

akauppi:
Also, how about placing images under `images/` (or `.images/`) folder. Customary in projects I've seen, and makes it easy for the visitor to spot the `.md` files.

### DIFF
--- a/docs/NewAPI.md
+++ b/docs/NewAPI.md
@@ -6,16 +6,16 @@ Reactive Streams for Kafka. New API
 ## Core ##
 
 Consumer in new API represents two processes:
- - message emmit
+ - message emit
  - offset commit
 
-The message emmit represented as a `message` `Out` and the offset commit represented as a `commit` `In` and a `confirmation` `Out`.
+The message emit represented as a `message` `Out` and the offset commit represented as a `commit` `In` and a `confirmation` `Out`.
 
 ![Consumer shape](./Consumer.png)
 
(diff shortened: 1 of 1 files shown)


docs/NewAPI.md
```diff
@@ -67,7 +67,7 @@ If you do not care about confirmation you can use producer as `Sink` and create
 
 ![Producer shape](./Producer-sink.png)
 
-To complete producer just complete it `In`.
+To complete producer just complete it `In`. <!-- something missing here?? -->
```
13h3r: I am talking about `In` and `Out` ports of flow.

akauppi: Sorry, I still do not understand that sentence. Could you suggest one here and I'll inject that to the PR - or merge and correct this sentence separately?

13h3r: @akauppi I want to explain how to close producer. Producer is represented as `Flow`. Flow has two ports - `In` port and `Out` port. You publish messages into `In` port and get confirmation from `Out` port. If you want to complete producer you should compete `In` port. It can be done via akka-stream API. I can provide any other clarifications :)

akauppi: Would you be okay with:
>To complete the producer just complete its `In` port.

13h3r: Sure!

Issue #81, opened 2016-02-19
The New API ProducerProvider sample does not compile


akauppi:
The code samples in https://github.com/softwaremill/reactive-kafka/blob/master/docs/NewAPI.md are not included in the project (at core/test/scala/examples), and the `ProducerProvider` sample does not seem to compile.

I was expecting that code + docs would be committed to `master` at the same time - if documents are referring to some sample, it should compile, right? Otherwise following the project becomes really hard. So I presume this is a bug?

...Will provide more information soon...

commit 2086da0074e306cbf6dfe5763fc01a897e6b0ffe

akauppi:
The easier thing to fix is the producer sample: either use `prop` instead of `props` or make the parameters into a tuple:

```
.props("some.props" -> "value")
```

However, it would make sense to keep both consumer and producer samples consistent.

What I don't know is how to get the `provider` parameter for the graph sample.

PR #85, opened 2016-02-19
NewAPI examples from the docs (Note: one of the examples does not compile)

akauppi:
As mentioned in #81, the samples don't compile. This PR should be fixed in a way that they do.

### DIFF
--- a/docs/NewAPI.md
+++ b/docs/NewAPI.md
@@ -105,7 +105,7 @@ val producerProvider = ProducerProvider[Array[Byte], String](
     "localhost:9092",
     new ByteArraySerializer(),
     new StringSerializer()
-).props("some.props", "value")
+).props("some.props" -> "value")
 
 val producer = producerProvider.apply()
 ```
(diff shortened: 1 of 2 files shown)


akauppi:
Got the `.props` stuff corrected.
Now, the NewAPI sample has one line using `.prop(String, String)` and another using `.props(Tuple[String,String])`. This is probably confusing.

Is there really a valid case for having the `.prop` variant (setting just one property), at all, since the `.props` method can be used for 1..n prop changes just as well. As a user, I would suggest removing `.prop` since it only complicates things, adds to the burden of testing (right?) etc. I don't see the need for it.

docs/NewAPI.md
```diff
@@ -101,11 +101,9 @@ val consumer = consumerProvider.apply() //6
...
 Here it is an example of producer provider usage:
 ```scala
-val producerProvider = ProducerProvider[Array[Byte], String](
-    "localhost:9092",
-    new ByteArraySerializer(),
-    new StringSerializer()
-).props("some.props", "value")
+val producerProvider = 
+    ProducerProvider[Array[Byte], String]("localhost:9092", new ByteArraySerializer, new StringSerializer)
+    .props("some.props" -> "value")
```
akauppi: Changed the layout of the sample to match that of the consumer - to show they are similar.

PR #86, opened 2016-02-23
Moving DummyConsumer and DummyProducer away from core / main

akauppi:
The two files are examples, and they don't belong among the `main` packaging. There is already `core/src/test/scala/examples`, so I suggest placing them there.

Note: These samples are Very Good to be around. However, they would be even better with some comments telling why and what is being done.

Also, I have some questions so would like to keep this PR for discussion for now, so no hasty merge, thank you. :)


core/src/test/scala/examples/DummyConsumer.scala
```diff
@@ -1,4 +1,4 @@
-package com.softwaremill.react.kafka2
+package examples
```
akauppi: Is it okay to simplify the package name?

I don't mind it being long, but I think example (and test) package names are best being something else than the package itself.

core/src/test/scala/examples/DummyConsumer.scala
```diff
@@ -9,6 +9,8 @@ import org.apache.kafka.common.serialization.{ByteArrayDeserializer, StringDeser
 
 import scala.concurrent.Await
 import scala.util.Failure
+import com.softwaremill.react.kafka2._
+import scala.concurrent.duration._
 
 object Streams {
```
akauppi: This Streams object is used by both `DummyConsumer` and `DummyProducer`. Should it be in a separate file, or should all the three be in the same file?

core/src/test/scala/examples/DummyConsumer.scala
```diff
@@ -24,6 +26,11 @@ object Streams {
   }
 }
 
+// tbd. What is the purpose of this?
```
akauppi: Short description of what the sample does?

akauppi:
At least one further note: In `DummyProducer`, the `ProducerProvider` is placed to a val called `producer`. I found that slightly weird. In Consumer side the same is called `provider`.  I think this is simply an oversight by the one making the samples, and both would be better off having `provider`, right?

13h3r:
Should we remove `NewAPIExample` after this? Looks like it all suits for the same purpose.

core/src/test/scala/examples/DummyConsumer.scala
13h3r: Looks ok for me

core/src/test/scala/examples/DummyConsumer.scala
13h3r: Lets move it into separate file and keep examples as short as possible

core/src/test/scala/examples/DummyConsumer.scala
13h3r: This example reads from topic `dummy`, processes messages and commit offset into kafka after processing. This provides at least once delivery guarantee.

Also this example show how to perform grace shutdown.

core/src/test/scala/examples/DummyProducer.scala
```diff
@@ -1,10 +1,17 @@
...
+package examples
 
 import akka.actor.ActorSystem
 import akka.stream.scaladsl.Source
 import akka.stream.{ActorMaterializer, ActorMaterializerSettings}
 import org.apache.kafka.common.serialization.{ByteArraySerializer, StringSerializer}
 
+import com.softwaremill.react.kafka2._
+
+// tbd. What is the purpose of this?
```
13h3r: Publish `1 to 10000` into kafka and waiting for publishing confirmation. Also provides an example of correct shutdown

core/src/test/scala/examples/DummyConsumer.scala
```diff
@@ -32,7 +25,11 @@ object DummyConsumer extends App with LazyLogging {
       .withInputBuffer(16, 16)
   )
 
-  val provider = ConsumerProvider("localhost:9092", new ByteArrayDeserializer, new StringDeserializer)
+  val provider = ConsumerProvider(
+    "localhost:9092",
+    new ByteArrayDeserializer,
+    new StringDeserializer
+  )
```
akauppi: The Consumer and Producer had different indentation. Now they look more alike.

core/src/test/scala/examples/Snippets.scala
```diff
@@ -10,7 +10,7 @@ import scala.language.postfixOps
 /**
  * Code samples for the documentation.
  */
-object examples {
+object Snippets {
```
akauppi: renamed the file and object to avoid `examples.examples`

core/src/test/scala/examples/package.scala
```diff
@@ -0,0 +1,20 @@
+import akka.actor.ActorSystem
+import akka.stream.scaladsl.Sink
+
+import scala.util.Failure
+
+// Tools available for the examples
+//
+package object examples {
```
akauppi: Decided to place this in the package object (kind of trashbag for anything to be seen in a certain package - avoids needing to name an object for it).

akauppi:
I believe the PR would now be ready for merge, according to the comments I got from @13h3r .

Please test if the code really runs - I'm not 100% sure it did for me. Actually, we could write the instructions on how to run the examples somewhere.

core/src/test/scala/examples/DummyProducer.scala
```diff
@@ -1,10 +1,18 @@
...
 import akka.stream.{ActorMaterializer, ActorMaterializerSettings}
 import org.apache.kafka.common.serialization.{ByteArraySerializer, StringSerializer}
 
+import com.softwaremill.react.kafka2._
+
+// Publishes 1 to 10000 (as strings) into Kafka and waits for publishing confirmation.
+// Also provides an example of correct shutdown.
+//
+// Usage:
+//    sbt core/test:run
```
13h3r: Can not figure out how to run this with `test:run`. Could you explain please?

13h3r:
Please, rebase it on top of master. There are some conflicts right now

core/src/test/scala/examples/DummyProducer.scala
akauppi: Try `sbt core/test:run` at the project main level.

akauppi:
@13h3r It was only a conflict on the removal of `NewAPIExample`. Now fixed.

Note: The PR is not "rebased" in the way that it carries multiple commits. Should I do it some other way (as a single commit). If so, please instruct a bit.

core/src/test/scala/examples/DummyConsumer.scala
kciesielski: :+1:

PR #88, opened 2016-02-23
??? triggers dead-code warning, with 'sbt test' 

akauppi:
Hadn't noticed this in #87 - sorry.

With the code out of the comments, the use of `-Ywarn-dead-code` kicks in with `sbt test` (kind of surprising, I would suppose `???` not to be treated as dead code). Anyways, it may be best to do other stuff and keep the graph in comments.

Or remove the whole file, as unneeded?

### DIFF
--- a/core/src/test/scala/examples/NewAPIExample.scala
+++ b/core/src/test/scala/examples/NewAPIExample.scala
@@ -13,7 +13,9 @@ object NewAPIExample {
 
   //--- Graph ---
   {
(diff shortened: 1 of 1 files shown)


13h3r:
Do we still need this example after moving DummyProducer and DummyConsumer into examples?

akauppi:
Nope. Removing the file in #86 so closing this PR.

PR #92, opened 2016-02-25
brokerList -> bootstrapServers in the samples (README) where code wouldn't compile

akauppi:
It seems there's been a change of names from `brokerList` (which I personally find very nice) to `bootstrapServers` (which seems to be the Kafka property name). Currently, the code and comments have both.

The problem is that the three cases this PR targets wouldn't compile, if taken into code.

As a larger point, I feel having code samples in text in the README may be a bad idea. They don't get tested by test suites.


kciesielski:
Thanks. That's indeed a bit painful. Maybe something like https://github.com/tpolecat/tut can help

PR #95, opened 2016-02-28
Suggestions for organizing the core tests

akauppi:
Here are some suggestions on how the tests could be made clearer to newcomers - and more detached from the code itself, which I see valuable. Please judge and discuss.

1. Tests should not be in the same namespace as implementation

This is a guideline I've taken to code I do myself. I would argue that it gives an absolute good (i.e. is not a matter of taste). It forces test code to be more like actual use cases, disallowing "peeks" into the internals of the implementation (one such actually spotted in this PR). Also, conceptually I feel it's clearer for understanding a project.

&nbsp;2. Namespace depth of tests does not matter

Since tests are not being published, they don't need to have unique namespaces. Any project can simply have its tests under `test` (or anything), without conflicts with other projects. This actually helps if working on the command line (IDE usually flattens directory structures to one-step anyways).

&nbsp;3. Would like to have unit tests and integration tests in different package

Those two are very different beasts. Unit tests are usually most important *before* the code is written, to set the bar on when a unit is done. After it is, they serve less purpose.

Integration tests are very important throughout a project, since they measure whether the whole is 'done' or not.

Separating the package names (e.g. `test` for integration and `test.unit` for unit tests) allows detecting references where these two domains mix (i.e. keeping unit tests actually unit tests), and more importantly, running unit tests only, for example with `sbt "testOnly core/*test.unit"`.

In this PR, I haven't filtered pure unit tests away from the tests, yet. All tests are in the `test` package.

&nbsp;4. Separating test tool coding from actual tests

When looking at tests, it's relevant to see what is tested. Therefore only the test code is in the root level of `core/src/test/scala` folder. Tools are placed under a `tools/´ folder. Note: they could probably use the normal `test` package, unlike the PR now has. Placing them in a separate package name has really no added benefit.

&nbsp;5. Unified Spec vs. Test postfix in class names

There may be also some other points of view - I'll comment them on the line wise comments, below.

### DIFF
--- a/core/src/main/scala/com/softwaremill/react/kafka/commit/ConsumerCommitter.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka/commit/ConsumerCommitter.scala
@@ -14,7 +14,9 @@ import scala.language.postfixOps
 /**
  * This should become deprecated as soon as we remove KafkaActorSubscriber/KafkaActorPublisher and related stuff.
  */
-private[commit] class ConsumerCommitter[K, V](
+// Commenting out the package private, to make tests more distant from code (using separate package names). AKa270216
+//private[commit]
+class ConsumerCommitter[K, V](
     consumerActor: ActorRef,
     consumerProperties: ConsumerProperties[_, _]
 ) extends Actor with ActorLogging {
--- a/core/src/test/scala/ConsumerCommitterTest.scala
+++ b/core/src/test/scala/ConsumerCommitterTest.scala
@@ -1,11 +1,12 @@
-package com.softwaremill.react.kafka.commit
+package test
 
 import akka.actor._
 import akka.pattern.ask
 import akka.testkit.{ImplicitSender, TestKit}
 import akka.util.Timeout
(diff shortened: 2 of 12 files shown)


core/src/main/scala/com/softwaremill/react/kafka/commit/ConsumerCommitter.scala
```diff
@@ -14,7 +14,9 @@ import scala.language.postfixOps
 /**
  * This should become deprecated as soon as we remove KafkaActorSubscriber/KafkaActorPublisher and related stuff.
  */
-private[commit] class ConsumerCommitter[K, V](
+// Commenting out the package private, to make tests more distant from code (using separate package names). AKa270216
+//private[commit]
+class ConsumerCommitter[K, V](
```
akauppi: The fact that tests were accessing a package private class probably implies that something should be reviewed within the particular test, or the class.

13h3r:
This is out of date now. Feel free to actualize and reopen

PR #101, opened 2016-03-01
Newapi/producer tests

13h3r:
Tests for new API producer functionality

### DIFF
--- a/core/src/main/scala/com/softwaremill/react/kafka2/Producer.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka2/Producer.scala
@@ -2,7 +2,7 @@ package com.softwaremill.react.kafka2
 
 import java.util.concurrent.TimeUnit
 
-import akka.stream.scaladsl.{Sink, Flow}
+import akka.stream.scaladsl.{Flow, Sink}
 import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue, InHandler, OutHandler}
 import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
 import com.typesafe.scalalogging.slf4j.LazyLogging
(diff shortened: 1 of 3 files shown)


core/src/test/scala/com/softwaremill/react/kafka2/ProducerTest.scala
```diff
@@ -0,0 +1,242 @@
...
+    val result = Mockito.mock(classOf[KafkaProducer[K, V]])
+    Mockito.when(result.send(any[ProducerRecord[K, V]], any[Callback])).thenAnswer(new Answer[java.util.concurrent.Future[RecordMetadata]] {
+      override def answer(invocation: InvocationOnMock) = {
+        val record = invocation.getArguments()(0).asInstanceOf[ProducerRecord[K, V]]
+        val callback = invocation.getArguments()(1).asInstanceOf[Callback]
+        handler(record, callback).onComplete {
+          case Success(value) if !closed => callback.onCompletion(value, null)
+          case Success(value) if closed => callback.onCompletion(null, new Exception("Kafka producer already closed"))
+          case Failure(ex: Exception) => callback.onCompletion(null, ex)
+          case Failure(ex) => ???
```
kciesielski: Is this for the `Throwable` case when Throwable is not an Exception? Maybe this should be called `case Failure(throwableUnsupported)` or something similar to indicate that it's not some unfinished code.

core/src/main/scala/com/softwaremill/react/kafka2/Producer.scala
```diff
@@ -34,32 +34,64 @@ class ProducerSendFlowStage[K, V](producerProvider: () => KafkaProducer[K, V])
...
-        override def onPush(): Unit = {
+        override def onPush() = {
           val msg = grab(messages)
           val result = Promise[(ProducerRecord[K, V], RecordMetadata)]
           producer.send(msg, new Callback {
-            override def onCompletion(metadata: RecordMetadata, exception: Exception): Unit = {
+            override def onCompletion(metadata: RecordMetadata, exception: Exception) = {
               val completion = Option(metadata).map(m => Success((msg, m))).getOrElse(Failure(exception))
               result.complete(completion)
+              decrementConfirmation.invoke(())
```
13h3r: We are talking about kafka's callback, right?
AFAIU callback will be invoked when message will be sent. Callback will be invoked regardless a result of sending - both for success and failure.

core/src/test/scala/com/softwaremill/react/kafka2/ProducerTest.scala
13h3r: done

core/src/test/scala/com/softwaremill/react/kafka2/ProducerTest.scala
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  it should "emit confirmation in same order as inputs" in {
+    val input = 1 to 3 map recordAndMetadata
+
+    val client = {
+      val inputMap = input.toMap
+      new ProducerMock[K, V](ProducerMock.handlers.delayedMap(100 millis)(x => Try{ inputMap(x) }))
+    }
+    val probe = Source
```
13h3r: done

core/src/main/scala/com/softwaremill/react/kafka2/Producer.scala
kciesielski: Yes, I mean Kafka callback. Just wanted to be sure that it won't be lost (timeout?). If we don't call `decrementConfirmation.invoke()` for some reason, we will end up with incosistency.

13h3r: @kciesielski I scouted a kafka client code and found that the callback and the future completion performed right after the client have a result of a batch sending. If something hang batch sending we will be hanged too. 

Any ideas how to check it?

kciesielski: Could you point me to the sources in Kafka code?

13h3r: In `Sender` there is method `completeBatch` which call `batch.done` which completes Future and call `Callback`

13h3r:
Anything else I have to do to merge this?

kciesielski:
@13h3r It's ok, thank you. Let's merge.

PR #104, opened 2016-03-11
just a few minor things from reviewing the code

patriknw:


core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
```diff
@@ -161,11 +168,10 @@ class ManualCommitConsumer[K, V](consumerProvider: ConsumerProvider[K, V])
           logger.trace(s"Start commit {}. Commits in progress {}", toCommit, awaitingConfirmation.toString)
           consumer.commitAsync(toCommit, new OffsetCommitCallback {
             override def onComplete(offsets: util.Map[TopicPartition, OffsetAndMetadata], exception: Exception): Unit = {
-              awaitingConfirmation -= 1
```
ktoso: Thanks for fixing it right away :)

kciesielski:
looks good, thanks :)

core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
13h3r: This callback called from stream which call `consumer.poll`. This shape owns `consumer` and this callback called from this shape stream already.

PR #105, opened 2016-03-11
enable travis

patriknw:
just `test:compile` for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
 - sleep 5
 env:
   matrix:
-    - SCRIPT="sbt test"
+    - SCRIPT="sbt -jvm-opts .travis-jvmopts test"
 # important to use eval, otherwise "&&" is passed as an argument to sbt rather than being processed by bash
 script: eval $SCRIPT
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: scala
+sudo: false
+scala:
+  - "2.11.8"
+jdk:
+  - oraclejdk8
+env:
+  matrix:
+    - SCRIPT="sbt test:compile"
+# important to use eval, otherwise "&&" is passed as an argument to sbt rather than being processed by bash
+script: eval $SCRIPT
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot
+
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: scala
+sudo: false
+scala:
+  - "2.11.7"
```
ktoso: could bump to 2.11.8

patriknw: ok, updated

patriknw:
enabled tests on travis, with get and run of kafka/zookeeper

13h3r:
Should we also add some doc how to use travis build locally?

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+scala:
+  - "2.11.8"
+jdk:
+  - oraclejdk8
+before_install:
+- wget http://apache.mirrors.spacedump.net/kafka/0.9.0.1/kafka_2.11-0.9.0.1.tgz -O kafka.tgz
+- mkdir -p kafka && tar xzf kafka.tgz -C kafka --strip-components 1
+- nohup bash -c "cd kafka && bin/zookeeper-server-start.sh config/zookeeper.properties &"
+- nohup bash -c "cd kafka && bin/kafka-server-start.sh config/server.properties &"
+- sleep 5
```
ktoso: we could put that in a script, call the script from here – then people could use it locally too.

ktoso:
Good point @13h3r though let's go the direction of https://github.com/akka/reactive-kafka/pull/105/files#r55833791 maybe?

.travis.yml
```diff
@@ -0,0 +1,22 @@
+language: scala
+sudo: false
+scala:
+  - "2.11.8"
+jdk:
+  - oraclejdk8
+before_install:
+- wget http://apache.mirrors.spacedump.net/kafka/0.9.0.1/kafka_2.11-0.9.0.1.tgz -O kafka.tgz
+- mkdir -p kafka && tar xzf kafka.tgz -C kafka --strip-components 1
+- nohup bash -c "cd kafka && bin/zookeeper-server-start.sh config/zookeeper.properties &"
```
13h3r: I am not familiar with travis, but if we started it here then where we plan stop it?

ktoso: you don't. the container gets killed.

13h3r: Got it. Thanks

.travis.yml
13h3r: then we also provide stop script too

ktoso: true

PR #109, opened 2016-03-11
add CONTRIBUTING.md

patriknw:
* pretty much a copy of akka-persistence-cassandra


CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
+# Contributing to Akka Reactive Kafka
+
+In case of questions about the contribution process or for discussion of specific issues please visit the [akka/dev gitter chat](https://gitter.im/akka/dev).
```
13h3r: should we use akka/reactive-kafka room in gitter?

patriknw: Good catch, yes we should

PR #111, opened 2016-03-15
Consumer tests

13h3r:
- base tests for consumer
- separate timeouts for data fetching and offset commit
- prevent multiple poll scheduling
- correct cancellation handling for confirmationOut

### DIFF
--- a/core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
@@ -92,9 +92,18 @@ class ManualCommitConsumer[K, V](consumerProvider: () => KafkaConsumer[K, V])
 
     val logic = new TimerGraphStageLogic(shape) {
       private case object Poll
-
+      private val POLL_INTERVAL = 10.millis
+      private val POLL_TIMEOUT = 10.millis
       private var awaitingConfirmation = 0L
       private var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
+      private var pollScheduled = false
+
+      private def schedulePoll() = {
+        if (!pollScheduled) {
+          pollScheduled = true
+          scheduleOnce(Poll, POLL_INTERVAL)
+        }
+      }
 
       private def poll() = {
         def setupConsumer() = {
@@ -125,16 +134,16 @@ class ManualCommitConsumer[K, V](consumerProvider: () => KafkaConsumer[K, V])
         }
         if (needPolling) {
           setupConsumer()
-          val msgs = consumer.poll(100)
+          val msgs = consumer.poll(POLL_TIMEOUT.toMillis)
           handleResult(msgs)
-          if (needPolling) scheduleOnce(Poll, 100.millis)
+          if (needPolling) schedulePoll()
         }
         if (stopping && isClosed(commitIn) && awaitingConfirmation == 0) {
           completeStage()
         }
       }
 
-      val pollCallback = getAsyncCallback[Unit] { _ => poll() }
+      val pollCallback = getAsyncCallback[Unit] { _ => schedulePoll() }
 
       private def pushMsg(msg: ConsumerRecord[K, V]) = {
         logger.trace("Push element {}", msg)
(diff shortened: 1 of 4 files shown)


13h3r:
How can I rerun this?

13h3r:
Could anyone check this? @kciesielski?

13h3r:
OK TO TEST

patriknw:
there is a rebuild button in travis: https://travis-ci.org/akka/reactive-kafka/builds/116046317
I clicked it
otherwise, close and open PR, or push something to the branch

13h3r:
rebased

core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
```diff
@@ -75,7 +89,7 @@ object ManualCommitConsumer {
   }
 }
 
-class ManualCommitConsumer[K, V](consumerProvider: ConsumerProvider[K, V])
+class ManualCommitConsumer[K, V](consumerProvider: () => KafkaConsumer[K, V])
```
patriknw: What do you think about having only the settings for creating the `KafkaConsumer` as parameter, and we would do the `new KafkaKonsumer` based on those settings? The settings would be immutable, and can be shared across materializations, i.e. no need for a factory.

patriknw:
@13h3r I would like to see how the end-user would use this api. Can you add code that illustrates that? It  doesn't have to be a running test, but compiled code (like a documentation sample).

core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
13h3r: There are several point why I chose this API:
- setup of `KafkaCosumer` is not just properties set. It is also offset setup, topic subscriptions and possibly manual partitions assignment. So it can not be setup in terms of just properties. My idea was to build another API to setup `KafkaConsumer`. This API may share reusable immutable parts.
- you can easily share `() => KafkaConsumer ` aswell
- easy mocking / API extension. In case of mocking or performing some setup actions that are out of scope you can use just manual setup like 
```scala
ManualCommitConsumer(() => {
  // just create whatever you want
})
```

13h3r:
@patriknw check https://github.com/akka/reactive-kafka/blob/master/core/src/test/scala/examples/DummyConsumer.scala and https://github.com/akka/reactive-kafka/blob/master/core/src/test/scala/examples/DummyProducer.scala

Also if you are interested in more complex example I can share some real world code

core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
patriknw: thanks for clarifying
`offset` and manual partitions could be part of the settings
mocking of the `KafkaKonsumer` is an interesting point, but I think that should be done by replacing the `Source` itself, instead of mocking the rather involved `KafkaConsumer` api

13h3r: I dont know way to manually assign partition through the properties. Same for offsets for topics/partitions (in case of manual external offset tracking). Could you point me to way to do this?

Mocking `KafkaConsumer` is useful for testing reactive-kafka itself. Like all tests in this commit :) In case of explicit creation of `KafkaConsumer` I dont know how to test all these cases of commit fails, empty poll responses and so.

patriknw:
that is a good example, thanks

13h3r:
@patriknw could we merge these tests and continue to discuss kafka2 API in this ticket?

I just want to have this in master to be able to use these tests in future changes. Next if we want to change existing API will just use another PR.

core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
patriknw: The settings could just carry the offset and partitions information and we call the methods on the `KafkaConsumer` when we create it. The settings doesn't have to be only backed by a `Map`. Some of these settings should perhaps also be populated from config.

perhaps we can keep this for our own testing purposes, but not expose it in end user api?

patriknw:
that's fine, I would prefer some better commit messages (e.g. WIP), next time, thanks

core/src/main/scala/com/softwaremill/react/kafka2/Consumer.scala
13h3r: That was my initial idea: having composable immutable reusable items like `KafkaConsumer => KafkaConsumer` which perform some setup for consumer. 

The problem with this approach is creating initial `KafkaConsumer`, because I want to force some properties will be set (like servers list, serializers and deserializers).

I modeled this with `ConsumerProvider` which should provide type safe API to configure kafka properties (not done yet, but can be implemented). First implementation takes `ConsumerProvider` as parameter, but next I ask a question: should we keep coupling with graph API and consumer setup API at all? And the answer is no, because no coupling is better then coupling :)

Hope I explained all my decision path well.

PR #113, opened 2016-03-17
API discussion

patriknw:
As a complement to the descriptions in https://github.com/akka/akka-meta/issues/13 and https://github.com/akka/akka-meta/issues/14 I have created API stubs for option 1 and 3 with example usage.

Let's discuss!

What other typical usage scenarios are there?

### DIFF
--- a/core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package com.softwaremill.react.kafka_api_opt1
+
+import scala.collection.immutable
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+import akka.NotUsed
+import akka.stream.Graph
+import akka.stream.Inlet
+import akka.stream.Outlet
+import akka.stream.Shape
+import akka.stream.SourceShape
+import akka.stream.scaladsl.Flow
+import akka.stream.scaladsl.GraphDSL
+import akka.stream.scaladsl.Source
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.apache.kafka.clients.consumer.OffsetAndMetadata
+import org.apache.kafka.common.TopicPartition
+
+object Consumer {
+
+  // FIXME those should probably be real case classes (or value classes)?
+  type CommitMsg = Map[TopicPartition, OffsetAndMetadata]
+  type CommitConfirmation = Future[CommitMsg]
+
+  case class ConsumerShape[K, V](
+      commit: Inlet[CommitMsg],
+      messages: Outlet[ConsumerRecord[K, V]],
+      confirmation: Outlet[CommitConfirmation]
+  ) extends Shape {
+    override def inlets: immutable.Seq[Inlet[_]] = immutable.Seq(commit)
+    override def outlets: immutable.Seq[Outlet[_]] = immutable.Seq(messages, confirmation)
+    override def deepCopy(): Shape = ConsumerShape(
+      commit.carbonCopy(),
+      messages.carbonCopy(),
+      confirmation.carbonCopy()
+    )
+    override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape = {
+      require(inlets.size == 1)
+      require(outlets.size == 2)
+      ConsumerShape(
+        inlets.head.as[CommitMsg],
+        outlets(0).as[ConsumerRecord[K, V]],
+        outlets(1).as[CommitConfirmation]
+      )
+    }
+  }
+
+  trait Control {
+    def stop(): Unit
+  }
+
+  def record2commit: Flow[ConsumerRecord[_, _], Map[TopicPartition, OffsetAndMetadata], NotUsed] =
+    Flow[ConsumerRecord[_, _]]
+      .map { msg =>
+        Map(new TopicPartition(msg.topic(), msg.partition()) -> new OffsetAndMetadata(msg.offset()))
+      }
+
+  def batchCommit(maxCount: Int, maxDuration: FiniteDuration): Flow[CommitMsg, CommitMsg, NotUsed] =
+    Flow[Map[TopicPartition, OffsetAndMetadata]]
+      .groupedWithin(maxCount, maxDuration)
+      .map {
+        _.flatten.groupBy(_._1).map {
+          case (topic, offsets) => (topic, offsets.map(_._2).maxBy(_.offset()))
+        }.toMap
+      }
+
+  /**
+   * The full graph stage.
+   */
+  def stage[K, V](consumerProvider: () => KafkaConsumer[K, V]): Graph[ConsumerShape[K, V], Control] = ???
+
+  /**
+   * Plain source, commits immediately, i.e. useful when offset is saved in target store.
+   */
+  def source[K, V](consumerProvider: () => KafkaConsumer[K, V]): Source[ConsumerRecord[K, V], Control] = ???
+
+  /**
+   * Don't care about commit confirmations.
+   */
+  def flow[K, V](consumerProvider: () => KafkaConsumer[K, V]): Flow[CommitMsg, ConsumerRecord[K, V], Control] = ???
+
+  /**
+   * Automatically commit when processor flow is done.
+   */
+  def process[K, V](
+    consumerProvider: () => KafkaConsumer[K, V],
+    processor: Flow[ConsumerRecord[K, V], ConsumerRecord[K, V], NotUsed]
+  ): Source[CommitConfirmation, Control] =
+    Source.fromGraph(GraphDSL.create(stage(consumerProvider)) { implicit b => kafka =>
+      import GraphDSL.Implicits._
+      kafka.messages ~> processor ~> Consumer.record2commit ~> kafka.commit
+      SourceShape(kafka.confirmation)
+    })
+
+}
+
--- a/core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
@@ -1,0 +1,108 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package com.softwaremill.react.kafka_api_opt1
+
+import scala.concurrent.Future
+import akka.Done
+import com.softwaremill.react.kafka_api_producer.Producer
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+import akka.stream.scaladsl.Flow
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import akka.stream.ClosedShape
+import akka.stream.scaladsl.Sink
+import akka.stream.scaladsl.GraphDSL
+import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.RunnableGraph
+import akka.NotUsed
+import akka.stream.scaladsl.Broadcast
+import akka.stream.scaladsl.Merge
+import akka.stream.scaladsl.Unzip
+import akka.stream.scaladsl.Zip
+
+trait Example {
+  implicit val ec = scala.concurrent.ExecutionContext.Implicits.global
+
+  def consumerProvider: () => KafkaConsumer[Array[Byte], String] = ???
+
+  def producerProvider: () => KafkaProducer[Array[Byte], String] = ???
+
+  class DB {
+    def save(offset: Long, data: String): Future[Done] = ???
+  }
+}
+
+// Consume messages and store a representation, including offset in DB
+object Example3a extends Example {
+  val db = new DB
+
+  // this possibly commits the offset before it's saved in db,
+  // but that's not a problem for externally stored offset
+  // (would also work with auto-commit)
+  Consumer.source(consumerProvider)
+    .mapAsync(1) { record =>
+      db.save(record.offset, record.value)
+    }
+
+}
+
+// Consume messages and store a representation, including offset, in DB
+object Example3b extends Example {
+  val db = new DB
+
+  val save = Flow[ConsumerRecord[Array[Byte], String]]
+    .mapAsync(1) { record =>
+      db.save(record.offset, record.value).map(_ => record)
+    }
+  Consumer.process(consumerProvider, save)
+
+}
+
+// Connect a Consumer to Provider
+object Example4a extends Example {
+
+  RunnableGraph.fromGraph(GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
+    import GraphDSL.Implicits._
+
+    val consumer = builder.add(Consumer.flow(consumerProvider))
+    val producer = builder.add(Producer.flow(producerProvider).mapAsync(1)(identity))
+    val unzip = builder.add(Unzip[ConsumerRecord[Array[Byte], String], ProducerRecord[Array[Byte], String]])
+    val zip = builder.add(Zip[ConsumerRecord[Array[Byte], String], ProducerRecord[Array[Byte], String]])
+
+    val f1 = Flow[ConsumerRecord[Array[Byte], String]]
+      .map(r => (r, new ProducerRecord[Array[Byte], String]("topic2", r.value)))
+    consumer ~> f1 ~> unzip.in
+    unzip.out0 ~> zip.in0
+    unzip.out1 ~> producer ~> zip.in1
+    zip.out.map { case (consumerRecord, _) => consumerRecord }.via(Consumer.record2commit) ~> consumer
+
+    // here we don't take care of the commit confirmations, what should we do with them?
+
+    ClosedShape
+  })
+
+}
+
+// Connect a Consumer to Provider
+object Example4b extends Example {
+
+  RunnableGraph.fromGraph(GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
+    import GraphDSL.Implicits._
+
+    val consumer = builder.add(Consumer.flow(consumerProvider))
+    val producer = builder.add(Producer.flow2[Array[Byte], String, ConsumerRecord[Array[Byte], String]](producerProvider).mapAsync(1)(identity))
+
+    val f1 = Flow[ConsumerRecord[Array[Byte], String]]
+      .map(r => (new ProducerRecord[Array[Byte], String]("topic2", r.value), r))
+    consumer ~> f1 ~> producer
+    producer.via(Consumer.record2commit) ~> consumer
+
+    // here we don't take care of the commit confirmations, what should we do with them?
+
+    ClosedShape
+  })
+
+}
--- a/core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package com.softwaremill.react.kafka_api_opt3
+
+import scala.collection.immutable
+import scala.concurrent.duration.FiniteDuration
+import akka.stream.Inlet
+import org.apache.kafka.common.TopicPartition
+import akka.stream.stage.GraphStageWithMaterializedValue
+import org.apache.kafka.clients.consumer.OffsetAndMetadata
+import akka.stream.Outlet
+import org.apache.kafka.clients.consumer.ConsumerRecord
+import akka.stream.SourceShape
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import akka.stream.Shape
+import akka.stream.scaladsl.GraphDSL
+import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.Flow
+import scala.concurrent.Future
+import akka.NotUsed
+import akka.Done
+
+object Consumer {
+
+  trait Message[K, V] {
+    def transaction: Transaction
+    def record: ConsumerRecord[K, V]
+  }
+
+  trait Transaction {
+    def commit(): Future[Done]
+  }
+
+  trait Control {
+    def stop(): Unit
+  }
+
+  def source[K, V](consumerProvider: () => KafkaConsumer[K, V]): Source[Message[K, V], Control] = ???
+
+}
+
--- a/core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package com.softwaremill.react.kafka_api_opt3
+
+import scala.concurrent.Future
+
+import akka.Done
+import com.softwaremill.react.kafka_api_producer.Producer
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+
+trait Example {
+  implicit val ec = scala.concurrent.ExecutionContext.Implicits.global
+
+  def consumerProvider: () => KafkaConsumer[Array[Byte], String] = ???
+
+  def producerProvider: () => KafkaProducer[Array[Byte], String] = ???
+
+  class DB {
+    def save(offset: Long, data: String): Future[Done] = ???
+  }
+}
+
+// Consume messages and store a representation, including offset, in DB
+object Example3 extends Example {
+  val db = new DB
+
+  Consumer.source(consumerProvider)
+    .mapAsync(1) { msg =>
+      db.save(msg.record.offset, msg.record.value).map(_ => msg.transaction.commit())
+    }
+}
+
+// Connect a Consumer to Provider
+object Example4a extends Example {
+  Consumer.source(consumerProvider)
+    .map(msg => (new ProducerRecord[Array[Byte], String]("topic2", msg.record.value), msg.transaction))
+    .to(Producer.sinkWithConsumerTransaction(producerProvider))
+}
+
+// Connect a Consumer to Provider
+object Example4b extends Example {
+  Consumer.source(consumerProvider)
+    .map(msg => (new ProducerRecord[Array[Byte], String]("topic2", msg.record.value), msg.transaction))
+    .via(Producer.flow2(producerProvider))
+    .mapAsync(1)(identity)
+    .mapAsync(1)(transaction => transaction.commit)
+}
--- a/core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package com.softwaremill.react.kafka_api_producer
+
+import scala.concurrent.Future
+import akka.NotUsed
+import akka.stream.scaladsl.Flow
+import akka.stream.scaladsl.Sink
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.clients.producer.RecordMetadata
+import com.softwaremill.react.kafka_api_opt3.Consumer
+
+object Producer {
+
+  def flow[K, V](producerProvider: () => KafkaProducer[K, V]): Flow[ProducerRecord[K, V], Future[ProducerRecord[K, V]], NotUsed] =
+    ???
+
+  def sink[K, V](producerProvider: () => KafkaProducer[K, V]): Sink[ProducerRecord[K, V], NotUsed] =
+    flow(producerProvider).to(Sink.ignore)
+
+  /**
+   * This is for integration with opt3 Consumer
+   */
+  def sinkWithConsumerTransaction[K, V](producerProvider: () => KafkaProducer[K, V]): Sink[(ProducerRecord[K, V], Consumer.Transaction), NotUsed] = ???
+
+  /**
+   * This is an alternative to `flow` and `sinkWithConsumerTransaction`, the `DoneMsg` can carry the `Transaction`.
+   */
+  def flow2[K, V, DoneMsg](producerProvider: () => KafkaProducer[K, V]): Flow[(ProducerRecord[K, V], DoneMsg), Future[DoneMsg], NotUsed] = ???
+}
+
--- a/core/src/main/scala/com/softwaremill/react/kafka_api_producer/Example.scala
+++ b/core/src/main/scala/com/softwaremill/react/kafka_api_producer/Example.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package com.softwaremill.react.kafka_api_producer
+
+import akka.stream.scaladsl.Source
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+
+trait Example {
+  def producerProvider: () => KafkaProducer[Array[Byte], String] = ???
+}
+
+object Example1 extends Example {
+  Source(1 to 10000)
+    .map(_.toString)
+    .map(new ProducerRecord[Array[Byte], String]("topic1", _))
+    .to(Producer.sink(producerProvider))
+}
+
+object Example2 extends Example {
+  Source(1 to 10000)
+    .map(_.toString)
+    .map(new ProducerRecord[Array[Byte], String]("topic1", _))
+    .via(Producer.flow(producerProvider))
+    .mapAsync(1)(identity)
+    .map { record =>
+      println(s"${record.topic}/${record.partition} ${record.value}")
+      record
+    }
+}
--- a/core/src/test/scala/com/softwaremill/react/kafka2/ConsumerTest.scala
+++ b/core/src/test/scala/com/softwaremill/react/kafka2/ConsumerTest.scala
@@ -1,3 +1,7 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
 package com.softwaremill.react.kafka2
 
 import java.util.{Map => JMap}


core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
```diff
@@ -0,0 +1,34 @@
...
+import akka.stream.scaladsl.Flow
+import akka.stream.scaladsl.Sink
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.clients.producer.RecordMetadata
+import com.softwaremill.react.kafka_api_opt3.Consumer
+
+object Producer {
+
+  def flow[K, V](producerProvider: () => KafkaProducer[K, V]): Flow[ProducerRecord[K, V], Future[ProducerRecord[K, V]], NotUsed] =
```
patriknw: The `Future` was discussed in https://github.com/akka/akka-meta/issues/14, but is it really worth it?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
```diff
@@ -0,0 +1,108 @@
...
+    val producer = builder.add(Producer.flow(producerProvider).mapAsync(1)(identity))
+    val unzip = builder.add(Unzip[ConsumerRecord[Array[Byte], String], ProducerRecord[Array[Byte], String]])
+    val zip = builder.add(Zip[ConsumerRecord[Array[Byte], String], ProducerRecord[Array[Byte], String]])
+
+    val f1 = Flow[ConsumerRecord[Array[Byte], String]]
+      .map(r => (r, new ProducerRecord[Array[Byte], String]("topic2", r.value)))
+    consumer ~> f1 ~> unzip.in
+    unzip.out0 ~> zip.in0
+    unzip.out1 ~> producer ~> zip.in1
+    zip.out.map { case (consumerRecord, _) => consumerRecord }.via(Consumer.record2commit) ~> consumer
```
patriknw: connecting one topic to another, with some transformation, should be super simple, and this is far from that

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
```diff
@@ -0,0 +1,108 @@
...
+      .map(r => (new ProducerRecord[Array[Byte], String]("topic2", r.value), r))
+    consumer ~> f1 ~> producer
+    producer.via(Consumer.record2commit) ~> consumer
+
+    // here we don't take care of the commit confirmations, what should we do with them?
+
+    ClosedShape
+  })
+
+}
```
patriknw: one more example of option1: https://github.com/akka/reactive-kafka/blob/master/core/src/test/scala/examples/DummyConsumer.scala

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
```diff
@@ -0,0 +1,108 @@
...
+  // but that's not a problem for externally stored offset
+  // (would also work with auto-commit)
+  Consumer.source(consumerProvider)
+    .mapAsync(1) { record =>
+      db.save(record.offset, record.value)
+    }
+
+}
+
+// Consume messages and store a representation, including offset, in DB
```
13h3r: is this comment correct? if we store offset in DB why should we care about committing offset to kafka?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
```diff
@@ -0,0 +1,108 @@
...
+    val zip = builder.add(Zip[ConsumerRecord[Array[Byte], String], ProducerRecord[Array[Byte], String]])
+
+    val f1 = Flow[ConsumerRecord[Array[Byte], String]]
+      .map(r => (r, new ProducerRecord[Array[Byte], String]("topic2", r.value)))
+    consumer ~> f1 ~> unzip.in
+    unzip.out0 ~> zip.in0
+    unzip.out1 ~> producer ~> zip.in1
+    zip.out.map { case (consumerRecord, _) => consumerRecord }.via(Consumer.record2commit) ~> consumer
+
+    // here we don't take care of the commit confirmations, what should we do with them?
```
13h3r: just use them to know when flow is completed

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
```diff
@@ -0,0 +1,108 @@
...
+
+    ClosedShape
```
13h3r: I prefer to wire confirmation to `Sink.lastOption` and expose it to be able to know when flow is finished

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
```diff
@@ -0,0 +1,51 @@
...
+  class DB {
+    def save(offset: Long, data: String): Future[Done] = ???
+  }
+}
+
+// Consume messages and store a representation, including offset, in DB
+object Example3 extends Example {
+  val db = new DB
+
+  Consumer.source(consumerProvider)
```
13h3r: How to correct shutdown such source? It should send `OnComplete` that indicates that `Source` is completed, but at which point it shutdown underlying client? What if next flow will not call commit at all?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
```diff
@@ -0,0 +1,51 @@
...
+    .mapAsync(1) { msg =>
+      db.save(msg.record.offset, msg.record.value).map(_ => msg.transaction.commit())
+    }
+}
+
+// Connect a Consumer to Provider
+object Example4a extends Example {
+  Consumer.source(consumerProvider)
+    .map(msg => (new ProducerRecord[Array[Byte], String]("topic2", msg.record.value), msg.transaction))
```
13h3r: From my point of view the problem here is hidden by directly creating `new ProducerRecord[Array[Byte], String]`. In my experience you have a `Flow[ConsumerMessage, ProducerMessage]`. Having this flow requires bypassing `transaction` too as in Option1.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
13h3r: Yes, with this API it is boilerplate heavy

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
```diff
@@ -0,0 +1,51 @@
...
+    .to(Producer.sinkWithConsumerTransaction(producerProvider))
+}
+
+// Connect a Consumer to Provider
+object Example4b extends Example {
+  Consumer.source(consumerProvider)
+    .map(msg => (new ProducerRecord[Array[Byte], String]("topic2", msg.record.value), msg.transaction))
+    .via(Producer.flow2(producerProvider))
+    .mapAsync(1)(identity)
+    .mapAsync(1)(transaction => transaction.commit)
```
13h3r: I like that flow looks like flow. Also two `mapAsync` looks a bit strange

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
```diff
@@ -0,0 +1,43 @@
...
+
+object Consumer {
+
+  trait Message[K, V] {
+    def transaction: Transaction
+    def record: ConsumerRecord[K, V]
+  }
+
+  trait Transaction {
+    def commit(): Future[Done]
```
13h3r: Such semantic allows to commit transaction multiple times

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+// Consume messages and store a representation, including offset, in DB
+object Example3 extends Example {
+  val db = new DB
+
+  Consumer.source(consumerProvider)
+    .mapAsync(1) { msg =>
+      db.save(msg.record.offset, msg.record.value).map(_ => msg.transaction.commit())
```
13h3r: Any thought how to setup and perform batch commits with this semantic?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: Any thought how to setup and perform batch commits with this semantic?

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
13h3r: What about exposing `RecordMetadata`?

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * This is for integration with opt3 Consumer
+   */
+  def sinkWithConsumerTransaction[K, V](producerProvider: () => KafkaProducer[K, V]): Sink[(ProducerRecord[K, V], Consumer.Transaction), NotUsed] = ???
+
+  /**
+   * This is an alternative to `flow` and `sinkWithConsumerTransaction`, the `DoneMsg` can carry the `Transaction`.
+   */
+  def flow2[K, V, DoneMsg](producerProvider: () => KafkaProducer[K, V]): Flow[(ProducerRecord[K, V], DoneMsg), Future[DoneMsg], NotUsed] = ???
```
13h3r: Is it just bypassing `DoneMsg` via `Flow[ProducerRecord, Future[RecordMetadata]]`?

13h3r:
I glad to see detailed API without GraphDSL at all. Looks very promising :+1:

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
patriknw: I agree, we should have an api that supports this use case. It should not involve commits at all.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
patriknw: it should be idempotent

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
patriknw: you mean that there is a performance advantage of grouping together a bunch of commits (for different topics/partitions) and send them to Kafka in one call? such thing could be implemented inside the source, I think.

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
patriknw: I probably miss something, but when I looked at `RecordMetadata` it didn't add anything that's not in the `ProducerRecord`. (topic, partition, offset)

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
patriknw: the idea was that you could pass on a message such as the `ConsumerRecord` or the `Transaction` without using unzip/zip

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
drewhk: Is that an issue?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: You can create a commitBatcher stage that you add right after the source. It buffers up Transactions emits its own ones instead. Thus, it intercepts commits and forwards them in a batched fashion to the original source.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
```diff
@@ -0,0 +1,102 @@
...
+object Consumer {
+
+  // FIXME those should probably be real case classes (or value classes)?
+  type CommitMsg = Map[TopicPartition, OffsetAndMetadata]
+  type CommitConfirmation = Future[CommitMsg]
+
+  case class ConsumerShape[K, V](
+      commit: Inlet[CommitMsg],
+      messages: Outlet[ConsumerRecord[K, V]],
+      confirmation: Outlet[CommitConfirmation]
```
drewhk: why is this needed? Can't this be handled by the Consumer internally? Also, if the commit is lost, won't the Consumer replay the message eventually anyway?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  def record2commit: Flow[ConsumerRecord[_, _], Map[TopicPartition, OffsetAndMetadata], NotUsed] =
+    Flow[ConsumerRecord[_, _]]
+      .map { msg =>
+        Map(new TopicPartition(msg.topic(), msg.partition()) -> new OffsetAndMetadata(msg.offset()))
+      }
+
+  def batchCommit(maxCount: Int, maxDuration: FiniteDuration): Flow[CommitMsg, CommitMsg, NotUsed] =
+    Flow[Map[TopicPartition, OffsetAndMetadata]]
+      .groupedWithin(maxCount, maxDuration)
```
drewhk: you can use the `batch` here which uses the backpressure signal instead of a timer to size the batches (and has a maximum, too).

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
```diff
@@ -0,0 +1,108 @@
...
+    def save(offset: Long, data: String): Future[Done] = ???
+  }
+}
+
+// Consume messages and store a representation, including offset in DB
+object Example3a extends Example {
+  val db = new DB
+
+  // this possibly commits the offset before it's saved in db,
+  // but that's not a problem for externally stored offset
```
drewhk: why is that so? I probably miss a detail here.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
drewhk: Composition is important once other transaction-aware connectors are created (JMS, etc)

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
drewhk: that example might want to use mapAsyncUnordered?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
patriknw: You can start a consumer from an offset, and if you store that offset in the target "DB" you can have once-and-only-once. If you don't have such a target you can let Kafka store the offset, and that is what the commit is all about. This documentation is very good: http://kafka.apache.org/090/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
drewhk: And also, this is the same issue with option 1

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: This materializes to `Control`, that can be used to stop it. What should happen if next flow will not commit at all? Is that even fixable?

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
RayRoestenburg: `ProducerRecord` does not have `offset`. `RecordMetadata` is result of producing `ProducerRecord`(s).

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: > Having this flow requires bypassing transaction too as in Option1.

I don't really understand this point.

drewhk:
> Looks like Option 3 can be implemented on top of Option 1. 

The thing is, that even the reverse direction works: you can implement Option 1 over Option 3.

> What if we keep both APIs?

I also think that we will need both style of APIs, but I would opt for using Option 3 first. Also, there will be always use-cases which we don't cover, I would be happy if we could keep complexity low.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
13h3r: That example is just compilable example of API.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
drewhk: But isn't it a possibility that:
 1. message is committed by Consumer.source
 2. DB storage call fails

Or did I misunderstand something here?

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
RayRoestenburg: How are you intending to handle errors by the way, how does the user of the flow know about why it failed, like for instance producing can fail with `MessageSizeTooLargeException`?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: I assume that `Source` manages kafka consumer lifecycle. When I call `stop` what will happens to kafka consumer? We can not stop it because we have to be able to handle their commit and we have no idea when all these messages processed.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: Just try to combine this flow having not `ConsumerMessage => ProducerMessage`, but `Flow[ConsumerMessage, ProducerMessage]`

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
RayRoestenburg: Not sure if this is what you mean @drewhk, but with once-and-only-once you do not commit anything back to Kafka, you re-read from kafka from last known in target "DB". 'processing of a msg' equals write to target "DB".

drewhk: Ah, ok, I see.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: But the topic from which we consume has no natural end-of-stream marker, it might go on forever. I.e. stopping only makes sense if you need to shut down processing and then you have to deal with lost commits anyway.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: I still not see what is wrong here. Why is the transaction bypassed? It is passed along as far as I see.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
RayRoestenburg: @drewhk It would be nice to have a way to say from topic X, partition N, stream from offset A to offset B. That might still take forever (if you don't validate B exists already), but in many cases can stop. And then it's very nice to be able to say, stream from beginning to current end (and what that is) across all partitions of topic.

drewhk: But in that case it is easy, the Source waits until all confirmations (if needed) and then onCompletes the downstream. So I don't see a problem here.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
13h3r: Semantic of offset committing is not only acknowledging current message, but marking all messages before as processed and all messages after as not processed. 

What is semantic of `Message.transaction.commit()`?

Assume we have M1 with offset 1 and M2 with offset 2. What if I call M2.commit and then M1.commit? What if M1 already commited?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: Thanks @drewhk. If we solve problem with kafka consumer lifecycle it will work.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: To provide at most once delivery you have to know that your offset is committed.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: My idea is to provide batch commit which will not overflow in case of high input rate, but will commit every period of time in case of low input rate

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
drewhk: I would say that unless explicitly overridden (with a possible configuration flag) commits should be gapless, i.e. similar semantics to mapAsync:

 - M4 `commit()` called
 - M2 `commit()` called
 - M1 `commit()` called
 - M1 M2 is committed to Kafka
 - M3 `commit()` called
 - M3 M4 committed to Kafka

Btw, how does the current implementation handle this? Also, this is not specific to Option 3, Option 1 uses an explicit channel not a hidden side-channel but the problems are exactly the same.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
13h3r: It still can be composed, since `Consumer.processX(consumerProvider, f1)` is a `Source[CommitConfirmation]`. The only problem (is it problem at all?) it not having one smooth flow, but integration with other stream stuff is still available

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
drewhk: I think `batch` is better than time based:
 - it does not overflow as it has a maximum capacity (just like groupedWithin)
 - it adapts the batch window to the backpressure of downstream. If downstream is very slow, batches will be larger, and if donwstream is fast enough there will be no batching at all.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: > But the topic from which we consume has no natural end-of-stream marker, it might go on forever. I.e. stopping only makes sense if you need to shut down processing and then you have to deal with lost commits anyway.

Why anything should be lost during normal shutdown?

> But in that case it is easy, the Source waits until all confirmations (if needed) and then onCompletes the downstream. So I don't see a problem here.

Not emitting onComplete may dead lock your stream (with splitAfter for example). Also I may never call commit at all (in case of external offset storing)

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
drewhk: I don't understand this scenario, but I haven't used Kafka before. Can you give me a scenario with steps how could this work? I always find strange if an ack is acked, especially if the side that receives the final ack is not the one that can do the recovery (replay).

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: > Why anything should be lost during normal shutdown?

Commits are lost, because the message has been processed, but system has been shut down before Kafka has been notified. This was a reply to "We can not stop it because we have to be able to handle their commit and we have no idea when all these messages processed.". We can change the semantics though that stop just waits until all currently emitted messages have been committed, then stops the Kafka consumer and then emits onComplete.

> Not emitting onComplete may dead lock your stream (with splitAfter for example). 
I don't see any deadlock here. Can you build an example that deadlocks? 

> Also I may never call commit at all (in case of external offset storing)

But then I don't understand your original question. What *should* it do if you never commit? I mean, forget streams for now and just explain to me what you think are the desired semantics in normal non-stream terms so we can be on the same page (I fear some misunderstanding here).

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
13h3r: 1. Idempotent `commit`. Since commit related not only to this message, but to messages before and after it performing commit of message can change processing state of other messages. Should we still treat it as idempotent?

2. Option 1 vs Option 3. Option 1 enforce stable order of commit which from my point of view is more clear. `commit` can be called anywhere (even outside the stream) and also may be called multiple times (with unclear behavior).

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: Assume we have a stream process which is represented as `A => B`. We need at most once delivery. For me implementation will looks like:
- fetch message
- commit offset and *wait for commit confirmation*
- perform `A => B`
- publish to next topic

I am may be wrong in implementation so please correct me if any

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
drewhk: > Idempotent commit. Since commit related not only to this message, but to messages before and after it performing commit of message can change processing state of other messages. Should we still treat it as idempotent?

Idempotent != Independent.  A message M will not be committed unless `.commit()` has been called on it. That message M will be committed irregardless how many `.commit()` calls were invoked after the first.

> Option 1 vs Option 3. Option 1 enforce stable order of commit which from my point of view is more clear

This is not true. There is nothing that enforces ordering in Option 1, the only difference is that there is an explicit channel, but people can (and will) do weird stuff with that. Also, in some cases you will need multiple downstream commits, or routing to different backends depending on message content before committing to Kafka and then there are genuine races independently of the option.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: > But then I don't understand your original question. What should it do if you never commit? I mean, forget streams for now and just explain to me what you think are the desired semantics in normal non-stream terms so we can be on the same page (I fear some misunderstanding here).

I will try to clarify the situation. Are you talking about situation when my process which gets data from kafka is hanged and I want to shutdown? Sure there is no way except just force a shutdown and lost some data.

I was talking about case when your process can be shutdown in reasonable time. Sure I should be ready for hanging, but also I want to have correct predictable stream shutdown if everything goes ok to minimize data lost.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
drewhk: Ok, I see. But I don't see why should we expose this to users, in at-most-once there is already the option of losing stuff, so I don't see why the Source itself cannot handle this internally. In other words, what does it buy to expose this extra confirmation compared to just simply handling it in the source itself?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: Because to use this flow you have to drop transaction and to have it after flow processed you have to bypass it. Or you should have flow like `Flow[ConsumerMessage, (ProducerMessage, Transaction)]`?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: It seems like GH formatted my previous comment wrong, so there is one question missing from me:

> Not emitting onComplete may dead lock your stream (with splitAfter for example). 

I don't see any deadlock here. Can you build an example that deadlocks? The onComplete will be emitted, just after last message has been committed (at least in the scenario I described).

> I was talking about case when your process can be shutdown in reasonable time. Sure I should be ready for hanging, but also I want to have correct predictable stream shutdown if everything goes ok to minimize data lost.

Then I propose the following external `stop()` semantics:
 - `stop()` is called externally
 - Source stops emitting new elements
 - Waits until all messages are committed, or configurable timeout fires
 - on timeout do an onFailure otherwise an onComplete on the stream
 - release the Kafka Consumer

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: I think the Producer Sink should always take a Transaction (maybe it should be in the ProducerMessage) and in the case where you don't care about commits you can pass in something like `Transaction.noop` which does nothing on `commit()`.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
13h3r: > A message M will not be committed unless .commit() has been called on it. 

This is not true.

M2.commit commits M1 even if M1.commit is not called.

> This is not true. There is nothing that enforces ordering in Option 1, the only difference is that there is an explicit channel, but people can (and will) do weird stuff with that. Also, in some cases you will need multiple downstream commits, or routing to different backends depending on message content before committing to Kafka and then there are genuine races independently of the option.

I agreed that this still allows me to made a mistake. Any ideas how to force this?

Also my idea is not committing messages, but committing offsets. This may force user to pay attention that this is not message acknowledging, but offset committing

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: Why should we couple `Producer` and `Transaction` at all? There is nothing about `Transaction` in `Producer` domain.

Yes it can be used together, but there is a lot of cases when `Producer` used with out consumer.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: Because you have to know your offset after restart. So you have to commit your offset. If you use autocommit for example it is still possible to emit message and dont perform commit.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
drewhk: > M2.commit commits M1 even if M1.commit is not called.

With my proposed semantics that I described above, giving an example sequence of events also above, your statement is simply not true. Maybe you refer to Kafka behavior? Because I refer to the intended semantics of the Source, which we are in full control of, i.e. how we schedule commits to Kafka.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
drewhk: But in this case, isn't it just simpler to feed the seen sequnce numbers coming from the Source into that storage where you would store the offset (to be usable in restart) instead? I.e.:

```
source.mapAsync(msg => storeSeqOf(msg).map(_ => msg))
```

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
13h3r: Got you. Yes I was referencing kafka behavior. So with your solution we change semantic: 
- in kafka we operates offsets
- in reactive-kafka we ack messages.

In kafka you can simply drop messages that you not interesting in, but in reactive-kafka you should ack every message. This may be confusing for kafka users. Are you sure we want to move away from kafka semantic?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: Why should I use external storage if kafka offset storage fit to me?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
drewhk: I don't see otherwise working in face of not-strictly ordered acknowledgements. I.e. if someone commits M2 but not yet M1 then it would be just as surprising to see M1 committed in the end. I don't know maybe we can make this configurable?

As for move away from Kafka, we should do it sparingly, but in the end we would like to see an Acknowledgement API that can go into akka-streams and provide an interop layer between various projects that can handle these acknowledged operations. Anyway, we have two options for the semantics of out-of-order commits sketched here already :)

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
drewhk: Because if that is fine for you, then why do you need the commit confirmation explicitly? I understood that you will use the value last returned from there to know where to replay from, but then obviously you need to store it to survive restart. I guess I misunderstood something, can you clarify? (I guess I need some :coffee: too :))

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: > I don't see any deadlock here. Can you build an example that deadlocks? The onComplete will be emitted, just after last message has been committed (at least in the scenario I described).

```scala
//pseudo code
Source
  .splitAfter()
  .map()
  .fold()
  .mergeSubstreams
  .map(/* commit here */)
```

Without emitting `onComplete` `splitAfter` will not stop and will hold some not processed and not committed elements.

Also `sliding` may hold some elements that will never be emitted at all.

> Then I propose the following external stop() semantics:

I dont like put timeout into `Source`. IMO owner of `Source` should decide and implement shutdown process. All `Source` should do is try to shutdown.

This will always fail in case of some messages committed and some no (usual for kafka, because kafka operates offsets, not messages) and in case when messages not committed at all (external offset storing).

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
drewhk: Perhaps it is better to rephrase as questions:
 - What will you do with the commit confirmation data? What action will you perform on receiving it? 
 - What will you do after system restart and how would you use the data returned in the commit confirmation?

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
patriknw: @RayRoestenburg ah, that was the missing piece, thanks. It should of course be part of the output.

Errors, isn't the stream simply completed with failure?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
patriknw: wouldn't it make more sense to have that as a `Future` in the materialized value then, and skip the `confirmations` port?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: > Also sliding may hold some elements that will never be emitted at all.

That is a good point! Same is true for `grouped()`.

So then modified:

- stop() is called externally
- Source stops emitting new elements and calls onComplete
- Waits until all messages are committed, or configurable timeout fires
- release the Kafka Consumer

> I dont like put timeout into Source. IMO owner of Source should decide and implement shutdown process. All Source should do is try to shutdown.

I don't see anything wrong with a timeout. We can allow timeout of zero (terminate now, don't care about missing commits) and very long timeouts (wait as long as every commit comes back even if that takes practically forever).

> and in case when messages not committed at all (external offset storing).

It can be done then with timeout of zero then.

patriknw: I agree with those stop sematics.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: Yes, but eventually (when we unify this kind of API for interop) it might be that the thing that feeds into such a Producer is something like a peristent Actor, or an Http request that will reply only once the thing is committed, etc. So it is the future composability that I am thinking here.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
drewhk: We can even have `.commit(transitive: Boolean = false)` where setting it to true means that a sequence number X automatically commits all other sequence numbers Y < X.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
patriknw: I was toying with two different ideas.

1) `Producer.sinkWithConsumerTransaction`: the producer sink is aware of the concept of a consumer offset `Transaction` and accepts it as input and will commit the transaction when the "send" is completed successfully.

2) `Producer.flow2`: the producer stage is not aware of `Transaction` but makes it easy to pass on a `Transaction` that can be committed as the next step in the flow.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: How else can I be sure that offset is committed? I do not understand another way to achieve this. Could you explain

drewhk: > How else can I be sure that offset is committed?

But what will you do with the fact that you are sure that the offset is committed? And what will you do if you are unsure if the last one was committed or not? (See my questions above)

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: > Yes, but eventually (when we unify this kind of API for interop) it might be that the thing that feeds into such a Producer is something like a peristent Actor, or an Http request that will reply only once the thing is committed, etc. So it is the future composability that I am thinking here.

For me as kafka user it is very natural to work in terms of kafka. Why not keep this API kafka related and provide some mid-tier adapters in case of interop API?

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
13h3r: > As for move away from Kafka, we should do it sparingly, but in the end we would like to see an Acknowledgement API that can go into akka-streams and provide an interop layer between various projects that can handle these acknowledged operations.

I was thinking that you guys have some wider constrains on API design than me and now I confirm it :)

As I said before - what about splitting reactive-kafka API in two levels: 
- Low level close to kafka client API (option 1). Fits good for most kafka users.
- Ack API which is based on top of low level API.

Thanks for discussion, but I need to take a break :)

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: You cannot add support from the outside on the sink to commit `Transaction` objects on successful per-element action, therefore it cannot be done with adapters. The Sink either supports it or not.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
patriknw: Great discussion. So we have two alternatives.

One is described clearly here: https://github.com/akka/reactive-kafka/pull/113/files/e7665c1e513507cf4282fbd86ac951cc8a28ed64#r56641807

The other alternative, if I understand it correctly:
A commit for a (topic, partition, offset) is transitive for all offsets up to that offset and it will be performed immediately. For ordering convenience a commit of an earlier offset than the previous (highest) will be ignored, i.e. no going back.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
patriknw: kafka managed offsets can only be used for at-least-once and at-most-once. I guess the confirmation of the commit is needed for at-most-once, but I have no clue how to wire that up with option-3, while it's trivial in option-1.

patriknw: (if you read previous comment in email there was a typo option-1 <-> option-3)

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
patriknw: but with a flow you can, or?

drewhk: With a Flow you can, of course. And of course you can do the opposite (pseudocode):
```
val flowVersion = Graph {
  in ~> bcast ~> producerSink
  bcast ~> mapAsync { msg => msg.txt.onCommitted /* Returns a future firing on commit */ } ~> out
  FlowShape(in, out)
}
```

drewhk: i.e. make the flow from the non-flow version

core/src/main/scala/com/softwaremill/react/kafka_api_producer/Producer.scala
RayRoestenburg: @patriknw re: errors, that sounds good, never mind, I was looking at some code and mixed some stuff up, time to sleep..

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
ktoso: The latest stop sounds good I think.
Timeout I'd allow but perhaps not enable by default (as attribute perhaps?)

drewhk: > Timeout I'd allow but perhaps not enable by default (as attribute perhaps?)

That I would find strange since that would allow all messages in-flight, written to database, etc. I think it is better to wait a bit for all in-flight messages to report back before pulling the plug.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: It can be done by zipping emitted element with its commit confirmation

13h3r: > What will you do with the commit confirmation data? What action will you perform on receiving it?

start at-most-once operation

> What will you do after system restart and how would you use the data returned in the commit confirmation?

I will not use it at all. Kafka just start sending messages from last comitted offset

drewhk: > start at-most-once operation

What do you mean by that? Can you write pseudocode to explain?

> I will not use it at all. Kafka just start sending messages from last comitted offset

If it is not used, then why is it needed?

13h3r: Should I also explain in details how offset in kafka and consumer works together or you familiar with this?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Example.scala
13h3r: Check discussion about at-least-once and at-most-once delivery in other thread

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: > You cannot add support from the outside on the sink to commit Transaction objects on successful per-element action, therefore it cannot be done with adapters. The Sink either supports it or not.

But you can have an API with kafka semantic and then lift it to your semantic.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
patriknw: >> start at-most-once operation
> What do you mean by that? Can you write pseudocode to explain?

I think he means:
1. receive message
2. commit its offset
3. when commit completed: send rocket to moon

If things crash and you restart you will not receive the message again (and 0 or 1 rocket was used)

drewhk: Well, maybe that helps. But my general confusion is that I see no value in the commit confirmation at all. There are two scenarios:
 - the commit confirmation is not important. In that case why do we need it.
 - the commit confirmation is important. In this case either its contents are important or the fact of its reception. If the contents are important, then they must be persisted otherwise they will be lost during a crash. If the contents are not important but the reception is, then I don't see what you can do if you don't get a confirmation, i.e. then it adds no new layer of safety.

Altogether, I don't see how this commit confirmation helps here, but maybe I miss something. Maybe others can help here @patriknw @rkuhn @ktoso?

drewhk: > 3. when commit completed: send rocket to moon

Ok, but this is where I don't get it: commit does not complete. Silence. Should I send rocket? Was the message committed or can I see it again?

patriknw: Good question. Well, you can't send the rocket. I guess it means that you don't know, and have to restart. I don't know if Kafka has timeouts for such commits, and if it would timeout that would only mean "you don't know if it was committed or not", i.e. you have to restart. In our case that means stream failure.

13h3r: Lets set away streams at all. We have kafka client and we assign a topic/partition to it. The question is - from what partition client should read? There are two solutions: get offset from kafka builtin offset storage or get it from somewhere (like database or so). 

We are interested in builtin mechanism here. So, we want to get offset from kafka and kafka have no idea about this consumer. Then you use predefined behavior: reading from start or from end. Next, when you got message you have to signal kafka builtin offset storage that you done with message. This is what we call here offset committing. So committing offset is just command to kafka to remember your offset. In kafka it is (consumerId, topic, partition) -> offset.

To achieve at-most-once I have to commit offset immediately after fetching it. But commit is asynchronous in kafka client API, then we have to wait confirmation.

Pseudo code will be the same:
// here we have no offset setup at all
fetch message
commit offset and wait for commit confirmation // after confirmation I am sure that after restarting kafka client will automatically set offset to commited.
perform A => B
publish to next topic

13h3r: > Ok, but this is where I don't get it: commit does not complete. Silence. Should I send rocket? Was the message committed or can I see it again?

If it is at-most-once you should not

drewhk: > Good question. Well, you can't send the rocket. I guess it means that you don't know, and have to restart. I don't know if Kafka has timeouts for such commits, and if it would timeout that would only mean "you don't know if it was committed or not", i.e. you have to restart. In our case that means stream failure.

But this can be all handled inside the Source. I still don't see much value in exposing the CommitConfirmation since as a *user* I have no tool in my hand to redeliver the commit. I would rather have this behavior as configuration of the Source and only emitting elements after they have been commit-confirmed (retrying if necessary) and failing the stream if commit-confirmation did not arrive in time.

I guess the basic mismatch is how low-level the API should be. I personally prefer higher level tools since the client is always there for low level access.

13h3r: > I guess the basic mismatch is how low-level the API should be. I personally prefer higher level tools since the client is always there for low level access.

This is key difference between Option 1 and Option 3 :) Option 1 trying to be as close as possible to kafka client API and keep kafka semantic. Option 3 is about creating new level of abstraction on top of kafka client.

Btw, if commit failed you can retry it in Option 1

drewhk: > Btw, if commit failed you can retry it in Option 1

If you make it non-idempotent. Which is possible with Option 3 as well. In fact Option 3 can have `commit()` return a Future[Done] to signal commit-confirmation. 

Don't get me wrong, I am not married to Option 3, but originally I envisioned something like Option 1 for acknowledged streams in general, but after playing around I realized that it invites all kinds of pain for any non-trivial processing chains so I gave up on it and now favor Option 3.

drewhk: But now I understand the at-lest-once delivery scenario :)

drewhk: As an exercise, we can try commit-retry with both options to see how the code looks like.

patriknw: > In fact Option 3 can have commit() return a Future[Done]

option 3 `commit` already has `Future[Done]` as that signal

drewhk: then 
```scala
source.mapAsync { msg => msg.transaction.commit().map(_ => msg) }.via(launchRocketLogic)
```
does what is needed?

patriknw: > This is key difference between Option 1 and Option 3 :) Option 1 trying to be as close as possible to kafka client API and keep kafka semantic. Option 3 is about creating new level of abstraction on top of kafka client.

Yes, we want an api that is easy to use. Personally I don't think option 1 is easy to use, and then I'm probably above average stream user. In what way doesn't option 3 keep the kafka semantics? Yes, we have talked about the ordering of commits and such, but same semantics can be implemented with both options, as far as I see.

13h3r: Patrik, for me coupling commit and messages on API level is a change of semantic. Kafka is pretty low level tool and it operates only offsets, not messages and their commits. That is why in Option 1 input for commit accepts not message, but offsets. You can do transformation, grouping or any other stuff with messages (even drop it and forget) and you still have commit with an offset. This is what kafka client API provides.

On top of this semantic you can build any higher level API/abstraction you needed for your task. We still can provide predefined ways to common cases or even supply our vision high level API.

I like how simple Option 3 (but I have to write one big message about stopping :)), but I worry about introducing new abstraction level on top of existing kafka client API without having access to low level stuff in terms of streams.

patriknw: > `source.mapAsync { msg => msg.transaction.commit().map(_ => msg) }.via(launchRocketLogic)`

yes, exactly, and the at-least-once is

`source.mapAsync { msg => bank.creditMyAccout(msg.record.value.money).map(_ => msg.transaction.commit()) }`

drewhk: > I like how simple Option 3 (but I have to write one big message about stopping :)), but I worry about introducing new abstraction level on top of existing kafka client API without having access to low level stuff in terms of streams.

There I disagree. We have for example a file Sink but it is no way a replacement for generic file APIs. It gives a tool to do the most common file operation with streams: append. There is already a low level API for files, just as there is already a low level API for Kafka: the official client. The reason why one would want a pre-packaged Kafka Akka Streams adapter is to embed it in a larger streaming context, and as such composability is a primary concern. Option 1 composes very badly as it needs awkward boilerplate and weird streams setup.

Just to give a real life example flow inspired by real-world:
 - consume from Kafka queue
 - transform then enrich messages from DB. This can be one-to-many transformation, aka `mapConcat` (i.e. you need can only commit the original message if *all* derivative messages has been processed)
 - route the resulting messages to one or more of a set of backends (DB, JMS, other Kafka topic), ack original message once all these acked back.

Try to build this with mock implementations using Option 1 and Option 3.

patriknw: Thanks for clarifying. Then I understand what you mean. I was only thinking of that as a syntactic difference (because the `commit` method would know the identifiers under the hood), but now I understand what you mean.

rkuhn: Thanks for [the explanation](https://github.com/akka/reactive-kafka/pull/113/files/e7665c1e513507cf4282fbd86ac951cc8a28ed64#r56672329), I finally understand what the commit confirmation is used for: it enables at-most-once semantics to be expressed. What about offering these semantic packages separately, in optimized form? We could offer a Source that encapsulates the immediate commit logic and produces elements at-most-once, and we could offer a Flow that produces messages and consumes commits. These would be like Option 1 but split up by use-case.

Then Option 3 comes in when you need the full power, because with that you can express both an at-least-once part and an at-most-once part, with commit and confirmation in between (modeled using `mapAsync`).

drewhk: > What about offering these semantic packages separately, in optimized form? We could offer a Source that encapsulates the immediate commit logic and produces elements at-most-once, and we could offer a Flow that produces messages and consumes commits. 

@rkuhn what is wrong with the simple one-liners me and @patriknw showed, using the simple `commit()` API (Option 3) returning a `Future[Done]`:

```scala
// At-most-once
source.mapAsync { msg => msg.transaction.commit().map(_ => msg) }.via(launchRocketLogic)

// At-least-once
source.mapAsync { msg => bank.creditMyAccout(msg.record.value.money).map(_ => msg.transaction.commit()) }
```

The issue with Option 1 style processing will be that the only thing that really composes with them is `map()`. Basically everything else like `filter()`, `mapConcat()`, any fan-in/-out will need custom stages and/or graphs to work with that option. I actually recommend to try out some processing patterns in code using both styles.

rkuhn: The at-most-once example you should would be just packaged up for convenience, that is exactly what I meant with the Source.

The at-least-once example you show only works when dealing with a single action that returns a Future, in particular it would be awkward to connect this kind of stream to a KafkaSink. Therefore I was proposing an additional API that exposes this as a Flow that can be combined with a KafkaSink Flow using `join`. Of course this is probably not a frequent use-case, but it can stand in for any number of confirmed destinations that we want to forward some data to.

drewhk: > The at-least-once example you show only works when dealing with a single action that returns a Future,

How so?

>  in particular it would be awkward to connect this kind of stream to a KafkaSink. 

What I proposed in another discussion here that KafkaSink takes an Acknowledgable element, not a raw element and it just calls `commit()` itself.

> Of course this is probably not a frequent use-case, but it can stand in for any number of confirmed destinations that we want to forward some data to.

You can do the same with Opt 3.

drewhk: Anyway, this discussion can go endlessly, I recommend to try out API proposals, preferably those who propose it. I can come up with my simple proposal on Monday.

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
13h3r: I am no sure that such semantic is correct for me. I will explain design principles I followed and ask you to correct me if these principles was wrong.

As an API designer I am sure that graph stages should have their lifecycle rules and be composable in terms of lifecycle. It means for me that the user should have guaranties about how stages will work and handle their resources. Since very first version (0.3) I was sure that such contract looks like this:
- stage should clean up resources on completion
- indication of completion is onComplete or onFailure on all outputs

This contract is very clear and simple and I always follow it. This allows for example easily have endless `Source[Source[Source[T]]]` and be sure about resources usage, because all I have to do is to control number of simultaneous working instances of `Source`. Option 1 was designed in such way - you can easy be notified when resources released. 

With timeout stop semantic is also possible to be notified about stage clean (with `Future` from `stop` for example), but for me it is not so natural as notification via stage completion.

Another possible problem is having _rich_ (I will messages rich if it is not plain data) message which functionality depends on something that is not visible from the current context. Using such kind of messages may introduce class of problems that can not be found at compile time. You may end up with a message that have no chance to be completed and you have no way to be notified that your client is closed.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: > There I disagree. We have for example a file Sink but it is no way a replacement for generic file APIs. It gives a tool to do the most common file operation with streams: append. 

@drewhk brilliant example. Now I agree, that we dont need to reflect all functionality of kafka client. But, we should have a way to combine kafka client API and reactive kafka

drewhk: I was thinking on the weekend, maybe the low level API should be actor based for more flexibility and build the stream one on top of it?

13h3r: Was thinking about the same idea. The idea of the kafka consumer is event loop. All interactions performed in event loop. If you call commit and dont call poll nothing will happens. Also consumer is not thread safe.

The whole event loop idea can be implemented nice with actor and ask pattern. The event loop will be managed by this actor.

patriknw: Sounds interesting. Do we then agree that we should go forward with Option 3 as the "high level" API and explore an appropriate low level api?

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
atamborrino: I second @13h3r on a pue time-based commit API. Committing offset should not be at the same rate of receiving message, especially when commiting to Zookeeper with of course has not the same throughput that Kafka. Setting batch commit interval is actually a tradeoff between at-least-once semantic and pressure put on the system receiving the commit. So I don't see why the speed of downstream would dictate the pressure put on the system receiving the commit, these are separate things.

Also, with a pure time-based I dont think there can be a overflow issue as if we stick with the low-level Kafka API, you just keep the max offsets for each partition. This does not go well with the "commit per msg instead of offset" way of thinking of Option 3, but such a batchCommit could go well with low-level close-to-Kafka Option 1. And I think such things are important for high-performance Kafka use.

drewhk: Ok, makes sense. The streams API is rich enough to express both approaches so I am not sure we need to build this in.

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
13h3r: @patriknw yes, I agree that we need API with 20% complexity to solve 80% tasks. But I am not sure about _rich_ messages with `commit` on a board. Here it is why - https://github.com/akka/reactive-kafka/pull/113#discussion_r56780246.

p.s. Not sure != dont like. I am just not sure and try to find why

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Consumer.scala
patriknw: We discussed this a but during the standup meeting today and one conclusion is that we acknowledge that there two different alternatives as discussed here, but for the Reactive Kafka we should stick to the semantics that Kafka itself provides, i.e. storing an offset position. That means that the user does not have to commit every single message.

drewhk: The short summary is that it is possible to bridge between the two models (offset-based, transitive; message-based, non-transitive) so it might be better to explicitly have both and adapters to provide conversions. We will see how this will work out.

drewhk: (Kafka would need to support only the offset-based one since there would be adapters for the other one)

core/src/main/scala/com/softwaremill/react/kafka_api_opt1/Consumer.scala
patriknw: Thanks for all great thoughts. Regarding _rich_ messages we don't intend to hide the meta data, such as topic, partition, offset. The user can act on that information to decide when to invoke `commitOffset`. The `commitOffset` method would talk to the `Source` via a side-channel (probably impl with Promise).

The reason why we prefer a side channel (`commitOffset` method) over the stream based inlet for commit messages is only because we think it will be too complicated for the user to wire up such graphs.

13h3r: @patriknw, @drewhk, @ktoso, @rkuhn could you also please comment on https://github.com/akka/reactive-kafka/pull/113#discussion_r56780246? Am I right about stage lifecycle contract?

rkuhn: Yes, of course the stage lifecycle must match the Kafka client lifecycle. But this has no influence on the API, the issues are the same, no matter whether you use streams or an async notification backchannel (i.e. there is an inherent problem in shutting down a KafkaSource while waiting for outstanding confirmations—because these might never arrive).

core/src/main/scala/com/softwaremill/react/kafka_api_opt3/Example.scala
drewhk: > stage should clean up resources on completion

Yes, on *stage* completion, but that does not mean that the stage cannot live after all of its ports has been closed!. In fact, how would a Sink, wrapping an API with an async `flush()` method work after its input has been closed (since that is the point where it must flush)? In fact there is a method called `setKeepGoing` which allows the stage to stay alive even after all ports have been closed.

> indication of completion is onComplete or onFailure on all outputs

... and `cancel` for inputs. These signal that no more elements (request) will flow, but has nothing to do with whether the stage has finished or not. In fact, we have always warned people that just because a Sink.file received an onComplete is in no way indicator that the file has been closed, this is why these Sinks materialize to a Future which signals actual completion.

> This allows for example easily have endless Source[Source[Source[T]]] and be sure about resources usage, because all I have to do is to control number of simultaneous working instances of Source

This is already not true even if we don't consider stages. Imagine the following setup (pseudocode):

```scala
val source Graph.create {
   infiniteSource ~> bcast ~> someSink
   bcast ~> take(10) ~> out
   SourceShape(out)
}
```

The above Source will finish its output much earlier than the composite Source actually finishes (which will only happen if `someSink` cancels).

> You may end up with a message that have no chance to be completed and you have no way to be notified that your client is closed.

I don't understand this part. How is this different from the immediate onComplete? The messages which are still flowing downstream while the onComplete starts to propagate further upstream are already non-committable as the Source is no longer alive in your model. This is exactly what is mitigated (up to a timeout) by my proposal.

13h3r: Thanks for the explanation and examples. It was good discussion :)

patriknw:
Thanks for very valueable discussion. Implementation will follow in https://github.com/akka/reactive-kafka/pull/115
More api feedback is of course welcome.

PR #115, opened 2016-03-23
Implementation of new api

patriknw:
So far only stub of scaladsl api based on discussion in https://github.com/akka/reactive-kafka/pull/113
My intention is to continue the implementation here.
More api feedback is of course very welcome.


patriknw:
I have pushed a first working implementation. Much credit to @13h3r, I used his GraphStage implementation as base for this.

I'll add real integration tests and documentation, but please do a first review now.

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
```diff
@@ -0,0 +1,55 @@
...
+      .mapAsync(1)(identity)
+      .mapAsync(1)(_.message.passThrough.commit())
+      .to(Sink.ignore)
+
+  /**
+   * Publish records to Kafka topics and then continue the flow. Possibility to pass through a message, which
+   * can for example be a [[Consumer.CommittableOffset]] or [[Consumer.CommittableOffsetBatch]] that can
+   * be committed later in the flow.
+   */
+  def flow[K, V, PassThrough](producerProvider: () => KafkaProducer[K, V]): Flow[Message[K, V, PassThrough], Future[Result[K, V, PassThrough]], KafkaProducer[K, V]] =
```
patriknw: I still don't see the need for that `Future`. When would you not `mapAsync(1)(identity)` directly afterwards? Why would it be needed to do something else?

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  final case class Result[K, V, PassThrough](
+    offset: Long,
+    message: Message[K, V, PassThrough]
+  )
+
+  /**
+   * Publish records to Kafka topics.
+   */
+  def plainSink[K, V](producerProvider: () => KafkaProducer[K, V]): Sink[ProducerRecord[K, V], KafkaProducer[K, V]] =
```
patriknw: Why do we need the `KafkaProducer` as materialized value?

13h3r: No reason to keep it. It was just unused part of old idea.

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
13h3r: The are two cases:
- you do care about order of messages and uses `mapAsync`
- you do not care about order and uses `mapAsyncUnordered` for better throughput.

core/src/main/scala/akka/kafka/scaladsl/ProducerExample.scala
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.kafka.scaladsl
+
+import akka.stream.scaladsl.Source
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+import akka.NotUsed
+
+trait ProducerExample {
```
13h3r: should we move this file to test or examples folder?

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -0,0 +1,89 @@
...
+   */
+  trait Committable {
+    def commit(): Future[Done]
+  }
+
+  trait CommittableOffset extends Committable {
+    def offset: Offset
+  }
+
+  final case class Offset(topic: String, partition: Int, offset: Long)
```
13h3r: I dont like this name. In terms of kafka offset is just long. And offset applicable to partition and topic

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -0,0 +1,89 @@
...
+import akka.NotUsed
+import akka.dispatch.ExecutionContexts
+import scala.concurrent.ExecutionContext
+import org.apache.kafka.common.TopicPartition
+import org.apache.kafka.clients.consumer.OffsetAndMetadata
+import akka.kafka.internal.ConsumerStage.CommittableOffsetBatchImpl
+
+object Consumer {
+
+  trait Message[K, V] {
```
13h3r: Do we really need this? What about plain consumer message class from kafka?

core/src/main/scala/akka/kafka/scaladsl/ProducerExample.scala
patriknw: absolutely, just forgot

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
patriknw: but for better throughput, shouldn't you do batching in the first place?

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
patriknw: thanks for clarifying

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
patriknw: `OffsetLocation` ?

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
patriknw: this makes it possible to implement `plainSouce` and `committableSouce` with the same `GraphStage` and not do additional copy of objects. We could use `ConsumerRecord` for the `plainSource` but then I think we should have two variants of the `GraphStage`. That should be possible.

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
patriknw: Ignore my previous comment, I was thinking about something else.

Ok, we'll keep the Future

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+      if (stage == null) Future.failed(new IllegalStateException(s"commit not supported, $offset"))
+      else stage.commit(offset)
+  }
+
+  trait Committer {
+    def commit(offset: Consumer.Offset): Future[Done]
+    def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Future[Done]
+  }
+
+  final case class CommittableOffsetBatchImpl(override val offsets: Map[TopicPartition, OffsetAndMetadata])(stage: Option[Committer])
```
drewhk: why an Option here while at the same time CommittableOffsetImpl uses a nullable field?

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  trait Control {
+    def stop(): Future[Done]
+    def stopped: Future[Done]
+  }
+
+  /**
+   * No support for committing offsets to Kafka. Can be used when offset is stored externally or with auto-commit.
+   */
+  def plainSource[K, V](consumerProvider: () => KafkaConsumer[K, V]): Source[Message[K, V], Control] =
```
drewhk: shouldn't these take optionally a start and end offset?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+      private def schedulePoll() = {
+        if (!pollScheduled) {
+          pollScheduled = true
+          scheduleOnce(Poll, POLL_INTERVAL)
+        }
+      }
+
+      private def poll() = {
+        def setupConsumer() = {
+          // TODO do we really need to support change of assignments after startup?
```
drewhk: what is an assignment?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+        def pollTimeout = {
+          val data = Option(POLL_DATA_TIMEOUT).filter(_ => isAvailable(out) && !buffer.hasNext)
+          def commit = Option(POLL_COMMIT_TIMEOUT).filter(_ => awaitingConfirmation > 0)
+          data.orElse(commit)
+        }
+
+        pollTimeout match {
+          case Some(timeout) =>
+            setupConsumer()
+            // TODO poll is blocking, can't we do this without blocking?
```
drewhk: we can do adaptive blocking:
 - poll-spin a fixed amount (let's say 16 iterations, but needs to measure)
 - if poll-spin budget runs out, schedule poll
 - after N unsuccessful poll, block (producer is slow enough that blocking cost is negligible)

Not saying we should, just throwing in the idea.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+            val msgs = consumer.poll(timeout.toMillis)
+            handleResult(msgs)
+            if (pollTimeout.isDefined) schedulePoll()
```
drewhk: it is defined here, since we matched on `Some(timeout)`

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+
+        if (stopping && awaitingConfirmation == 0) {
+          completeStage()
+        }
+      }
+
+      val pollCallback = getAsyncCallback[Unit] { _ => schedulePoll() }
+
+      private def pushMsg(msg: Consumer.CommittableMessage[K, V]): Unit = {
+        logger.trace("Push element {}", msg)
```
drewhk: hmm, do we need this?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+        push(out, msg)
+      }
+
+      setHandler(out, new OutHandler {
+        override def onPull(): Unit = {
+          if (!buffer.hasNext) poll()
```
drewhk: does this mean that the buffer is only filled up once a new pull comes in? I.e.:

1. pull comes in
2. delay until new batch is fetched
3. elements emitted

Can't we detach this and preemptively try to fetch new batches when we are below a low watermark? (OFC then blocking cannot be called until the condition buffer empty && isAvailable(out))

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
drewhk: as I noticed in https://github.com/akka/reactive-kafka/pull/115/files#r57306713 you should actually only call blocking call if in pulled state.

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
patriknw: you set the start offset via `KafkaConsumer`, we will have settings/factory for `KafkaConsumer` and that is such an setting

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+
+      // impl of GraphStage.Committer that is called from the outside via CommittableOffsetBatchImpl
+      override def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Future[Done] = {
+        // not important that we complete with failure during stopping race,
+        // it will be completed with TimeoutException
+        if (stoppedPromise.isCompleted)
+          Future.failed(new IllegalStateException("ConsumerStage is stopped, offset commit not performed"))
+        else {
+          val done = Promise[Done]()
+          val c = materializer.scheduleOnce(COMMIT_TIMEOUT, new Runnable {
```
drewhk: why need the materializer? This is a TimerStage

drewhk: or if this is called from the outside then materializer is not safe to call.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: you can set what topics and partitions that you are interested in, it's possible to change that for a running `KafkaConsumer` but I suggest that we should only support that when creating the stage

core/src/main/scala/akka/kafka/scaladsl/ConsumerExample.scala
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+// Consume messages at-least-once
+object AtLeastOnceExample extends ConsumerExample {
+  val db = new DB
+
+  Consumer.committableSource(consumerProvider)
+    .mapAsync(1) { msg =>
+      db.update(msg.value).map(_ => msg.committableOffset.commit())
```
drewhk: shouldn't that be a `flatMap` (or an andThen)?

core/src/main/scala/akka/kafka/scaladsl/ConsumerExample.scala
```diff
@@ -0,0 +1,107 @@
...
+
+// Consume messages at-least-once, and commit in batches
+object AtLeastOnceWithBatchCommitExample extends ConsumerExample {
+  val db = new DB
+
+  Consumer.committableSource(consumerProvider)
+    .mapAsync(1) { msg =>
+      db.update(msg.value).map(_ => msg.committableOffset)
+    }
+    .batch(max = 10, first => CommittableOffsetBatch.empty.updated(first)) { (batch, elem) =>
```
drewhk: There were comments about people preferring a timer based option (groupedWithin), maybe an example for that is also useful.

drewhk: maybe we should measure these actually

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: yes, we can work on the algorithm, it should be enough to use different/adaptive `poll(timeout)` calls, `poll(0)` is the non-blocking version.

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
drewhk: but the stream will produce them in sequence anyway. I don't yet see the value of that Future.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: yes, this is called from the outside, bummer.
The intention was to schedule a timeout to fail the future to avoid uncompleted futures. It could be a race between stopping and `commitCallback.invoke`. If stopped `invoke` will do nothing.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -58,26 +52,27 @@ class ProducerSendFlowStage[K, V](producerProvider: () => KafkaProducer[K, V])
...
+          val r = Promise[Producer.Result[K, V, P]]
+          producer.send(msg.record, new Callback {
             override def onCompletion(metadata: RecordMetadata, exception: Exception) = {
-              val completion = Option(metadata).map(m => Success((msg, m))).getOrElse(Failure(exception))
-              result.complete(completion)
+              if (exception == null)
+                r.success(Producer.Result(metadata.offset, msg))
+              else
+                r.failure(exception)
               decrementConfirmation.invoke(())
```
drewhk: NotUsed?

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -43,7 +37,7 @@ class ProducerSendFlowStage[K, V](producerProvider: () => KafkaProducer[K, V])
       var completionState: Option[Try[Unit]] = None
 
       def checkForCompletion() = {
```
drewhk: is there an upper time limit on this?

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
drewhk: Is there a failure case? Trying to commit but it failed?

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
drewhk: Ah. Ok, this factory method is a bit confusing. Can't we have a source version that does not expose the factory method but creates a consumer itself?

core/src/main/scala/akka/kafka/scaladsl/ConsumerExample.scala
patriknw: it should indeed be flatMap, thanks

core/src/main/scala/akka/kafka/scaladsl/ConsumerExample.scala
patriknw: yes, good idea

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
patriknw: the failure case is in the `else` case

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
patriknw: you mean a timeout? I hope Kafka has a timeout for the `commitAsync` calls, which means that `awaitingConfirmation` will always be decremented down to 0.
Do you mean an addition timeout for the stage itself, yes we could have that.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
drewhk: ok

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
patriknw: I agree, I will work on that api

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: `handleResult` may push some elements and `isAvailable(out)` may be changed.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: guys, it is here for other reason.

Since kafka consumer is event loop based to get commit confirmation you should call `poll` even if you do not have a demand. So you should mark that you do not interested in new messages, but interested to get commit confirmation

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
13h3r: `PartitionOffset`?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+    def updated(committableOffset: Consumer.CommittableOffset): CommittableOffsetBatch = {
+      val s = stage match {
+        case s @ Some(_) => s
+        case None =>
+          committableOffset match {
+            case c: CommittableOffsetImpl => Option(c.stage)
+            case _ => throw new IllegalArgumentException(
+              s"Unknow CommittableOffset, got [${committableOffset.getClass.getName}], " +
+                s"expected [${classOf[CommittableOffsetImpl].getName}]"
+            )
```
johanandren: Type check in runtime feels weird, but I can't see any obvious way to not do it.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: Here I had "real" usage for `Option`, for the the empty instance. For `CommittableOffsetImpl` I used `null` from the test. Not sure it's worth the extra allocation for each message to wrap that in an Option that will never be null in production.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: I have no problem with the `resume` and `pause` calls. I'm concerned about all the garbage that is created by `consumer.assignment().asScala.toSeq: _*` and that's why I would like to grab that at construction time only.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+          logger.debug("Stop producing messages")
+          complete(out)
+        }
+
+        if (stopping && awaitingConfirmation == 0) {
+          completeStage()
+        }
+      }
+
+      val pollCallback = getAsyncCallback[Unit] { _ => schedulePoll() }
```
johanandren: The other vars and vals are private, I guess it doesn't matter much which but it should be consistent?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
johanandren: The name confused me a bit, we are not really "setting up" but toggling between paused and resuming consumption of the partitions.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+          scheduleOnce(Poll, POLL_INTERVAL)
+        }
+      }
+
+      private def poll() = {
+        def setupConsumer() = {
+          // TODO do we really need to support change of assignments after startup?
+          //      Lots of allocations here.
+          if (isAvailable(out)) {
+            consumer.resume(consumer.assignment().asScala.toSeq: _*)
```
johanandren: The docs of resume says "resume partitions that has been paused" but as far as I understand it we might be doing this every timeout, is that not a problem? Shouldn't there be an additional condition here whether it is paused and only then resume?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+      })
+
+      private val decrementConfirmation = getAsyncCallback[Unit] { _ =>
+        awaitingConfirmation -= 1
+        logger.trace(s"Commits in progress {}", awaitingConfirmation.toString)
+      }
+
+      // impl of GraphStage.Committer that is called from the outside via CommittableOffsetImpl
+      override def commit(offset: Consumer.Offset): Future[Done] = {
+        // committed offset should be the next message the application will consume, i.e. +1
```
13h3r: Are sure about +1?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+              Consumer.CommittableMessage(record.key, record.value, committable)
+            }
+            pushMsg(buffer.next())
+          }
+        }
+
+        def pollTimeout = {
+          val data = Option(POLL_DATA_TIMEOUT).filter(_ => isAvailable(out) && !buffer.hasNext)
+          def commit = Option(POLL_COMMIT_TIMEOUT).filter(_ => awaitingConfirmation > 0)
+          data.orElse(commit)
```
johanandren: Could be implemented with much fewer allocations (especially the `Option(CONSTANT)`s), if folded with the pollTimeout pattern match it could probably be allocation free even?

johanandren: Ah, used in more than one place below so I see it cannot be combined with the match block below.

johanandren: Maybe add `()` to emphasize it is looking at mutable state and is not pure?

core/src/main/scala/akka/kafka/scaladsl/ConsumerExample.scala
13h3r: batching is usual pattern. Should we have less verbose way to batch commits?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+    def commit(offset: Consumer.Offset): Future[Done]
+    def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Future[Done]
+  }
+
+  final case class CommittableOffsetBatchImpl(override val offsets: Map[TopicPartition, OffsetAndMetadata])(stage: Option[Committer])
+      extends CommittableOffsetBatch {
+
+    def updated(committableOffset: Consumer.CommittableOffset): CommittableOffsetBatch = {
+      val s = stage match {
+        case s @ Some(_) => s
```
13h3r: what if there are different stages? Like fetching from two different sources, zipping, performing business action and then commit?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+                s"expected [${classOf[CommittableOffsetImpl].getName}]"
+            )
+          }
+      }
+
+      val offset = committableOffset.offset
+      // committed offset should be the next message the application will consume, i.e. +1
+      CommittableOffsetBatchImpl(offsets.updated(
+        new TopicPartition(offset.topic, offset.partition),
+        new OffsetAndMetadata(offset.offset + 1)
```
13h3r: Could we hide this +1 logic (still did not get why should we have it)

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+      ))(s)
+    }
+
+    override def commit(): Future[Done] = {
+      if (offsets.isEmpty)
+        Future.successful(Done)
+      else {
+        stage match {
+          case Some(s) => s.commit(offsets)
+          case None =>
```
13h3r: Not sure how it possible. I dont like having `offsets` and `stage` separately. They are reflect complex state and can be out of sync.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: Should we expose this to user level? Here we have decision that last offset wins. There at least two another strategies: fail in case of last offset is less that previous and biggest offset wins.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+
+  final case class CommittableOffsetBatchImpl(override val offsets: Map[TopicPartition, OffsetAndMetadata])(stage: Option[Committer])
+      extends CommittableOffsetBatch {
+
+    def updated(committableOffset: Consumer.CommittableOffset): CommittableOffsetBatch = {
+      val s = stage match {
+        case s @ Some(_) => s
+        case None =>
+          committableOffset match {
+            case c: CommittableOffsetImpl => Option(c.stage)
```
13h3r: This cast looks ugly :(

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: There is underlying map with "active" stages. There is no network activity with this call

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: I agree, it can improved. But we also should pay attention to how it optimized in kafka client. There is also prefetch algorithm in it. Check sources.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: Is scheduling timeout affects performance here? Is it expensive operation?

13h3r: Are scheduleOnce thread safe?

ktoso: The act of scheduling an event is very cheap – wait-free even.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+          override def onComplete(offsets: util.Map[TopicPartition, OffsetAndMetadata], exception: Exception): Unit = {
+            if (exception == null)
+              done.success(Done)
+            else
+              done.failure(exception)
+            logger.trace(s"Commit completed {} - {}", offsets, done.future.value)
+            decrementConfirmation.invoke(())
+          }
+        })
+        poll()
```
13h3r: This can be optimized to scheduling poll after https://issues.apache.org/jira/browse/KAFKA-3412 will be fixed

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,267 @@
...
+            pollScheduled = false
+            poll()
+          case msg => super.onTimer(msg)
+        }
+      }
+
+      override def postStop(): Unit = {
+        logger.debug("Stage completed. Closing kafka consumer")
+        consumer.close()
+        stoppedPromise.success(Done)
```
13h3r: do we care about how stage completed (success or failure)?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: Got it about scheduling. Thanks

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: ok, adding `updatedWithLatest` ("biggest offset wins")

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: I'm not sure I get what you mean. It's ok to commit to any stage, also with a batch like this that was built up from many `CommittableOffsetImpl`. I'm just using the first stage reference.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: That is how it's documented in KafkaConsumer: "The committed offset should be the next message your application will consume, i.e. lastProcessedMessageOffset + 1."

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: stage is None when creating the empty batch (e.g. seed in a foldLeft)

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: ok, changed name and grabbing the list in the constructor

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: I have improved this

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: I removed the scheduled timeout and fixed the stopping race instead

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: ok, added a TODO

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: I don't think we care, that is communicated downstreams. The Done Future is more a signal for completion of the Control.stop call

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
patriknw: added a stop timeout

core/src/main/scala/akka/kafka/scaladsl/ConsumerExample.scala
patriknw: we can add convenience flows for that, if needed

patriknw:
Thanks for all great feedback. I have incorporated most of it, I think. See latest 3 commits. I will comment more later.

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
13h3r: `mapAsyncUnordered` may change order of input stream. In some case order of messages matters.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -226,42 +181,166 @@ private[kafka] class ConsumerStage[K, V](consumerProvider: () => KafkaConsumer[K
...
+  private var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
+  private var pollScheduled = false
+  private val assignments = consumer.assignment().asScala.toArray
+
+  private val pollCallback = getAsyncCallback[Unit] { _ => schedulePoll() }
+  private val stopCallback = getAsyncCallback[Unit] { _ => stopInternal() }
+
+  setHandler(out, new OutHandler {
+    override def onPull(): Unit = {
+      // TODO preemptively try to fetch new batches when we are below a low watermark?
```
13h3r: Should akka-stream control this via `RequestStrategy`?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -165,57 +138,39 @@ private[kafka] class ConsumerStage[K, V](consumerProvider: () => KafkaConsumer[K
...
       }
 
-      // impl of GraphStage.Committer that is called from the outside via CommittableOffsetBatchImpl
+      // impl of ConsumerStage.Committer that is called from the outside via CommittableOffsetBatchImpl
       override def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Future[Done] = {
-        // not important that we complete with failure during stopping race,
-        // it will be completed with TimeoutException
         if (stoppedPromise.isCompleted)
-          Future.failed(new IllegalStateException("ConsumerStage is stopped, offset commit not performed"))
+          Future.failed(new IllegalStateException(s"ConsumerStage is stopped, offset commit not performed"))
```
13h3r: why `s` here?

core/src/test/resources/logback-test.xml
```diff
@@ -10,6 +10,7 @@
     <!--<logger name="org.apache.kafka.common.utils.AppInfoParser" level="ERROR"/>-->
     <!--<logger name="org.apache.kafka.clients.NetworkClient" level="ERROR"/>-->
     <!--<logger name="com.softwaremill.react" level="TRACE"/>-->
```
13h3r: drop this then?

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -42,10 +44,10 @@ object Consumer {
...
 
   trait CommittableOffset extends Committable {
-    def offset: Offset
+    def offset: PartitionOffset
   }
 
-  final case class Offset(topic: String, partition: Int, offset: Long)
+  final case class PartitionOffset(topic: String, partition: Int, offset: Long)
 
   object CommittableOffsetBatch {
```
13h3r: Why are we keeping implementation in internal package? We can just build map `(topic, partition) -> msg` and update some elements based on user rule. Next on commit we should commit all messages in this map. Looks like there is no dependency on implementation.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
13h3r: But then `if (offsets.isEmpty)` will triggers

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,346 @@
...
+      override def commit(offset: Consumer.PartitionOffset): Future[Done] = {
+        // committed offset should be the next message the application will consume, i.e. +1
+        commit(Map(new TopicPartition(offset.topic, offset.partition) ->
+          new OffsetAndMetadata(offset.offset + 1)))
+      }
+
+      // impl of ConsumerStage.Committer that is called from the outside via CommittableOffsetBatchImpl
+      override def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Future[Done] = {
+        if (stoppedPromise.isCompleted)
+          Future.failed(new IllegalStateException(s"ConsumerStage is stopped, offset commit not performed"))
```
13h3r: why s here?

13h3r: Also there are a lot of such unused `s` :)

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -0,0 +1,93 @@
...
+    def commit(): Future[Done]
+  }
+
+  trait CommittableOffset extends Committable {
+    def offset: PartitionOffset
+  }
+
+  final case class PartitionOffset(topic: String, partition: Int, offset: Long)
+
+  object CommittableOffsetBatch {
```
13h3r: Why are we keeping implementation in internal package? We can just build map (topic, partition) -> msg and update some elements based on user rule. Next on commit we should commit all messages in this map. Looks like there is no dependency on implementation.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,346 @@
...
+  private var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
+  private var pollScheduled = false
+  private val assignments = consumer.assignment().asScala.toArray
+
+  private val pollCallback = getAsyncCallback[Unit] { _ => schedulePoll() }
+  private val stopCallback = getAsyncCallback[Unit] { _ => stopInternal() }
+
+  setHandler(out, new OutHandler {
+    override def onPull(): Unit = {
+      // TODO preemptively try to fetch new batches when we are below a low watermark?
```
13h3r: Should akka-stream control this via RequestStrategy?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -0,0 +1,346 @@
...
+    def commit(offset: Consumer.PartitionOffset): Future[Done]
+    def commit(offsets: Map[TopicPartition, OffsetAndMetadata]): Future[Done]
+  }
+
+  final case class CommittableOffsetBatchImpl(override val offsets: Map[TopicPartition, OffsetAndMetadata])(stage: Option[Committer])
+      extends CommittableOffsetBatch {
+
+    override def updated(committableOffset: Consumer.CommittableOffset): CommittableOffsetBatch = {
+      val s = stage match {
+        case s @ Some(_) => s
```
13h3r: What will happen in this example (pseudocode):

```scala
source1 ~> zip; zip ~> someAction ~> batch ~> commit
source2 ~> zip
```

Here all commits will be sent to `source1` or `source2`, but it should be sent based on `topic, partition`.

13h3r:
Sorry for doubled notifications. Github assigned my comments to wrong lines after using "Show new changes" button :(

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: right, copy pasta, I'll cleanup

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
patriknw: it's the reference to the stage, which should not be exposed to user

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: Ah, now I finally understand what you are trying to say. The topic/partition/offset is specific for a consumer id (KafkaConsumer). We need to group the "maps" by their origin stage. That is possible to solve. Good catch, thanks.

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
13h3r: why do we need this reference?

patriknw: In the end the commitAsync must be performed by the stage. The stage emits messages that contains individual CommitableOffsetImpl, with ref to their origin stage. The purpose of this CommitableOffsetBatch is to gather several CommittableOffset and commit them in one call to the KafkaKonsumer. Therefore, to commit the batch I need the reference to the stage.

I'm not sure I understand what you mean.

13h3r: Why this code will not work (concept):

```scala
  case class CommittableOffsetBatchImplPubic(commits: Map[TopicPartition, CommittableOffset]) extends CommittableOffsetBatch {
    override def updated(offset: CommittableOffset) = {
      val newOffset = new TopicPartition(offset.offset.topic, offset.offset.partition) -> offset
      CommittableOffsetBatchImplPubic(commits + newOffset)
    }
    override def updatedWithLatest(offset: CommittableOffset) = ???
    override def commit() = {
      implicit val ec = ExecutionContexts.sameThreadExecutionContext //TODO use other context
      Future.sequence(commits.map(_._2.commit())).map(_ => Done)
    }
    override def offsets = ???
  }
```

patriknw: > Why this code will not work (concept):

Is that a question?

commit for topic1 and topic2 could be grouped together if they originated from the same stage. That would be more efficient than committing topic1 and topic2 individually, as in your example. Is my assumption of that wrong or is it not needed?

13h3r: I try to move batcher implementation from impl package to user level to allow users to implement batching in their ways. What do you think about this?

I am not sure about efficiency of commits on kafka client level. Right now (0.9.0.1) `commitAsync` just schedules commits to be performed in next poll session. But in 0.10.0 it can be changed (https://issues.apache.org/jira/browse/KAFKA-3412 and https://github.com/apache/kafka/pull/1108). Looks like that that in 0.10 commit batching will be more efficient.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: I have implemented support for this now, see c3b7030

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -0,0 +1,99 @@
...
+  )
+
+  object CommittableOffsetBatch {
+    val empty: CommittableOffsetBatch = new CommittableOffsetBatchImpl(Map.empty, Map.empty)
+  }
+
+  trait CommittableOffsetBatch extends Committable {
+    def updated(offset: CommittableOffset): CommittableOffsetBatch
+
+    def getOffset(key: ClientTopicPartition): Option[Long]
```
patriknw: @13h3r with these two methods the user should be able to implement whatever batching she likes, e.g. `updatedWithLatest` can be implemented.

patriknw:
I have added the `ConsumeSettings` and `ProducerSettings`, see 04cda7e

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
```diff
@@ -0,0 +1,56 @@
...
+      .mapAsync(1)(_.message.passThrough.commit())
+      .to(Sink.ignore)
+
+  /**
+   * Publish records to Kafka topics and then continue the flow. Possibility to pass through a message, which
+   * can for example be a [[Consumer.CommittableOffset]] or [[Consumer.CommittableOffsetBatch]] that can
+   * be committed later in the flow.
+   */
+  def flow[K, V, PassThrough](settings: ProducerSettings[K, V]): Flow[Message[K, V, PassThrough], Future[Result[K, V, PassThrough]], NotUsed] =
+    Flow.fromGraph(new ProducerStage[K, V, PassThrough](() => settings.creteKafkaProducer())) //FIXME is that Future needed?
```
patriknw: @13h3r @drewhk We discussed the need for the `Future` in the output elements and had slightly different opinions. To understand better I tried this example, which is not far from reality:

    val rnd = new Random(0)
    val start = System.nanoTime()
    Source(1 to 100)
      .map { n =>
        println(s"# upstream $n")
        n
      }
      .map { n =>
        Future {
          println(s"# produce begin $n")
          Thread.sleep(100 + (rnd.nextDouble() * 100).toLong)
          println(s"# produce end   $n")
          n
        }
      }
      .mapAsync(1) { n =>
        n
      }
      .map { n =>
        println(s"# downstream $n, ${(System.nanoTime() - start).nanos.toMillis} ms")
        n
      }
      .runWith(Sink.ignore)

    Thread.sleep(20000)

The results for completing the 100 elements with `withAutoFusing(true)`:

* mapAsync(1): 15804 ms
* mapAsync(10): 1877 ms
* mapAsyncUnordered(10): 1670 ms

and the results with `withAutoFusing(false)`:

* mapAsync(1): 1340 ms
* mapAsync(10): 960 ms
* mapAsyncUnordered(10): 823 ms

I don't think users will get this, and we should remove the Future and define a parallelism setting.

WDYT?

It's also important that an earlier commit doesn't overwrites a later commit. That could happen in my example because of the spawned Future, but in the real code I think it's safe because the calls are serialized by the stage and then I assume that the kafka driver will execute things sequentially.

13h3r: I am sorry, I did not get what is wrong with this example. Could you please explain? Is the problem is in fusing?

Also, why do you think that the earlier commit do not overwrites the later commit? AFAIU kafka handles just offset and updating is just change of offset without any checks.

ktoso: AFAIU the point is that the thing that people will do is `mapAsync(1): 15804 ms` which is 10x slower than all the other ones, so having the future there exposes a risk of people panicking about "omg it's so slow" – if I understood correctly

patriknw: Nothing is wrong. I just wanted to illustrate that it it is not trivial for users to the "right thing" when exposing low level futures that must be mapAsynced. We better do it for them and make it a config setting.

13h3r: Ok, lets add `mapAsync` and drop possibility of `mapAsyncUnordered`

dbuschman7:
All the Scala GraphDSL examples look to be removed from the PR. Is this pattern going to be supported going forward and/or are there plans to add these back in?

patriknw:
We had long discussions back and forth and in the end we decided for the api that you see here. The graph alternative was too cumbersome to wire up. See #113 if you are interested in the discussion.

patriknw:
Pushed a few more commits, more settings and first integration test.
I will add more integration tests and write documentation tomorrow.

patriknw:
I have added documentation. Please take a final look before I squash and merge.

13h3r:
JFYI @patriknw @drewhk 
Yesterday I tried to wire 3 input topics, processing stage and 2 output topics with old graph stage API with at least once guarantee. It was so complex and weird that now I am totally on the side of _rich_ messages. You were right :)

patriknw:
Thanks for that update!
My plan is to release a milestone version today and ask for community feedback.

drewhk:
@13h3r Thanks for checking out!

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Sink that is aware of the [[Consumer#CommittableOffset committable offset]]
+   * from a [[Consumer]]. It will commit the consumer offset when the message has
+   * been published successfully to the topic.
+   *
+   * Note that there is always a risk that something fails after publishing but before
+   * committing, so it is "at-least once delivery" semantics.
+   */
+  def commitableSink[K, V](settings: ProducerSettings[K, V]): Sink[Message[K, V, Consumer.Committable], NotUsed] =
```
davidrusu: :man_facepalming: :woman_facepalming: 
it should be spelted `committable`

Issue #118, opened 2016-03-24
Manual commits are not handled appropriately after a consumer rebalance


saurabhdaftary:
May be I am missing something but this is the behavior I am observing with 0.9.0 reactive-kafka when using ```consumerWithOffsetSink```:

On a consumer rebalance say when you happen to add a new consumer node the old consumer still keeps committing offsets to partition it no longer holds after a rebalance.

The rebalance seem to work ok from brokers point of view - on running ConsumerGroupCommand I see that the partitions are redistributed after I add a new consumer. 

After the rebalance the old existing consumer still commits the same last offset it had when it was holding on to a partition which has now moved to the new consumer.

Ideally, you would expect that given the offset did not increase it will filter that partition-offset pair out in its next commit flush interval (as per the below code):

```
 def commitGatheredOffsets(): Unit = {
    log.debug("Flushing offsets to commit")
    scheduledFlush = None
    val offsetMapToFlush = partitionOffsetMap.diff(committedOffsetMap)
    if (offsetMapToFlush.nonEmpty)
      consumerActor ! CommitOffsets(offsetMapToFlush)
  }
```
However, as you can see from the below logs, even though the ```committedOffsetMap``` contains an entry for this stale partition/offset pair, the consumer still keeps committing it:

```
DEBUG [2016-03-24 00:33:25,343] [] [] [kafka-events-processor-akka.actor.default-dispatcher-120] c.s.r.k.c.ConsumerCommitter akka://kafka-events-processor/user/$o - Flushing offsets to commit
DEBUG [2016-03-24 00:33:25,344] [] [] [] o.a.k.c.c.i.ConsumerCoordinator   - Committed offset 3740702 for partition my_topic-46
DEBUG [2016-03-24 00:33:25,336] [] [] [kafka-events-processor-kafka-publisher-dispatcher-109] c.s.r.k.KafkaActorPublisher akka://kafka-events-processor/user/$l - committed offsets: OffsetMap(Map(my_topic-46 -> 3740701, my_topic-67 -> 3727212,...))
TRACE [2016-03-24 00:33:28,333] [] [] [] o.a.k.c.c.i.ConsumerCoordinator  - Sending offset-commit request with {my_topic-71=OffsetAndMetadata{offset=2778977, metadata=''}, my_topic-55=OffsetAndMetadata{offset=3734666, metadata=''}, ...} to Node(1440332752, 172.22.5.202, 9092)
[...]

PR #119, opened 2016-03-25
Fix for not committing offsets for partitions which are no longer own…

saurabhdaftary:
…ed by this consumer instance

PR for #118 

Successful commits do not update `partitionOffsetMap`. Thus, one ends up committing the same offset on a partition repeatedly. This gets ugly when a partition is no longer assigned to this consumer instance. This is because in such cases we keep committing to the broker a stale fixed offset from this consumer which no in-fact no longer owns this partition.

Fix involves updating the `partitionOffsetMap` with successfully committed offsets (once we get a commit confirmation from the broker). 
Have tested this under load  of ~50,000 messages/sec by adding and removing consumer instances extremely frequently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: scala
+sudo: false
+scala:
+  - "2.11.8"
+jdk:
+  - oraclejdk8
+before_install:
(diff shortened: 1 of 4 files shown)


13h3r:
What about having test for this?

kciesielski:
I wrote a test, because it needed some updates to be done in the API, merging.

PR #122, opened 2016-04-01
update README

patriknw:
* rename artifact to akka-stream-kafka
* update examples


README.md
```diff
@@ -1,15 +1,151 @@
...
+  .withGroupId("group1")
+  .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
+```
+
+Consume messages and store a representation, including offset, in DB:
+
+```Scala
+  db.loadOffset().foreach { fromOffset =>
+    val settings = consumerSettings
+      .withFromOffset(new TopicPartition("topic1", 1), fromOffset)
```
ktoso: naming yak - maybe just fromOffset eventually?

ktoso:
LGTM

README.md
patriknw: yeah, feedback is what we want
here, all those settings modifiers start with `with`
might be enough with `withOffset`

Issue #129, opened 2016-04-03
[0.11-M1] Crashing with large flows


gzoller:
I'm seeing a crash with large flows.  My consume code is given here:

    count = 0
    val consumerSettings = ConsumerSettings(as, new ByteArrayDeserializer, new StringDeserializer, Set("lowercaseStrings"))
      .withBootstrapServers("192.168.99.100:9092")
      .withGroupId("group1")
      .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")

    val graph = RunnableGraph.fromGraph(GraphDSL.create() { implicit builder: GraphDSL.Builder[akka.NotUsed] =>
      import GraphDSL.Implicits._
      type In = CommittableMessage[Array[Byte],String]
      val src = Consumer.committableSource(consumerSettings.withClientId("client1"))  
      val commit = Flow[In].map{ msg => msg.committableOffset.commit()  }
      val show = Flow[In].map{ i => 
        count += 1
        if(i.value=="done") println(s"Time ($count): "+(System.currentTimeMillis() - now))
        i 
      }

      src ~> show ~> commit ~> Sink.ignore
      ClosedShape
    })
    now = System.currentTimeMillis()
    graph.run()

I restart Kafka each run so there's no left-over messages/offsets from previous run.
I pre-populate a single topic with just one partition for each run with some number of messages (Strings).
If I run this for a smaller number of messages pre-populated, maybe up to 10000, it works fine.
If I run it with 100,000 I get this error:

```
[ReactiveKafka-akka.kafka.default-dispatcher-11] ERROR akka.kafka.internal.CommittableConsumerStage$$anon$1 - Error in poll
java.lang.IllegalArgumentException: requirement failed
	at scala.Predef$.require(Predef.scala:207)
	at akka.kafka.internal.ConsumerStageLogic.handleResult$1(ConsumerStage.scala:372)
	at akka.kafka.internal.ConsumerStageLogic.poll(ConsumerStage.scala:384)
	at akka.kafka.internal.CommittableConsumerStage$$anon$1.poll(ConsumerStage.scala:145)
	at akka.kafka.internal.CommittableConsumerStage$$anon$1.commitAsync(ConsumerStage.scala:257)
	at akka.kafka.internal.CommittableConsumerStage$$anon$1.akka$kafka$internal$CommittableConsumerStage$$anon$$commitSingleInternal(ConsumerStage.scala:223)
[...]

PR #130, opened 2016-04-05
fix issue in toggleConsumption, #129

patriknw:
* the `require(!buffer.hasNext)` failed in poll
  if commit was requested when there was no demand and
  still things in the buffer from previous poll
* the reason was that we cached the assignments,
  they are automatically changed when using dynamic
  subscriptions and we must use the current value to
  resume/pause

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
@@ -323,7 +323,6 @@ private[kafka] abstract class ConsumerStageLogic[K, V, Out](
   private val pollDataTimeout = Some(settings.pollTimeout)
   private var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
   private var pollScheduled = false
-  private val assignments = consumer.assignment().asScala.toArray
 
   private val pollCallback = getAsyncCallback[Unit] { _ => schedulePoll() }
   private val stopCallback = getAsyncCallback[Unit] { _ => stopInternal() }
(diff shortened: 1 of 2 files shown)


patriknw:
Refs #129

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -360,15 +359,17 @@ private[kafka] abstract class ConsumerStageLogic[K, V, Out](
   protected def poll(): Unit = {
     try {
       def toggleConsumption() = {
+        // note that consumer.assignment() is automatically changed when using
+        // dynamic subscriptions and we must use the current value to resume/pause
         if (isAvailable(out))
-          consumer.resume(assignments: _*)
+          consumer.resume(consumer.assignment().asScala.toArray: _*)
```
drewhk: *ick*, that conversion is really ugly. Isn't this costly?

drewhk:
LGTM

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
patriknw: yeah, that was why I optimized it previously (this was the original version), but that was before knowing how this actually worked.

Perhaps we can hand-roll the conversion to save the intermediate scala representation. In the end it's a varargs call. Stupid api.

13h3r:
No changes for `require(!buffer.hasNext)`?

patriknw:
That `require` is still valid as far as I can see. Now we properly `pause` the topics when there is no demand and therefore the received messages from `poll` is 0.

13h3r:
May be we should take care about rebalance listener and update these topics?

patriknw:
@13h3r what do you mean by "update these topics"? Are you thinking of that as a way around the varargs conversions?

13h3r:
@patriknw 

The `require` fails because we got some messages during `poll`. We got them because we did not pause topics. We did call `pause` and some topics are not paused. This means list of topics/partitions was changed after we gathered them. I may be wrong, so correct me if any. 

If all above is true we have two ways - always get and convert these topics or keep them actual in case of changing. I am talking about subscription to topics/partitions changes and updating this list of topics to pause/resume.

If I am wrong it will be a fail :)

patriknw:
That is exactly the bug fix I have done here. The problem was that I grabbed the assignments once in the constructor. The fix is to grab them from the consumer every time, as in your original code.

We can work on performance improvement via `RebalanceListener` in a separate issue/pr.

Issue #117, opened 2016-03-23
remove dependency to scala-logging


patriknw:
less dependencies is better

PR #131, opened 2016-04-11
remove dependency to scala-logging, #117

patriknw:


patriknw:
Refs #117

core/src/main/scala/akka/kafka/internal/StageLogging.scala
```diff
@@ -0,0 +1,33 @@
...
+trait StageLogging { self: GraphStageLogic =>
+
+  private var _log: LoggingAdapter = _
+
+  protected def logSource: Class[_] = this.getClass
+
+  def log: LoggingAdapter = {
+    // only used in StageLogic, i.e. thread safe
+    if (_log eq null) {
+      materializer match {
```
ktoso: instead: `ActorMaterializer.downcast(materializer)`

ktoso:
Minor nitpick, LGTM otherwise

core/src/main/scala/akka/kafka/internal/StageLogging.scala
patriknw: but that throws?

ktoso: Yes, hm I see... here we continue without logging, maybe that's fine, not sure – print to System.error perhaps instead of println then?

patriknw: logging is not important in those stages, I can even remove the println, this is anyway a workaround until the akka feature has been implemented

ktoso: Okey, I see :+1:

Issue #93, opened 2016-02-25
Expose ConsumerRebalanceListener or something similar so one can handle partition assignment changes 


vishnusaran:
Hi,

Kafka exposes ConsumerRebalanceListener to handle scenarios where application needs to know when the partition assignment changes. In my scenario, i am aggregating data from Kafka and every time a  partition assignment changes, i have reset the aggregation. 
Is there way to expose the ConsumerRebalanceListener ReactiveKafka.consumer API?

Thanks!

kciesielski:
Hi, can you point me to some example of how a `ConsumerRebalanceListener` can be used with pure Kafka Client API? I'm not sure how we can fit this concept into reactive streams API.

13h3r:
From my point of view we can add additional `Out` for consumer which will emit information about rebalancing. So we will have such `Outs`:
- messages
- commit confirmation
- rebalance events

vishnusaran:
In my scenario, I am building an in-memory aggregated cache which aggregates all the data in kafka. I am not committing any offsets, so when a rebalance is triggered, the consumer will start consuming from the smallest offset (auto.offset.reset = smallest) and rebuild the cache. If i don't know if a rebalance is triggered, then I won't be able to reset the cache and the cache will have invalid information.

13h3r:
@vishnusaran do you want to completely reread all data from associated partitions or only from newly associated? What if some partition will be removed from client?

vishnusaran:
Since i do not commit the offsets, every time a  rebalance triggers, kafka will start sending data for all partitions starting from smallest offset (auto.offset.reset = smallest). So i this scenario, I will have to reset the cache. meaning i have reread all the data into the cache for all the partitions that i am currently assigned to. 
ConsumerRebalanceListener has hooks to partition getting removed and getting added. 
In my scenario, I just need to know that partition reassignment happened, so I can reset the cache.

If we add additional out, we can send the previous partition assignment and current assignment when a rebalance is triggered.

13h3r:
Right now we do not have any way to reset an offset of a running graph. We have only commit `In`. Correct solution IMO will be to restart a graph instance. But the graph instance owns instance of kafka client and we want to keep it. Right now I have no idea how to implement it in stream manner. 

Also we can provide materialized kafka consumer to user and you can setup your handler. But you also need to push information about rebalancing to your graph to invalidate current cache. Correct?

[...]

PR #133, opened 2016-04-12
Expose topic partitions to API

13h3r:
In this PR, I propose an API which exposes topic partitions. This API allows:
- backpressure per partition
- react to partition assignment and revoking

The idea here is to react to an automatic partition assignment (`subscribe` and `ConsumerRebalanceListener`) and emit a nested `Source` when partition assigned and close this `Source` when partition revoked.

### DIFF
--- a/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
+++ b/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
@@ -14,6 +14,7 @@ import akka.kafka.internal.PlainConsumerStage
 import akka.stream.scaladsl.Source
 import org.apache.kafka.clients.consumer.ConsumerRecord
 import akka.stream.ActorAttributes
+import org.apache.kafka.common.TopicPartition
 
 /**
  * Akka Stream connector for subscribing to Kafka topics.
@@ -171,5 +172,9 @@ object Consumer {
     }
   }
 
+  def committablePartitionedSource[K, V](settings: ConsumerSettings[K, V]):
+    Source[(TopicPartition, Source[CommittableMessage[K, V], Control]), Control] = {
+    ???
+  }
 }
 
--- a/core/src/test/scala/examples/ConsumerExample.scala
+++ b/core/src/test/scala/examples/ConsumerExample.scala
@@ -6,9 +6,7 @@ package examples
 
 import scala.concurrent.duration._
 import scala.concurrent.Future
-import akka.Done
-import org.apache.kafka.clients.consumer.KafkaConsumer
-import org.apache.kafka.clients.producer.KafkaProducer
+import akka.{Done, NotUsed}
 import org.apache.kafka.clients.producer.ProducerRecord
 import akka.kafka.scaladsl.Consumer.CommittableOffsetBatch
 import akka.kafka.scaladsl.Consumer
@@ -18,6 +16,8 @@ import akka.kafka.ConsumerSettings
 import akka.kafka.ProducerSettings
 import org.apache.kafka.common.TopicPartition
 import akka.actor.ActorSystem
+import akka.stream.{ActorMaterializer, ActorMaterializerSettings}
+import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
 import org.apache.kafka.common.serialization.StringDeserializer
 import org.apache.kafka.common.serialization.ByteArrayDeserializer
 import org.apache.kafka.clients.consumer.ConsumerConfig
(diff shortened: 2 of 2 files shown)


core/src/test/scala/examples/ConsumerExample.scala
```diff
@@ -136,3 +140,55 @@ object ConsumerToProducerWithBatchCommits2Example extends ConsumerExample {
...
+      source
+        .via(business)
+        .toMat(Sink.ignore)(Keep.both)
+        .run()
+    }
+    .mapAsyncUnordered(maxPartitions)(_._2)
+}
+
+// Join flows based on automatically assigned partitions
+object ConsumerWithOtherSource extends ConsumerExample {
```
13h3r: In this case, I have to join two topics. Messages in topics have the same rule of partition assignment: if the message was fetched from partition n then the corresponding message from another topic can be found in partition n too.
To prevent manual partition assignment I want to use built-in assignment from kafka for one topic, react to the assignment and perform the manual assignment to another topic.

13h3r:
A discussion about usecases here - #93

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -171,5 +172,9 @@ object Consumer {
     }
   }
 
+  def committablePartitionedSource[K, V](settings: ConsumerSettings[K, V]):
+    Source[(TopicPartition, Source[CommittableMessage[K, V], Control]), Control] = {
```
patriknw: doesn't this open up for the same kind of problems that we originally saw with `groupBy` emitting sources. @drewhk?

13h3r: ping @drewhk

kciesielski: @patriknw Is this issue with `groupBy` discussed somewhere on GH or Gitter, so I could take a look?

ktoso: Some discussion is in the docs: https://github.com/akka/akka/pull/19501/files
and the PR that made them SubFlow https://github.com/akka/akka/issues/19005
Source of Source is inherently a bit dangerous - one needs to make sure to never drop one etc (and other reasons, like gearpump not being able to do that)

kciesielski: Thx @ktoso!

13h3r: I tried to summarize what we want to achieve here - https://github.com/akka/reactive-kafka/issues/93#issuecomment-209861190

I read the explanation and still thinking that `Source[(Topic, Source)]` is viable option. All resources managed from main `Source` and substreams can be explicitly closed via `control.shutdown`. Yes we still can have potentially deadlock, but numbers of partitions are relatively small. 

Other approach is to use SubStreams, but I can not figure out how can we allow to perform a user code in reaction to a partition assignment.

Any other ideas welcome!

patriknw: I think the idea of emitting sources is nice, if it doesn't introduce resource leak traps. I don't know the details of `SubFlow` and if we can use that for a custom stage. Conversion of the groupBy stage to GraphStage is also pending, in progress I think.

13h3r: @patriknw should we wait for comments from @drewhk or just start implementation?

kciesielski: +1 to this source-of-sources solution. Having `control` should be enough for resource management. A new `Source` appearing in the stream is a much nice way to notify about reassignment. If we use `SubFlow` we can emit pairs of `(TopicPartition, CommittableMessage[K, V])`. Detecting partition assignment could be then possible by manual aggregation and analysis of incoming `TopicPartition` values which doesn't feel good.

patriknw: I'll ask Endre to take a look at this tomorrow.

drewhk: I think SubFlow is inherently hard to extend, so that is a no-go. You can achieve the same with groupBy though and having a (Partition, Msg) pair and have your source-of-sources as an addition instead of the baseline API. I am fine with pure source-of-sources as well, but then Sources must have clear lifecycle semantics when they happen to be dropped somewhere and no-one consumes them.

13h3r:
I will implement it as `Source[(TopicPartition, Source[Msg])]` and submit another review

patriknw:
Sounds good.

PR #136, opened 2016-04-13
Correct udpate of the committer state for ConsumerCommitter

pbarron:
The issues causes unnecessary re-committing of already committed offsets. There was a similar issue with KafkaCommitterSink which was fixed ([here](https://github.com/akka/reactive-kafka/commit/c691e194ac355b6c2df208a8ece54d8df501fee4)). The ConsumerCommitter which uses the same logic was, however,  not patched at the same time. The following PR corrects that.


kciesielski:
Thanks for spotting this. It would be good to have a test, similar to https://github.com/akka/reactive-kafka/commit/5d80f7a3df42686490af09a9062439051d9e4c2a, which first requires some API changes (like https://github.com/akka/reactive-kafka/commit/73fcaba652616334f294eaeb1b0e7d1164a8445e). I'll work on this.

PR #137, opened 2016-04-14
Support stop and shutdown in flow Control

### LINKED ISSUE #135: Now way to perform correct shutdown
There is no way to perform correct `Consumer` shutdown in the current API. 

Pseudocode to illustrate:

```scala
val control = Consumer.commitableSource(config)
  .mapAsync(10)(longAsyncTaskWhichTakes1SecondOrSo)
  .mapAsync(batchCommit) //commit once per second
  .to(Sink.ignore)
  .run()
```

The code above starts stream processing. Now there is only `control.stop()` to shut down the `Consumer`. This method waits for no commits in progress and shut down kafka consumer regardless messages processed by my async task.

With this semantic there is no way to perform correct shutdown even when everything is ok.

One of solution can be introducing `stop(finiteDuration)` method, which completes the `Source` immediately, but keeps kafka consumer online for given duration.

13h3r:
Introducing `stop` and `shutdown` in `Control` to be able to perform correct shutdown. Fixes #135

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
@@ -7,24 +7,21 @@ package akka.kafka.internal
 import java.util
 import java.util.Collections
 import java.util.concurrent.TimeoutException
-import scala.collection.JavaConverters._
-import scala.concurrent.Future
-import scala.concurrent.Promise
-import scala.concurrent.duration._
-import scala.util.control.NonFatal
-import akka.Done
-import akka.NotUsed
+
+import akka.{Done, NotUsed}
 import akka.kafka.ConsumerSettings
 import akka.kafka.scaladsl.Consumer
-import akka.kafka.scaladsl.Consumer.ClientTopicPartition
-import akka.kafka.scaladsl.Consumer.CommittableOffsetBatch
+import akka.kafka.scaladsl.Consumer.{ClientTopicPartition, CommittableOffsetBatch}
 import akka.stream._
-import akka.stream.scaladsl._
 import akka.stream.stage._
 import org.apache.kafka.clients.consumer._
 import org.apache.kafka.common.TopicPartition
-import java.util.concurrent.TimeUnit
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{Future, Promise}
+import scala.concurrent.duration._
 import scala.util.Try
+import scala.util.control.NonFatal
 
 /**
  * INTERNAL API
@@ -115,7 +112,7 @@ private[kafka] class CommittableConsumerStage[K, V](settings: ConsumerSettings[K
       private val pollCommitTimeout = Some(settings.pollCommitTimeout)
 
       private var awaitingConfirmation = 0L
-      private var stopping = false
+      private var shutdownInProgress = false
 
       private val decrementConfirmation = getAsyncCallback[NotUsed] { _ =>
         awaitingConfirmation -= 1
@@ -144,12 +141,12 @@ private[kafka] class CommittableConsumerStage[K, V](settings: ConsumerSettings[K
       override protected def poll(): Unit = {
         super.poll()
 
-        if (stopping && !isClosed(out)) {
+        if (shutdownInProgress && !isClosed(out)) {
           log.debug("Stop producing messages, commits in progress {}", awaitingConfirmation)
           complete(out)
         }
 
-        if (stopping && awaitingConfirmation == 0) {
+        if (shutdownInProgress && awaitingConfirmation == 0) {
           completeStage()
         }
       }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -111,18 +111,27 @@ object Consumer {
...
    */
   trait Control {
+
+    /**
+     * Stop producing messages from the `Source`. This does not stop underlying kafka consumer
+     * and does not unsubscribe from any topics/partitions.
+     *
+     * Call [[#shutdown]] to close consumer
+     */
+    def stop(): Unit
```
kciesielski: Shouldn't `stop` return a `Future` too?

13h3r: If it complete when stage is shutdown then it never completes until someone call `shutdown`. This is look confusing for me. 

Typical case for me is:
- call `stop`
- wait for correct stream completion
- call shutdown at stream completion or after some timeout for stream completion.

patriknw: The stop itself (stop reading) is an async operation so I think it makes sense to return `Future[Done]` to signal that the stop operation has been done, i.e. no more messages will be emitted. Might be needed when chaining things.

13h3r: Guys, agreed with you. Will implementit with Patrik's semantic

13h3r:
@patriknw @kciesielski added `Future` to `stop`. Please, take a look

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -435,9 +436,11 @@ private[kafka] abstract class ConsumerStageLogic[K, V, Out](
   // impl of Consumer.Control that is called from outside via materialized value
   override def isShutdown: Future[Done] = shutdownPromise.future
 
-  override def stop(): Unit = {
+  override def stop(): Future[Done] = {
     log.debug("Stop producing messages")
-    stopCallback.invoke(())
+    val result = Promise[Done]
```
patriknw: There is a problem with this. If the stage is already shutdown, this promise will never be completed. Better to implement it as is done for `shutdown`, i.e. with a `val stopPromise` in the stage logic.

13h3r: `result` will be completed in callback so every promise will be completed, but I agreed that calling `complete` multiple times smells.

patriknw: nope, the callback will not executed if the stage is stopped

patriknw: and the I mean as in `postStop`

13h3r: Got it. Nice catch!

13h3r:
Any comments on this?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -314,19 +314,18 @@ private[kafka] abstract class ConsumerStageLogic[K, V, Out](
...
   private var pollScheduled = false
-  private var stopped = false
 
   private val shutdownCallback = getAsyncCallback[Unit] { _ => shutdownInternal() }
-  private val stopCallback = getAsyncCallback[Promise[Done]] { p =>
+  private val stopCallback = getAsyncCallback[Unit] { p =>
     complete(out)
-    stopped = true
-    p.success(Done)
+    stopPromise.success(Done)
```
patriknw: should be `trySuccess`

patriknw:
LGTM, after that `trySuccess`, please squash before merging

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -419,22 +418,27 @@ private[kafka] abstract class ConsumerStageLogic[K, V, Out](
         None
     }
     // close() is blocking (stupid api that doesn't take timeout parameters)
-    // WakeupException will be thrown by close() if the wakeup from stopTimeout is triggered
+    // WakeupException will be thrown by close() if the wakeup from shutdownTimeout is triggered
     Try(consumer.close())
-    stopTimeoutTask.foreach(_.cancel())
-    stoppedPromise.success(Done)
+    shutdownTimeoutTask.foreach(_.cancel())
+    shutdownPromise.success(Done)
```
patriknw: @13h3r `stopPromise.trySuccess(Done)` here also

13h3r:
ping @patriknw

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -419,22 +418,27 @@ private[kafka] abstract class ConsumerStageLogic[K, V, Out](
...
     }
     // close() is blocking (stupid api that doesn't take timeout parameters)
-    // WakeupException will be thrown by close() if the wakeup from stopTimeout is triggered
+    // WakeupException will be thrown by close() if the wakeup from shutdownTimeout is triggered
     Try(consumer.close())
-    stopTimeoutTask.foreach(_.cancel())
-    stoppedPromise.success(Done)
+    shutdownTimeoutTask.foreach(_.cancel())
+    shutdownPromise.trySuccess(Done)
     super.postStop()
```
patriknw: @13h3r perhaps you missed my previous comment. There is a need for `stopPromise.trySuccess(Done)` here also.

13h3r: Miss it, sorry. Fixed now

patriknw:
thanks

PR #145, opened 2016-05-16
add javadsl, #126

patriknw:
Here is a first stab at the javadsl where I have taken the wrapping approach.

It's rather messy. The alternative is to put the "message" classes outside of dsl package and provide both java and scala api in them. The methods that require different signatures are:
```
def commit(): CompletionStage[Done]
def getOffset(key: ClientTopicPartition): Optional[Long]
```

I'm not sure which approach is best here.

I will also convert the examples (readme) and write some smoke tests.


patriknw:
Superseded by non-wrapping approach in https://github.com/akka/reactive-kafka/pull/146

PR #146, opened 2016-05-16
add javadsl, #126

patriknw:
I will also convert the examples (readme) and write some smoke tests.


core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -0,0 +1,119 @@
...
+   * Commit an offset that is included in a [[CommittableMessage]].
+   * If you need to store offsets in anything other than Kafka, this API
+   * should not be used.
+   *
+   * This interface might move into `akka.stream`
+   */
+  trait Committable {
+    def commitScaladsl(): Future[Done]
+
+    def commitJavadsl(): CompletionStage[Done]
```
patriknw: those method names are a bit artificial but must invent two different names

patriknw:
I have ported the examples to the javadsl. I think this is complete. Please review.

patriknw:
Refs #126

PR #164, opened 2016-06-29
Expose all offsets in batch commit

13h3r:
I found that it is impossible to get information about all offsets in batch commit. I need to monitor these offsets after they are committed. Looks like we can just expose them.


patriknw:
LGTM

PR #177, opened 2016-07-12
Expose kafka message

13h3r:
The idea here is to expose raw kafka message and not fields inside of it. I see no reason to hide this message from end user. Also we reduce amount of useless objects to create. 

Any ideas why we should keep kafka messages hidden from end-user?


patriknw:
LGTM

drewhk:
The only need for hiding it would be if it is not thread-safe or it exposes internals that can mess with the working of our model. My knowledge of Kafka is basically zero, so it is your call to evaluate.

If there is nothing unsafe about these types, then better to just expose them.

kciesielski:
As far as I know the API, this type is not unsafe and it doesn't express anything dangerous or messy, +1 to merge this

PR #180, opened 2016-07-13
Make internal trait Committer use traits instead of Impl

mmatloka:


kciesielski:
Thanks Michał. LGTM. I don't think there was any reason for using Impl there. Right, @13h3r ?

13h3r:
LGTM

PR #183, opened 2016-07-19
Benchmarks for 0.11

kciesielski:
More details: https://docs.google.com/document/d/101VfFarQNFBW23PYaT09MSLtmtsK03r7mViGt2PCa0Q/edit?usp=sharing


benchmarks/src/main/scala/akka.kafka.benchmarks/Timed.scala
```diff
@@ -26,18 +26,16 @@ object Timed extends LazyLogging {
...
-      logger.info(s"Test ${name}_$msgCount took ${after - now} ms")
-      reporter(metrics).report()
-    }
+    val msgCount = command.msgCount
+    logger.info(s"Generating fixture for ${name}_$msgCount")
+    val fixture = fixtureGen.generate(msgCount)
+    val metrics = new MetricRegistry()
+    val meter = metrics.meter(name + "_" + msgCount)
+    logger.info(s"Running benchmarks for ${name}_$msgCount")
+    val now = System.currentTimeMillis()
```
patriknw: use `System.nanoTime` instead, and convert to ms at the end

benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
```diff
@@ -0,0 +1,149 @@
...
+
+    var lastProcessedOffset = 0L
+    var accumulatedMsgCount = 0L
+    val semaphore = new Semaphore(1)
+
+    def doCommit(): Unit = {
+      semaphore.acquire()
+      accumulatedMsgCount = 0
+      val offsetMap = Map(new TopicPartition(fixture.topic, 0) -> new OffsetAndMetadata(lastProcessedOffset))
+      logger.debug("Committing offset " + offsetMap.head._2.offset())
```
patriknw: `if (logger.isDebugEnabled)`

patriknw: several places, not sure which ones are important or not

patriknw: ah, you are using LazyLogging

benchmarks/src/main/scala/akka.kafka.benchmarks/Benchmarks.scala
```diff
@@ -0,0 +1,44 @@
...
+        runPerfTest(cmd, ReactiveKafkaConsumerFixtures.noopFixtureGen(cmd), ReactiveKafkaConsumerBenchmarks.consumePlainNoKafka)
+      case "plain-consumer" =>
+        runPerfTest(cmd, KafkaConsumerFixtures.filledTopics(cmd), KafkaConsumerBenchmarks.consumePlain)
+      case "akka-plain-consumer" =>
+        runPerfTest(cmd, ReactiveKafkaConsumerFixtures.plainSources(cmd), ReactiveKafkaConsumerBenchmarks.consumePlain)
+      case "batched-consumer" =>
+        runPerfTest(cmd, KafkaConsumerFixtures.filledTopics(cmd), KafkaConsumerBenchmarks.consumerAtLeastOnceBatched(batchSize = 50000))
+      case "akka-batched-consumer" =>
+        runPerfTest(cmd, ReactiveKafkaConsumerFixtures.commitableSources(cmd), ReactiveKafkaConsumerBenchmarks.consumerAtLeastOnceBatched(batchSize = 50000))
+      case "at-most-once-consumer" =>
```
patriknw: this takes very long time to run, perhaps decrease number of messages (probably same with the akka one)

benchmarks/src/main/scala/akka.kafka.benchmarks/Benchmarks.scala
```diff
@@ -0,0 +1,44 @@
...
+        runPerfTest(cmd, KafkaConsumerFixtures.filledTopics(cmd), KafkaConsumerBenchmarks.consumeCommitAtMostOnce)
+      case "akka-at-most-once-consumer" =>
+        runPerfTest(cmd, ReactiveKafkaConsumerFixtures.commitableSources(cmd), ReactiveKafkaConsumerBenchmarks.consumeCommitAtMostOnce)
+      case "plain-producer" =>
```
patriknw: got exception when running this:
```
15:00:20.386 ERROR o.a.k.c.p.internals.RecordBatch - Error executing user-provided callback on message for topic-partition 50be6b0c-be7f-4f84-b2dc-8e406d2dd572-0:
java.lang.ArithmeticException: / by zero
	at akka.kafka.benchmarks.PerfFixtureHelpers$$anonfun$initTopicAndProducer$1$$anon$1.onCompletion(PerfFixtureHelpers.scala:38) ~[classes/:na]
	at org.apache.kafka.clients.producer.internals.RecordBatch.done(RecordBatch.java:107) ~[kafka-clients-0.10.0.0.jar:na]
	at org.apache.kafka.clients.producer.internals.Sender.completeBatch(Sender.java:312) [kafka-clients-0.10.0.0.jar:na]
	at org.apache.kafka.clients.producer.internals.Sender.handleProduceResponse(Sender.java:272) [kafka-clients-0.10.0.0.jar:na]
	at org.apache.kafka.clients.producer.internals.Sender.access$100(Sender.java:57) [kafka-clients-0.10.0.0.jar:na]
	at org.apache.kafka.clients.producer.internals.Sender$1.onComplete(Sender.java:358) [kafka-clients-0.10.0.0.jar:na]
	at org.apache.kafka.clients.NetworkClient.poll(NetworkClient.java:278) [kafka-clients-0.10.0.0.jar:na]
	at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:229) [kafka-clients-0.10.0.0.jar:na]
	at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:134) [kafka-clients-0.10.0.0.jar:na]
```

patriknw:
@kciesielski The tests looks great!
The results are not so good, or what do you think?

drewhk:
I recommend updating to Akka 2.4.9-RC2 for benchmarking. It improves on the Akka-Stream part, so if anything is bounded by Akka Streams performance should gain. Apart from the simple cases this is unlikely, but one less factor to consider.

drewhk:
Anyway, the numbers don't look too good. I think in the latter half of the week I can spend some time looking at these.

patriknw:
@kciesielski can we squash and merge this? We will try to allocate some time next week to investigate and improve the performance.

kciesielski:
Sorry for late response. I tried different tweaks and here are some observations. The numbers are rough estimates because I ran only one test (akka-plain-consumer) and only a few times.
1. Switching to 2.4.9-RC2 didn't improve the results.
2. Eliminating pause/resume calls gives ~10% speed up
3. Lowering interval between polls gives a little speed up (5-10%)
4. Eliminating interval between polls reduces performance (~10%)
5. Changing GC to GC1 slows reduces perf (~15%)
6. Changing GC to Parallel gives ~10% speed up

I was hunting for a bigger botleneck, so I didn't explore any of these options in depth. We can (and probably should) gain some performance by optimizing self poll calls and reducing allocations of short-living immutable objects. However, I believe that we can find more interesting bottlenecks.

benchmarks/src/main/scala/akka.kafka.benchmarks/Benchmarks.scala
kciesielski: Right, producer tests are unfinished, I haven't been running them yet, focused on looking for perf issues in consumer part.

benchmarks/src/main/scala/akka.kafka.benchmarks/Benchmarks.scala
kciesielski: I tested this one with 2000 messages. Do you think it should be hardcoded here until we get better perf for these cases?

benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
kciesielski: Yes, macros should do the work. Does `ActorLogging` require `if (logger.isDebugEnabled)`?

benchmarks/src/main/scala/akka.kafka.benchmarks/Benchmarks.scala
patriknw: Good. Endre and I started looking at perf today. We found that removing the dispatcher assignment everywhere except for the KafkaConsumerActor improves things. Blocking poll is only done there.

Still some things we think are strange. Will continue tomorrow.

benchmarks/src/main/scala/akka.kafka.benchmarks/Timed.scala
kciesielski: Ok, will do.

benchmarks/src/main/scala/akka.kafka.benchmarks/Benchmarks.scala
kciesielski: Iteresting. I forgot to mention that I experimented with PinnedDispatcher but it didn't do much good.

PR #185, opened 2016-08-02
Implemented optional serializers for settings classes.

e8kor:
Implemented optional serializers for producer settings and deserializers for consumer.

Implemented optional serializers and deserializers, if they are not passed to settings will check configuration and only then throw illegal argument exception, added tests to describe this behaviour.

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -92,6 +92,51 @@ object Subscriptions {
 
 object ConsumerSettings {
 
+  def apply[K, V](
+    system: ActorSystem,
+    keyDeserializer: Option[Deserializer[K]],
+    valueDeserializer: Option[Deserializer[V]]
+  ): ConsumerSettings[K, V] = {
+    val config = system.settings.config.getConfig("akka.kafka.consumer")
+    apply(config, keyDeserializer, valueDeserializer)
+  }
+
+  def apply[K, V](
+    config: Config,
+    keyDeserializer: Option[Deserializer[K]],
+    valueDeserializer: Option[Deserializer[V]]
+  ): ConsumerSettings[K, V] = {
+    val properties = ConfigSettings.parseKafkaClientsProperties(config.getConfig("kafka-clients"))
+    require(
+      keyDeserializer.isDefined || properties.contains(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG),
+      "Key deserializer should be defined or declared in configuration"
+    )
+    require(
+      valueDeserializer.isDefined || properties.contains(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG),
+      "Value deserializer should be defined or declared in configuration"
+    )
+    val pollInterval = config.getDuration("poll-interval", TimeUnit.MILLISECONDS).millis
+    val pollTimeout = config.getDuration("poll-timeout", TimeUnit.MILLISECONDS).millis
+    val stopTimeout = config.getDuration("stop-timeout", TimeUnit.MILLISECONDS).millis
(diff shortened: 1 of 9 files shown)


typesafehub-validator:
Hi @e8kor,

Before we proceed with reviewing this pull request, please sign the Typesafe Contributors License Agreement:

e8kor:
This feature was recently discussed in bitter with @kciesielski.

13h3r:
@e8kor could you explain your motivation in more details? Looks like simple `ConsumerSettings(config)` is enough and we dont need `ConsumerSettings (config, None, None)`. The only case is something like `ConsumerSettings(config, Some(???), None)` but it looks strange for me.

e8kor:
@13h3r Kafka producer and consumer by default creates serializer fields if user have not passed them but they are defined in config. User may have its custom key serializer defined in application and value serializer in config may be useful, I think it can be extremely useful for tests.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -133,7 +182,7 @@ object ConsumerSettings {
     keyDeserializer: Deserializer[K],
     valueDeserializer: Deserializer[V]
   ): ConsumerSettings[K, V] =
-    apply(system, keyDeserializer, valueDeserializer)
+    apply(system, Option(keyDeserializer), Option(valueDeserializer))
```
13h3r: will this `Option` hide an error when I pass null in serializers?

13h3r:
@e8kor So it will look like `ConsumerSettings(config, Some(???), None)`?

e8kor:
@13h3r  Yes it will look like this.

Regarding Java API, if client will pass null as key or value deserializer it will not throw NPE  in most or even all cases, its because Kafka consumer accept null's and instantiate serializers from config if such declared.

e8kor:
Initially, I created this pull request for discussion or suggestion to my implementation, so if reviewer does not like some lines I'm not insisting on them :)

13h3r:
Overall I am not with and not against this feature. I did not face cases when I can use it. @kciesielski @patriknw what do you think?

e8kor:
@13h3r Just to clarify, initial input on this feature is to have serializers configured in akka Kafka configuration under kafka-clients, its not just about to have some serializers passed and some is not.

13h3r:
@e8kor I got it. Sorry for misunderstanding. I like this feature, but what about slightly changed semantics:

```scala
// configured from config
ConsumerSettings(config)
ConsumerSettings(actorSystem)

//configured from apply
ConsumerSettings(config, ser, deser)

//null can not be passed
ConsumerSettings(config, null, deser) // will fail
```

I dont like `apply` with `Option` in parameters because it is useless to write `ConsumerSettings(config, Some(ser), Some(deser))`.

kciesielski:
@e8kor thanks for this PR.
I agree with Alex that we should avoid optional parameters, better to have additional `apply()`. However, this brings some risk of calling the config-based `apply()` and bumping into runtime erros. It also make it a bit more difficult to fully understand the API. More functions to choose -> more confusion for beginners. Is it really common that users want to define serializers in config file? I would recommend more type-safe approach with serializers set in code and resolved in compilation time.

e8kor:
@13h3r  I'll remove option based apply methods.
Regarding
```
ConsumerSettings(config, null, deser) 
```
In current implementation of reactive Kafka with is possible and it will not fail if key serializer declared in config, as I told before previously reactive-kafka accepts null's as serializers and not fail if they declared in config, what I trying to say is when client passing just a config  he understands that all necessary params to instantiate serializers are declared in config.

@kciesielski Config based serializer configuration is only configuration that I have used so far and thats how Kafka producers are recommended to declare.

patriknw:
LGTM

13h3r:
Do we need any documentation update here?

e8kor:
@13h3r If question related to me than I think no :)

13h3r:
Then LGTM

Issue #189, opened 2016-08-10
requirement failed message in 0.11-M4


xelax:
While running 0.11-M4 I've got this error:

2016-08-10 10:10:50,056 ERROR a.a.OneForOneStrategy - requirement failed
java.lang.IllegalArgumentException: requirement failed
	at scala.Predef$.require(Predef.scala:212)
	at akka.kafka.KafkaConsumerActor.poll(KafkaConsumerActor.scala:182)
	at akka.kafka.KafkaConsumerActor$$anonfun$receive$1.applyOrElse(KafkaConsumerActor.scala:101)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:484)
	at akka.kafka.KafkaConsumerActor.aroundReceive(KafkaConsumerActor.scala:56)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:526)
	at akka.actor.ActorCell.invoke(ActorCell.scala:495)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:257)
	at akka.dispatch.Mailbox.run(Mailbox.scala:224)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

kciesielski:
Can you reproduce this error somehow? It is caused by internal `.poll()` call returning different messages for different partitions than expected. It would be good to have a failing test for this.

xelax:
I got it when I had 3 different instances of a listener on the same queue and then I kill one of them, and I see the error in the console of the other one. I will try to see whether I can trigger this reliably.

kciesielski:
Thanks, @xelax 

@13h3r I suspect that consumer assignment changed before `poll()`  (https://github.com/akka/reactive-kafka/blob/master/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala#L167) hence poll returned more than expected, WDYT?

xelax:
Yes, it sounds likely to me, besides my listeners are getting updates in large batches with few thousands of largish messages in them, so it is quite possible that while getting the messages the consumer assignment get changed.

13h3r:
@xelax did you use auto or manual partition assignment assignment?

@kciesielski How assignment may be changed before poll? And how it may broke the code?

[...]

PR #190, opened 2016-08-10
Add more details to error on unexpected poll result

kciesielski:
This is related to #189, we need to have see which partitions were assigned and what was returned in case such error occurs.

### DIFF
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -190,8 +190,9 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
             }
         }
         //check the we got only requested partitions and did not drop any messages
-        require((rawResult.partitions().asScala -- partitionsToFetch).isEmpty,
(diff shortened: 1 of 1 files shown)


13h3r:
LGTM

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -190,7 +190,8 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
             }
         }
         //check the we got only requested partitions and did not drop any messages
-        require((rawResult.partitions().asScala -- requests.keys).isEmpty)
+        require((rawResult.partitions().asScala -- partitionsToFetch).isEmpty,
+          s"Unexpected records polled. Expected: $partitionsToFetch, result: ${rawResult.partitions()}, consumer assignment: ${consumer.assignment()}")
```
patriknw: since `poll` is performance critical we should not use `require`. It allocates a function for the second parameter. please use plain `if (...) throw IllegalArgumentException`

kciesielski: Ok, I'll update this PR.

johanandren:
LGTM

patriknw:
LGTM

PR #191, opened 2016-08-11
make samples runnable and fix client.id vs group.id

patriknw:
* Preparing the samples for inclusion in docs. I have verified
  most of them.
* Found that client.id is not the identifier of the consumer for the
  offset commits. It is group.id.


kciesielski:
LGTM

johanandren:
LGTM

PR #192, opened 2016-08-12
add materialized value to the sinks, and fix types in javadsl

patriknw:
* Future[Done] materialized value for the producer sinks
* the javadsl.Producer returned scaladsl.Sink and scaladsl.Flow


13h3r:
LGTM

PR #194, opened 2016-08-12
Fix rebalancing possibly leading to consumer not working.

jpeel:
If a partition assignment rebalance happens after RequestMessages has been sent but before messages are fetched for any of the requested partitions and there is no overlap between the partitions before and after the rebalance, a new RequestMessages will never be sent and the consumer is effectively dead until a rebalance occurs such that the consumer is reassigned one of the partitions in the RequestMessages that got the KafkaConsumerActor into a stuck state. The KafkaConsumerActor will keep polling on the old partitions and never send anything back that would trigger a new RequestMessages. Setting requested to false when partitions are assigned fixes the problem. My team has seen this problem quite often because we have periods with a great deal of rebalances. Since we adopted this change, we have not seen this problem.

Also, I suggest that we also add in something to cancel any requests from the old partitions in the case that requested is true when the partitions are revoked. There could be a var pendingRevokedPartitions of type set that is added to when partitions are revoked when requested is true. When the next RequestMessages is about to be sent from the SingleSourceLogic, it could check if there are any pendingRevokedPartitions and send a message, UnrequestMessages, to the KafkaConsumerActor which removes those from topic partitions from its requests for the sender. Unrequesting messages isn't as important as this PR, but without this change the consumer is querying unnecessarily for partitions that it isn't assigned.


jpeel:
I just ran the tests five times locally and they passed every time. The test that failed is one that was until very recently ignored because it was failing sometimes.

patriknw:
The test "resume consumer from committed offset" was not recently ignored. It was enabled in 052cfbf451d20b3ef982721789f695fd4245ab0e
Could 10.seconds be too short, sometimes?
I have restarted build of this PR.

patriknw:
@13h3r can you please take a look at this change?

13h3r:
@jpeel, thanks for PR. Could you please add a test for this?

13h3r:
otherwise LGTM

jpeel:
@13h3r I would like to add a test, but I don't see a way to reliably reproduce the problem. I need several consumers in the same group that subscribe to a topic that has several partitions. Then I need to trigger a rebalance by removing a consumer and re-adding it. Then I would send a message to all of the partitions, put all of the messages received by all of the consumers together and make sure that all of the messages put into the partitions come out of the consumers. Even then, there is no guarantee that the partition rebalance will cooperate such that the partitions assigned to the consumers have no overlap. However, the chance of that happening goes down extremely quickly with the number of partitions/consumers involved. With an equal number, N,  of partitions and consumers, there are N factorial number of possible ways and only one of those possible ways after a rebalance will make everything work. I will start working on a test that uses around 15 consumers/partitions unless someone has a better idea.

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -67,6 +67,7 @@ private[kafka] abstract class SingleSourceLogic[K, V, Msg](
   }
 
   val partitionAssignedCB = getAsyncCallback[Iterable[TopicPartition]] { newTps =>
+    requested = false
```
patriknw: I don't this is the correct thing to do. As far as I can see the interaction between the `SingleSourceLogic` and the `KafkaConsumerActor` is designed to have at most one outstanding `RequestMessages` in flight. This can result in that another `RequestMessages` is sent to the actor and I think that can result in weird things. I think we must find another solution.

A reproducer is a good start.

jpeel:
I don't see any problem with a second RequestMessages being sent. RequestMessages is an indicator for the KafkaConsumerActor to request messages for the topic partitions in the message and send them back to the sender of the RequestMessages request if any are received. Currently, there is no guarantee that all requested partitions from a previous request have received messages. If any messages are received for any partitions previously requested, requested is set to false and another RequestMessages is sent with all subscribed partitions. The only danger is one that is already faced: that previously assigned topic partitions for which this consumer is no longer assigned are still listed in the requests variable. I added a commit to this pull request that fixes that and also sets requested to false when topic partitions are revoked.

I also added a test for rebalancing after a RequestMessages has been received by the KafkaConsumerActor and no messages are in the topic. The test sets up a topic with many partitions. It starts an equal number of consumers to the number of partitions using the same group id for all consumers. No data is put into this topic to begin with. The test waits for the partitions to be assigned initially. Then another consumer is started with the same group id. The test waits for the rebalance to occur. A unique message is sent to each partition. If all messages are received then the test succeeds. If the `requests = false` line is removed for being assigned topic partitions and having them revoked in the SingleSourceLogic, the test fails consistently for me.

It looks like there is a failure in the travis build due to a change in a utility function in the IntegrationSpec. I will fix that in a minute.

patriknw:
hmm, I'm not convinced. I think we must think this through, so probably not today.

If the actor gets another `RequestMessages` from the same consumer as is already in progress it will just add it to the map and there will be two replies. With potential duplicate messages.

jpeel:
@patriknw I understand your concern. Multiple entries and multiple replies would be possible if using a multimap for the requests, but it is an immutable.Map where an insertion of the same key will overwrite the previous key/value pair.

However I do strongly encourage thinking this through and taking some time to make sure this is the right way to fix this. It took me quite a while to diagnose this as the problem and to prescribe a remedy for it. There may be a better way to fix this than what I have done. This was the simplest way that I could find.

13h3r:
I agree with @patriknw. What if we send second request for messages and there is reply in my mailbox already?

What about tracking requested partitions on Source side? Something like replacing `requested` with list of topics we did not request messages for.

jpeel:
@13h3r I like your idea. I do see the problem of requesting again when there is a reply my mailbox though if that were the case then right after the reply was received, another request would have been sent anyway, but wouldn't be in this case. Another thing that could help is in the async callback for adding partitions send a message to the SingleLogicStage actor rather than directly calling methods on the actor.

I think that keeping track of which partitions have been requested on the Source side is quite good, although I'm not fully convinced of the necessity of it. I can put together an attempt at that if other people are in favor of it.

patriknw:
@jpeel It would be great if you can try the implementation I did in https://github.com/akka/reactive-kafka/pull/216
I will release 0.11-RC2 now. If you had some tests it would be great if you can contribute them in a new PR.

Issue #195, opened 2016-08-15
Closing over mutable actor state


johanandren:
And mutating it (ouch):
https://github.com/akka/reactive-kafka/blob/master/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala#L98

13h3r:
This is fine, because callback's called in the same thread as poll called.

ktoso:
How come? https://kafka.apache.org/090/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#commitAsync(org.apache.kafka.clients.consumer.OffsetCommitCallback)
I don't see why this would be the case.

johanandren:
Even if it is executed in the same threadpool, it still isn't safe because the threadpool will contain multiple threads, no?

kciesielski:
I saw that too and discussed with @13h3r. This may seem unsafe but it isn't
because callbacks are called in the same thread. Maybe it could be pointed
out in a comment to clarify.

ktoso:
Same as the *calling* thread? this would make it not async though. 
Where is this configured?

13h3r:
Yes, the callback will be executed in the thread which called `poll`. This fact is not mention directly in documentation, but it is how kafka client designed.

johanandren:
Uff. So if I understand it correctly, the callback is executed on the thread that later calls `consumer.poll()` and since that always is the actor itself in the actor `poll()` method it is safe?

johanandren:
Darn, didn't see that comment before I wrote mine. Anyways, I'll PR with a comment that clarifies.

PR #196, opened 2016-08-16
Clarify async callback running on calling thread

johanandren:
Just to save someone elses half an hour when they com upon this :)

Refs #195


ktoso:
Yeah good to add such comment, thanks.

patriknw:
Good. Yeah, that was not obvious

PR #197, opened 2016-08-16
Reusable producer

13h3r:
Implements https://github.com/akka/reactive-kafka/issues/154

Also I removed dispatcher for producer because producer's API is non-blocking.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -4,17 +4,11 @@
 # the same layout. 
 akka.kafka.producer {
   # Tuning parameter of how many sends that can run in parallel.
-  parallelism = 5
+  parallelism = 100
   
   # How long to wait for `KafkaProducer.close`
   close-timeout = 60s
   
-  # Fully qualified config path which holds the dispatcher configuration
-  # to be used by the producer stages. Some blocking may occur.
-  # When this value is empty, the dispatcher configured for the stream
-  # will be used.
-  use-dispatcher = "akka.kafka.default-dispatcher"
-
   # Properties defined by org.apache.kafka.clients.producer.ProducerConfig
   # can be defined in this configuration section.
   kafka-clients {
--- a/core/src/main/scala/akka/kafka/javadsl/Producer.scala
+++ b/core/src/main/scala/akka/kafka/javadsl/Producer.scala
@@ -5,15 +5,17 @@
 package akka.kafka.javadsl
 
 import java.util.concurrent.CompletionStage
-import scala.compat.java8.FutureConverters.FutureOps
 
+import scala.compat.java8.FutureConverters.FutureOps
 import akka.Done
 import akka.NotUsed
 import akka.kafka.{ConsumerMessage, ProducerSettings}
 import akka.kafka.ProducerMessage._
 import akka.kafka.scaladsl
 import akka.stream.javadsl.{Flow, Sink}
-import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.clients.producer.{KafkaProducer, ProducerRecord}
+
+import scala.concurrent.duration.FiniteDuration
 
 /**
  * Akka Stream connector for publishing messages to Kafka topics.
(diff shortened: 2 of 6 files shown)


core/src/main/resources/reference.conf
```diff
@@ -4,17 +4,11 @@
 # the same layout. 
 akka.kafka.producer {
   # Tuning parameter of how many sends that can run in parallel.
-  parallelism = 5
+  parallelism = 100
```
13h3r: 5 is super small :)

core/src/main/scala/akka/kafka/javadsl/Producer.scala
```diff
@@ -31,6 +33,20 @@ object Producer {
       .asJava
 
   /**
+   * The `plainSink` can be used for publishing records to Kafka topics.
+   * The `record` contains a topic name to which the record is being sent, an optional
+   * partition number, and an optional key and value.
+   */
+  def plainSink[K, V](
+    producerProvider: () => KafkaProducer[K, V],
```
patriknw: does this have to be a `() =>` ? isn't the purpose to reuse the same `KafkaProducer` instance?

13h3r: You may want to get kafka producer from outside somehow (like spring bean or so) and put into reactive kafka

patriknw: but can't you do that before materialization? the factory `() =>` is normally when you want to create different instances for each materialization

13h3r: I want to unify both `ProducerSettings` and externally created producer API. In terms of stages I want to provide the API which allows creation of different `KafkaProducer` for each instances. That is why I kept `() => KafkaProducer` instead of just `KafkaProducer`.

patriknw: I'm sorry, I don't understand the problem. Here is a [patch](https://gist.github.com/patriknw/5f9eebb86410d281741b3f4734b96d34) that shows what I would like. It simply removes the `() =>` from the API.

I think that is a better API since it conveys the purpose of the parameter, to share same `KafkaProducer`.

13h3r: The only problem is a impossibility to create different `KafkaProducer` for multiple stream runs that created with ProducerProvider. Yes producer is thread safe, but may be it can be useful. Do we care about this scenario?

patriknw: I don't think that will be common usage, and for that you can define a new stream blueprint and materialize that.

patriknw:
ProducerStage must also know if it's responsible for closing the `producer` or not, see its `postStop`

core/src/main/resources/reference.conf
drewhk: why not 128? ;)

core/src/main/scala/akka/kafka/javadsl/Producer.scala
drewhk: Exactly, this is common in Akka Streams. For example TCP/HTTP connection Flows are tied to an address. If you need a new one, you can create a new Flow.

patriknw: In this case it's even better than Tcp/Http, here it can be materialized many times with same `KafkaProducer`. The `KafkaProducer` is thread safe.

drewhk: You can do that with a TCP outgoingConnection, too ;) Not that it matters, I just wanted to point out that there is precedent for the pattern here.

patriknw:
@13h3r Shall I do the final changes here? I would like this to be included in the release.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
```diff
@@ -42,6 +58,22 @@ object Producer {
...
+   * from a [[Consumer]]. It will commit the consumer offset when the message has
+   * been published successfully to the topic.
+   *
+   * Note that there is always a risk that something fails after publishing but before
+   * committing, so it is "at-least once delivery" semantics.
+   */
+  def commitableSink[K, V](
+    producerProvider: () => KafkaProducer[K, V],
+    closeTimeout: FiniteDuration,
+    parallelism: Int
```
patriknw: I looked some more at this. Having specific parameters like this will be difficult to maintain in a compatible way. It's better to pass in the `ProducerSettings`, which can easier be evolved.

patriknw:
Let's leave this out from the release, instead of rushing it.

patriknw:
I have included the dispatcher removal here: https://github.com/akka/reactive-kafka/pull/206

13h3r:
What we have to do here:
- close producer if needed
- wrap additional parameters
- pass producer by value

patriknw:
@13h3r that's a good summary of what is left here

iravid:
Would it be helpful if I pick this up and complete the changes that @13h3r summarized?

13h3r:
@iravid yes

PR #198, opened 2016-08-16
performance improvements of the consumer

patriknw:
* Dispatcher should not be assigned to the stages, since that introduces
  an unecessary async boundary. All blocking is done in the KafkaConsumerActor.
* scheduling of Poll is not correct, changed to more simple periodic poll
* shorter poll timeout when no requests, but possible commits are also driven
  by poll
* mapAsync(3) in the test for committing the batches
* more realistic batch size of 1000


benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
```diff
@@ -71,6 +71,8 @@ object KafkaConsumerBenchmarks extends LazyLogging {
 
     var lastProcessedOffset = 0L
     var accumulatedMsgCount = 0L
+    // TODO this is actually running single threaded, commitAsync is called on the poll thread.
+    //      It might not be waiting as it should for the commits to be completed?
     val semaphore = new Semaphore(1)
```
patriknw: @kciesielski We are not sure that this test does what it should

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -144,30 +141,30 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+    //set partitions to fetch
+    val partitionsToFetch = requests.keys.toSet
+    consumer.assignment().asScala.foreach { tp =>
+      if (partitionsToFetch.contains(tp)) consumer.resume(java.util.Collections.singleton(tp))
+      else consumer.pause(java.util.Collections.singleton(tp))
+    }
+
+    if (requests.isEmpty) {
+      // poll timeout of 0 doesn't work for the commits
+      val rawResult = consumer.poll(1)
```
patriknw: It's rather weird that the commits are not performed with `poll(0)`.
Is it safe to use `poll(1)` for minimum blocking? Seems to work with local Kafka. Could that be a problem when latency to Kafka server is larger?

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerBenchmarks.scala
```diff
@@ -69,7 +69,7 @@ object ReactiveKafkaConsumerBenchmarks extends LazyLogging {
         batch.updated(elem)
 
       }
-      .mapAsync(1) { m =>
+      .mapAsync(3) { m =>
```
13h3r: Will this possible cause unordered commits? What if I commit two messages with offset 1000 and 2000. In face last commit wins and if unordering is possible here it may cause offset 1000 to be committed last.

patriknw:
With those changes the `akka-batched-consumer` test result is around 450 k msg/s and `akka-plain-consumer` around 800 k msg/s.

13h3r:
I like how simple poll is now

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerBenchmarks.scala
drewhk: ah, this just remained here. We played around to see what effects what. Does not needed.

patriknw: The Commit message to the actor will arrive in order, which means that the `consumer.commitAsync` calls are done in order so I don't see where they can be unordered. We even call `poll` for each received `Commit`.

patriknw: @drewhk well, it makes a difference for performance

drewhk: ok

benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
kciesielski: The idea was to block the loop until `commitAsync()` responds. I don't think that poll/commit single-thread breaks this.

drewhk: that cannot happen on a single thread. If this ever gets blocked, then it would immediately deadlock.

drewhk: The other issue is that this test keeps polling while the commit is in flight and hence it does not measure the same thing as the reactive connector. The Akka consumer test will actually not request form the upstream in mapAsync until the commit has been finished, which means that topics are paused in the consumer actor until the commit finishes. The two are not really comparable.

kciesielski: When the commit is in flight the semaphore is acquired and the loop will poll only once and then pause with next `acquire`. When commit finishes, `onComplete` gets executed on the same thread, which leads to deadlock, right?

drewhk: well, I would put it in another way. If the `acquire()` ever blocks, then the only way to it to get unblocked would be that a previous commit callback is called, calling `release()`. But that can only happen in a `poll()`, which we will not call because the loop is blocked.

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerBenchmarks.scala
kciesielski: What if we have `mapAsync(3)` which will call `commitAsync(1)` which fails but then we call `commitAsync(2)` before we receive a callback on failure?

benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
kciesielski: Thanks @drewhk, that's actually exactly what I meant. So I can see how this is incorrectly designed, but I'm curious why it works. I've been running this test many times and it never froze, which should be quite likely.

drewhk: Yeah, it is weird. We actually had it deadlocked in the at-most-once tests, so this is not only theoretical. Double checked with Yourkit and the thread was actually waiting on that lock and did not progress anymore.

kciesielski: I see. So I tried to do manual threading and got in trouble, another reminder why Akka is awesome ;) I'll fix this test.

drewhk: To be honest, the Kafka client design is more than confusing.

patriknw:
@13h3r @kciesielski another change that improves the commit performance a lot. Current results:
akka-plain-consumer: 850,000 msg/s
akka-batched-consumer: 850,000 msg/s (batch size 1000)
akka-at-most-once-consumer: 4600 msg/s

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -159,10 +158,22 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
-        throw new IllegalStateException(s"Got ${rawResult.count} unexpected messages")
+
+      def checkNoResult(rawResult: ConsumerRecords[K, V]): Unit =
+        if (!rawResult.isEmpty)
+          throw new IllegalStateException(s"Got ${rawResult.count} unexpected messages")
+
+      checkNoResult(consumer.poll(0))
+
+      var i = 10
+      while (i > 0 && commitsInProgress > 0) {
```
kciesielski: Can you give a brief explanation why do we need this loop?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -159,10 +158,22 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+        LockSupport.parkNanos(10 * 1000)
+        val pollTime = System.nanoTime()
```
kciesielski: unused

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -144,31 +140,43 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
```
drewhk: this is not used

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -144,31 +140,43 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+        val pollTimeout = if (i == 1) 1L else 0L
```
drewhk: is this needed given that we had a `poll(0)` before the loop?

kciesielski:
Looks good!
Stats from EC2 (to compare with old benchmarks):
```
akka-plain-consumer 825,000 msg/s              vs 406,000
akka-batched-consumer 441,832 msg/s  vs 119,000
akka-at-most-once-consumer 2,600 msg/s    vs  20 (!)
```

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerBenchmarks.scala
patriknw: You have already processed record 1 and record 2, so it doesn't matter if you commit 2 but not 1. Or did I misunderstand? This is at-least-once delivery.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: commitAsync takes around 200 us on my machine, so poll(1) is wasteful because it means blocking for 1500 us. Therefore we first try poll(0) in a "busy loop". On my machine that succeeds after around 5th attempt with the current 10 us parkNanos.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: thanks

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: thanks

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: This implementation does:
* 10 attempts with poll(0)
* 1 attempt with poll(1)
* then scheduled message after 50 ms

We are of course somewhat guessing here. Might be different on other machines.

drewhk: ah, you reversed the loop direction ;)

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerBenchmarks.scala
13h3r: If you process message 1 and 2 and commit them in order of 2 and 1 then on restart you will start processing from last committed offset (1) and process message 2 once again

kciesielski: @patriknw right, thanks
@13h3r I don't think it's possible to commit in order 2,1 because `commitAsync` calls will always be performed in correct order.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: I see, thanks.

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerBenchmarks.scala
13h3r: @patriknw @kciesielski you are right. It should be fine here

patriknw:
The failing test is the previously know issue https://github.com/akka/reactive-kafka/issues/203

kciesielski:
I think it's safe to merge, the flakey test is unrelated.

PR #199, opened 2016-08-18
Benchmark fixes

kciesielski:
This PR aims to fix threading problems in consumer benchmarks + some other minor issues. Producer benchmarks will be fixed/finished in a separate PR.

### DIFF
--- a/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
+++ b/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
@@ -5,7 +5,7 @@
 package akka.kafka.benchmarks
 
 import java.util
-import java.util.concurrent.Semaphore
+import java.util.concurrent.atomic.AtomicBoolean
 
 import com.codahale.metrics.Meter
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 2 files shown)


benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
```diff
@@ -71,37 +71,40 @@ object KafkaConsumerBenchmarks extends LazyLogging {
 
     var lastProcessedOffset = 0L
     var accumulatedMsgCount = 0L
-    val semaphore = new Semaphore(1)
+    val commitInProgress = new AtomicBoolean(false)
```
patriknw: There is only one thread involved, as far as I understand the Kafka threading model. Callbacks are invoked by the thread invoking `poll`. That would mean that there is no need for atomic.

kciesielski: Ok, fixed

patriknw:
LGTM

patriknw:
I have restarted the Travis build. We have some issue with IntegrationSpec, but that is unrelated to this. Issue created: https://github.com/akka/reactive-kafka/issues/203

patriknw:
@kciesielski rebase and then merge

benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
```diff
@@ -71,39 +69,42 @@ object KafkaConsumerBenchmarks extends LazyLogging {
...
           lastProcessedOffset = record.offset()
-          if (accumulatedMsgCount >= batchSize || noCommitInProgress)
-            doCommit()
+          if (accumulatedMsgCount >= batchSize) {
+            if (!commitInProgress) {
+              commitInProgress = true
+              doCommit()
+            }
+            else // previous commit still in progress
+              consumer.pause(assignment)
```
drewhk: yep, now it does the same as the reactive version

drewhk:
Looks good!

Will you update the benchmark findings eventually?

kciesielski:
@drewhk Yes, I'll update the results.

PR #201, opened 2016-08-18
Producer benchmarks

kciesielski:
This PR finishes the producer benchmarks. Some initial test results:

plain-producer 308000 msg/s
akka-plain-producer 61000 msg/s

Tests performed with parallelism = 100.
I experimented with changing `mapAsync` in `Producer.flow` to `mapAsyncUnordered` but it didn't affect the results.

### DIFF
--- a/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaProducerBenchmarks.scala
+++ b/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaProducerBenchmarks.scala
@@ -17,7 +17,7 @@ object KafkaProducerBenchmarks extends LazyLogging {
    */
   def plainFlow(fixture: KafkaProducerTestFixture, meter: Meter): Unit = {
     val producer = fixture.producer
-    var lastPartStart = System.nanoTime()
+    @volatile var lastPartStart = System.nanoTime()
 
     for (i <- 1 to fixture.msgCount) {
       producer.send(new ProducerRecord[Array[Byte], String](fixture.topic, i.toString))
--- a/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaProducerBenchmarks.scala
+++ b/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaProducerBenchmarks.scala
@@ -7,18 +7,27 @@ package akka.kafka.benchmarks
 import com.codahale.metrics.Meter
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.kafka.clients.producer.ProducerRecord
+import scala.concurrent.duration._
 
 object KafkaProducerBenchmarks extends LazyLogging {
 
+  val logStep = 100000
   /**
    * Streams generated numbers to a Kafka producer. Does not commit.
    */
   def plainFlow(fixture: KafkaProducerTestFixture, meter: Meter): Unit = {
     val producer = fixture.producer
+    var lastPartStart = System.nanoTime()
 
     for (i <- 1 to fixture.msgCount) {
       producer.send(new ProducerRecord[Array[Byte], String](fixture.topic, i.toString))
       meter.mark()
+      if (i % logStep == 0) {
+        val lastPartEnd = System.nanoTime()
+        val took = (lastPartEnd - lastPartStart).nanos
+        logger.info(s"Sent $i, took ${took.toMillis} ms to send last $logStep")
+        lastPartStart = lastPartEnd
+      }
     }
     fixture.close()
   }
--- a/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaProducerBenchmarks.scala
+++ b/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaProducerBenchmarks.scala
@@ -6,15 +6,9 @@ package akka.kafka.benchmarks
 
 import com.codahale.metrics.Meter
 import com.typesafe.scalalogging.LazyLogging
-import org.apache.kafka.clients.consumer.OffsetAndMetadata
 import org.apache.kafka.clients.producer.ProducerRecord
-import org.apache.kafka.common.TopicPartition
-
-import scala.annotation.tailrec
-import scala.collection.JavaConversions._
 
 object KafkaProducerBenchmarks extends LazyLogging {
-  val pollTimeoutMs = 50L
 
   /**
    * Streams generated numbers to a Kafka producer. Does not commit.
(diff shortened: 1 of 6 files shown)


13h3r:
What about reimplementing producer sink as KafkaProducer.send without callback? Looks like most performance degradation is related to this callbacks

benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaProducerBenchmarks.scala
```diff
@@ -6,15 +6,9 @@ package akka.kafka.benchmarks
...
-import org.apache.kafka.common.TopicPartition
-
-import scala.annotation.tailrec
-import scala.collection.JavaConversions._
 
 object KafkaProducerBenchmarks extends LazyLogging {
-  val pollTimeoutMs = 50L
 
   /**
    * Streams generated numbers to a Kafka producer. Does not commit.
```
patriknw: The benchmark below is not doing the right thing, I think. It's not waiting for confirmations that the sends were successful. That is completely without backpressure, and can't be compared with the streaming version.

kciesielski:
Without waiting for confirmation we would loose backpressure and error
handling. I think we should do as @patriknw suggests and update the plain
benchmark to wait for confirmation. I’ll work on it and update this PR.

kciesielski:
Actually, the producer has its internal sending buffer and will block when the buffer is full so maybe we can leverage this for backpressure, while we use the callback only for error handling.

13h3r:
@kciesielski check `max.block.ms` and `buffer.memory` settings

patriknw:
The results will be better (I have tried) when we remove the dispatcher. In progress in https://github.com/akka/reactive-kafka/pull/197

I also noticed that the invocation of `decrementConfirmation` for each message is rather wasteful. I'll improve that part in https://github.com/akka/reactive-kafka/pull/198

patriknw:
Could be interesting to investigate, but feels wrong to me. That will make the stage blocking, and we must run it on separate dispatcher and such.

13h3r:
@patriknw, @kciesielski  I have two ideas how we can improve producer's performance:
1. What if we will have batch confirmation? Special batch producer will accept collection of messages and internally send all but last without confirmation and last one message with confirmation. This allows us to confirm only 1 of 1000's messages.
2. A way to use buffer blocking from producer is to implement it as external entity like did for consumer and be able to block for some time.

WDYT?

drewhk:
> What if we will have batch confirmation? Special batch producer will accept collection of messages and internally send all but last without confirmation and last one message with confirmation. This allows us to confirm only 1 of 1000's messages.

There is no need to change the API for this, the Sink itself can buffer messages and commit after a threshold. And of course if commit fails, try again with the unconfirmed batch. Also, it should then keep a periodic timer to flush the buffer from time to time even if it is not full yet.

drewhk:
(This can be of course configurable)

patriknw:
KafkaProducer already has support for batching (lingering, retries and such) so I don't think we need to implement that ourselves in the stage. Currently one message results in one Future for the result. I think @13h3r suggests that 1000 (batch size) messages would result in one Future. 

We would anyway need to have callbacks for each send, otherwise we would not be sure that all have succeeded.

drewhk:
OK I see!

patriknw:
@kciesielski ok to merge?

kciesielski:
@patriknw I'm working on better `plain-producer` test that simulates limited parallelism. I have it ready but it gives weird results, still investigating.

patriknw:
ok, no problem

kciesielski:
I updated producer benchmarks and here are some of my concerns:
- The streaming benchmark starts to slow down after some time. To illustrate this, I added logging 
of progress (sth like "took 8551 ms to send last 100000 msgs"). After some time this value starts to grow from hundreds of milliseconds to even tens of seconds. Can someone re-check? If this was caused by filled producer buffer then I'd expect the value to grow and stay on a steady level but it keeps growing.
- It is difficult to write a non-reactive simulation of continuous producer limited to "parallelism" factor. I tried with a simple loop and a local variable guarding maximum number of waiting confirmations. This version was super slow (~1000msg/s). I left the original impl but added similar logging as mentioned above.

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaProducerBenchmarks.scala
```diff
@@ -19,23 +19,35 @@ import scala.language.postfixOps
...
       .via(fixture.flow)
       .map {
-        msg => meter.mark(); msg
+        msg =>
+          meter.mark()
+          if (msg.offset % logStep == 0) {
+            val lastPartEnd = System.nanoTime()
+            val took = (lastPartEnd - lastPartStart).nanos
+            logger.info(s"Sent ${msg.offset}, took ${took.toMillis} ms to send last $logStep")
+            lastPartStart = lastPartEnd
```
kciesielski: I probably shouldn't be closing over an externally declared variable here, right?

patriknw: yes, use volatile

patriknw:
that is an interesting finding, we should absolutely look into that

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaProducerBenchmarks.scala
kciesielski: Sure, done.

13h3r:
@kciesielski could you explain why we need volatile here? I am not a pro with it :)

patriknw:
Putting mutable state outside the stage (`map` in this case) is a big anti-pattern. Will not work if materialized several times. This is just a test and we can accept some cheating, since we know that it's only one stage that will update it. It still must have visibility guarantees, since different threads may be used for processing different elements in the stream. That visibility is in practice ensured by Akka streams (Actors), but I think it's a good marker here anyway.

kciesielski:
Thanks for a great explanation @patriknw. I am fully aware that such approach would be unacceptable outside a test/benchmark.

patriknw:
@kciesielski alright to squash and merge this? Then we should look into the degraded producer performance as a separate issue.

kciesielski:
@patriknw sure, squashing and merging.

kciesielski:
@patriknw Looks like this repo needs squash merging to be enabled by admin and I don't have necessary permissions.

patriknw:
It's enabled as far as I can see. When clicking the Merge button I can see two choices.
Don't know why it's still building, though.

kciesielski:
I rebased with master and it got rebuilt. Squashing now, thanks.

patriknw:
@kciesielski With YourKit I found that the problem is `Source(Stream.from(0, 1))`
I assume you want `Source(0 to fixture.msgCount)`

I'm looking if there is something more.

kciesielski:
@patriknw Wouldn't `Source(0 to fixture.msgCount)` generate a large collection in memory if I want, for example, fifty million elements?

drewhk:
No, it will create a Range which is lazy and does not retain the numbers themselves.

kciesielski:
I see, thanks @drewhk.

PR #205, opened 2016-08-19
avoid async callback for each message in ProducerStage

patriknw:


drewhk:
I look into the code as I need a bit more context. Reviewing soon.

drewhk:
Ok, I see now. LGTM, great find.

kciesielski:
Nice, LGTM

patriknw:
the failing test was the one that we have ignored now

PR #210, opened 2016-08-23
[docs] Add a disclaimer on parallelism

kciesielski:
I think this example can lead to problematic design, so it needs some sort of a disclaimer.

### DIFF
--- a/docs/src/main/paradox/consumer.md
+++ b/docs/src/main/paradox/consumer.md
@@ -81,6 +81,8 @@ Scala
 Java
 : @@ snip [dummy](../../test/java/sample/javadsl/ConsumerExample.java) { #atLeastOnce }
 
+The above example uses separate `mapAsync` stages for processing and committing. This guarantees that for `parallelism` higher than 1 we will keep correct ordering of messages sent for commit. 
+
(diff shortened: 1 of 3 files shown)


docs/src/main/paradox/consumer.md
```diff
@@ -81,6 +81,8 @@ Scala
 Java
 : @@ snip [dummy](../../test/java/sample/javadsl/ConsumerExample.java) { #atLeastOnce }
 
+Important: The above example runs message processing and commit in the same `mapAsync` stage. This is fine as long as `parallelism` is set to 1. With higher parallelism it would be possible that some calls to `db.update` will be finished earlier than others and commit will be performed out of order. If you need higher parallelism, use separate stages for message processing and committing.
```
patriknw: good point, should we rewrite the example to use separate stages?

kciesielski: Good idea. I'll do this and update this PR.

drewhk:
LGTM, good point!

kciesielski:
PR updated. Now the Scala/Java examples use separate stages and there's a shorter disclaimer.

patriknw:
LGTM

Issue #211, opened 2016-08-23
Wrong validation at ProducerSettings apply method


glammers1:
Version: 0.11-RC1

[Actual] (https://github.com/akka/reactive-kafka/blob/master/core/src/main/scala/akka/kafka/ProducerSettings.scala#L69)

```scala
require(
      keySerializer.isDefined || properties.contains(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG),
      "Value serializer should be defined or declared in configuration"
    )
```

Expected

```scala
require(
      valueSerializer.isDefined || properties.contains(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG),
      "Value serializer should be defined or declared in configuration"
    )
```

`valueSerializer.isDefined` instead of `keySerializer.isDefined`

[ConsumerSettings is ok](https://github.com/akka/reactive-kafka/blob/master/core/src/main/scala/akka/kafka/ConsumerSettings.scala#L125)

kciesielski:
Thanks for reporting!

glammers1:
Could I create PR?

kciesielski:
Of course, you are welcome :) Can you add a test?

glammers1:
Perfect!

glammers1:
I've been trying to reproduce the bug in a test before change the code expecting the wrong behavior but it is not possible because the bug is related with the case in which we pass serializers as parameters (not args config) but one of them are None, however the apply methods with the Option signature for the Serializers are private so we can't call the methods outside the companion object with None values.

However, we can reproduce the error using the `create` method marked as `Java API` in its Scaladoc. This method also calls `apply` at the end but there is a different between this call and the others:

* Using create

```scala
def create[K, V](
    config: Config,
    keySerializer: Serializer[K],
    valueSerializer: Serializer[V]
  ): ProducerSettings[K, V] =
    apply(config, Option(keySerializer), Option(valueSerializer))
```

* Using apply

```scala
def apply[K, V](
    config: Config,
    keySerializer: Serializer[K],
    valueSerializer: Serializer[V]
  ): ProducerSettings[K, V] = {
    apply(config, Some(keySerializer), Some(valueSerializer))
  }
```

Note how `create` calls `apply` using `Option` instead of `Some`, that detail let me to test the bug passing a `null` value as serializer but I think that is not the correct way to test it. I don't know why exists that differences in the calls to apply.

Using `apply` developers couldn't reproduce the bug.

What do you think?

[...]

Issue #212, opened 2016-08-26
ProducerSettings should validate key serializer


dotta:
Creating a `ProducerSettings` with a `null` key serializer should fail early with an exception, as any attempt of consuming messages with the created `ProducerSettings` will not work.

Here is a self-contained example that can be used to reproduce the problem.

```
object Test {

  import akka.kafka.scaladsl.Producer
  import akka.actor.ActorSystem
  import akka.kafka.ProducerSettings
  import akka.stream.ActorMaterializer
  import akka.stream.ActorMaterializerSettings
  import akka.stream.scaladsl.Source
  import akka.stream.scaladsl.Flow
  import org.apache.kafka.common.serialization.ByteArraySerializer
  import org.apache.kafka.common.serialization.StringSerializer
  import org.apache.kafka.clients.producer.ProducerRecord

  implicit val system = ActorSystem("ReactiveKafka")
  implicit val materializer = ActorMaterializer(ActorMaterializerSettings(system))

  val producerSettings: ProducerSettings[Null, String] = ProducerSettings(system, null, new StringSerializer)
    .withBootstrapServers("localhost:9092")

  val flow = Flow.fromFunction { (message: Int) =>
    new ProducerRecord[Null, String]("topic-foo", message.toString)
  }
                          
  Source.single(1)
    .via(flow)
    .runWith(Producer.plainSink(producerSettings))

}
```

The stream materialization will result in an exception: `org.apache.kafka.common.config.ConfigException: Missing required configuration "key.serializer" which has no default value.`

The above snippet assumes a Kafka broker is running on localhost:9092.

patriknw:
Thanks for reporting, @dotta 
I think this is a duplicate, or at least very related to https://github.com/akka/reactive-kafka/issues/212

@glammers1 Please take a look at this one also and make sure the PR covers it.

dotta:
@patriknw I think you meant #211. Indeed, it is very similar, but #211 seems to be about the value serializer validation, while the problem I had was with the key serializer. But I suppose it's easy to fix the two tickets together.

patriknw:
yes, copy pasta

glammers1:
Yes, the option 2 in #211 

> we change the api to only have apply with Option parameters and create with Optional parameters

should resolve both issues.

btw, `ConsumerSettings` should be updated in the same manner.

PR #213, opened 2016-08-28
Change to Option/Optional key/value serde in Consumer/Producer Settings

glammers1:
* Now APIs (Consumer and Producer) only have `apply` with `Option` parameters and `create` with `Optional` parameters
* Additional two new tests were added for both Consumer/Producer Settings. One of these tests for the Producer reproduce the validation error explained in #211
* Change the `ProducerSettings` API to resolve the bug
* Change broken code due to API modifications

### DIFF
--- a/benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerFixtures.scala
+++ b/benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaConsumerFixtures.scala
@@ -19,7 +19,7 @@ case class ReactiveKafkaConsumerTestFixture[T](topic: String, msgCount: Int, sou
 object ReactiveKafkaConsumerFixtures extends PerfFixtureHelpers {
 
   private def createConsumerSettings(kafkaHost: String)(implicit actorSystem: ActorSystem) =
-    ConsumerSettings(actorSystem, new ByteArrayDeserializer, new StringDeserializer)
+    ConsumerSettings(actorSystem, Some(new ByteArrayDeserializer), Some(new StringDeserializer))
       .withBootstrapServers(kafkaHost)
       .withGroupId(randomId())
       .withClientId(randomId())
--- a/benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaProducerFixtures.scala
+++ b/benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaProducerFixtures.scala
@@ -27,7 +27,7 @@ object ReactiveKafkaProducerFixtures extends PerfFixtureHelpers {
   case class ReactiveKafkaProducerTestFixture[PassThrough](topic: String, msgCount: Int, flow: FlowType[PassThrough])
 
   private def createProducerSettings(kafkaHost: String)(implicit actorSystem: ActorSystem): ProducerSettings[K, V] =
-    ProducerSettings(actorSystem, new ByteArraySerializer, new StringSerializer)
+    ProducerSettings(actorSystem, Some(new ByteArraySerializer), Some(new StringSerializer))
       .withBootstrapServers(kafkaHost)
       .withParallelism(Parallelism)
 
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -4,6 +4,7 @@
  */
 package akka.kafka
 
+import java.util.Optional
 import java.util.concurrent.TimeUnit
 
 import akka.actor.ActorSystem
@@ -16,6 +17,7 @@ import org.apache.kafka.common.serialization.Deserializer
 import scala.annotation.varargs
 import scala.concurrent.duration._
 import scala.collection.JavaConverters._
+import scala.compat.java8.OptionConverters._
 
 sealed trait Subscription
 sealed trait ManualSubscription extends Subscription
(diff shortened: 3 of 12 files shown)


patriknw:
LGTM, thanks

kciesielski:
In #185 we discussed that we'd prefer to avoid using `Option` in this API. Can we keep multiple choices for `.apply()`?

patriknw:
The number of combinations become unmanageable. We could perhaps add one overload where both keySerializer and valueSerializer are mandatory (not Option params), but then we should remove the other variant that has no serializer params.

What about having those 4 (then bias is towards convenience of defining serializer in code):

```scala
 def apply[K, V](
      system: ActorSystem,
      keyDeserializer: Option[Deserializer[K]],
      valueDeserializer: Option[Deserializer[V]])

def apply[K, V](
      config: Config,
      keyDeserializer: Option[Deserializer[K]],
      valueDeserializer: Option[Deserializer[V]])

 def apply[K, V](
      system: ActorSystem,
      valueDeserializer: Deserializer[V])

def apply[K, V](
      config: Config,
      valueDeserializer: Deserializer[V])
```


Similar 4 for java, of course

glammers1:
I understand when both keySerializer and valueSerializer are mandatory, i.e, we will use it for defining serdes in code.

But I'm a little confused with the other variant... If we want to use the serdes defined in config, why should we allow optional parameters to `apply`? The calls to these methods would be:

```scala
ProducerSettings(conf, None, None)
```

Maybe we could change to:

```scala
 def apply[K, V](
      system: ActorSystem,
      keyDeserializer: Option[Deserializer[K]] = None,
      valueDeserializer: Option[Deserializer[V]]= None)

def apply[K, V](
      config: Config,
      keyDeserializer: Option[Deserializer[K]] = None,
      valueDeserializer: Option[Deserializer[V]] = None)
```

But according to the issue #185 and with the initial problem of the issues #211 #212 I think that we could have:

```scala
// just ActorSystem, serde from config
def apply[K, V](
    system: ActorSystem
  ): ProducerSettings[K, V] = {
    apply(system, None, None)
  }

// just Config, serde from config
  def apply[K, V](
    config: Config
  ): ProducerSettings[K, V] = {
    apply(config, None, None)
  }

// ActorSystem and serdes from code BUT if they are passed as null 
// will check configuration and only then throw illegal argument exception
  def apply[K, V](
    system: ActorSystem,
    keySerializer: Serializer[K],
    valueSerializer: Serializer[V]
  ): ProducerSettings[K, V] =
    apply(system, Option(keySerializer), Option(valueSerializer)) // <- Option instead of Some

// Config and serdes from code BUT if they are passed as null 
// will check configuration and only then throw illegal argument exception
  def apply[K, V](
    config: Config,
    keySerializer: Serializer[K],
    valueSerializer: Serializer[V]
  ): ProducerSettings[K, V] = {
    apply(config, Option(keySerializer), Option(valueSerializer)) // <- Option instead of Some
  }

// SAME for Java

  def create[K, V](
    system: ActorSystem
  ): ProducerSettings[K, V] = {
    apply(system, None, None)
  }

  def create[K, V](
    config: Config
  ): ProducerSettings[K, V] = {
    apply(config, None, None)
  }

  def create[K, V](
    system: ActorSystem,
    keySerializer: Serializer[K],
    valueSerializer: Serializer[V]
  ): ProducerSettings[K, V] =
    apply(system, Option(keySerializer), Option(valueSerializer))

  def create[K, V](
    config: Config,
    keySerializer: Serializer[K],
    valueSerializer: Serializer[V]
  ): ProducerSettings[K, V] =
    apply(config, Option(keySerializer), Option(valueSerializer))

// private methods!
  private def apply[K, V](
    system: ActorSystem,
    keySerializer: Option[Serializer[K]],
    valueSerializer: Option[Serializer[V]]
  ): ProducerSettings[K, V] =
    apply(system.settings.config.getConfig("akka.kafka.producer"), keySerializer, valueSerializer)

  private def apply[K, V](
    config: Config,
    keySerializer: Option[Serializer[K]],
    valueSerializer: Option[Serializer[V]]
  ): ProducerSettings[K, V] = { ... }
```

Is the actual API but using `Option` instead of `Some` in calls to `apply` for the Scala API. Being in use already on Java API.

patriknw:
There is also the combination where you want to define one of them in code and the other in config. Using `null` is ugly.

My thinking is that we should make defining them in code convenient, but make it possible to use config. When user sees the `Option`parameters that should trigger the thought that something else is perhaps needed if I pass `None` here.

patriknw:
That said, I'm also fine with only having the one with Option parameters.

glammers1:
I got it. Sorry for misunderstanding. I agree with using `null` is ugly. 

According to this discussion and comments in #185 

@13h3r said:

> I dont like apply with Option in parameters because it is useless to write ConsumerSettings(config, Some(ser), Some(deser)).

@kciesielski said:

> I agree with Alex that we should avoid optional parameters, better to have additional `apply()`. 

I think that we can use [the 4 defined](https://github.com/akka/reactive-kafka/pull/213#issuecomment-243080961) by @patriknw with `None` values by default in the case of `Option`.

patriknw:
No default values, please. That's a mess to evolve, in case we need to add more parameters later. The param also has a purpose, the user has to think.

glammers1:
Perfect, thanks for the explanation, I'm learning a lot here.

I'll update the PR ASAP.

patriknw:
Thanks

patriknw:
@glammers1 Any update here. I would like to get this in and release RC2 tomorrow.

glammers1:
@patriknw I've done a new PR #222 because the changes on this one does not apply after the discussion. I think we can close it without merge anything.

PR #214, opened 2016-08-29
fix memory leak in produer benchmark

patriknw:


patriknw:
Please review @kciesielski

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaProducerBenchmarks.scala
```diff
@@ -29,8 +29,7 @@ object ReactiveKafkaProducerBenchmarks extends LazyLogging {
   def plainFlow(fixture: ReactiveKafkaProducerTestFixture[Int], meter: Meter)(implicit mat: Materializer): Unit = {
     logger.debug("Creating and starting a stream")
     var lastPartStart = System.nanoTime()
-    val future = Source(Stream.from(0, 1))
```
drewhk: We really would need some good support for static code analysis. Stream should be not used ever and there should be ways to enforce this automatically.

drewhk:
LGTM, great find!

benchmarks/src/main/scala/akka.kafka.benchmarks/ReactiveKafkaProducerBenchmarks.scala
kciesielski: Learning a lot with you guys :) Thanks for spotting this.

patriknw:
it's a great exchange in both directions :-)

PR #215, opened 2016-08-29
don't publish docs to maven central

patriknw:


drewhk:
Ah,ok this is not Javadocs, just manual.

patriknw:
yeah, just a small thing on my todo-list

Issue #202, opened 2016-08-18
restart of KafkaConsumerActor not handled


patriknw:
Outstanding requests from `SingleSourceLogic` will never be replied to if the actor is restarted.

Perhaps an simple solution is to send something to them from `postStop` (also invoked by default for restarts), such as an empty `Messages` reply.

drewhk:
> Perhaps an simple solution is to send something to them from postStop (also invoked by default for restarts), such as an empty Messages reply.

This is racy, as there might be registrations in-flight we have not seen yet when postStop is called. I think the usual pattern should be used here (bit simplified, won't compile):

```scala
val subscriptions = new AtomicReference(Nil)

// In consumer stage
def register() = {
  val state = subscriptions.get match {
    case Closed => completeStage()
    case registrations =>
      // cas loop add to registration list
     actor ! PendingWakeups
  }
}
```

In the actor on PendingWakeups one must getAndSet(Nil) to empty the list safely. In postStop() a getAndSet(Completed).foreach(_.close()) will ensure that there are no dangling sources.

This is similar to ActorPublisher, and also the BroadcastHub.

patriknw:
but such in-flight registrations (requests for messages) will be received and handled by the new incarnation of the restarted actor

drewhk:
ah, you are talking about restart, sorry. But then there might be an issue with the completion, we need to check.

patriknw:
I think that is fine, there is watch both ways

PR #216, opened 2016-08-29
Harden KafkaConsumerActor requests in flight

patriknw:
This should solve what we discussed in https://github.com/akka/reactive-kafka/pull/194
Please review @jpeel and @13h3r

### DIFF
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -7,7 +7,7 @@ package akka.kafka
 import java.util
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.regex.Pattern
-import akka.actor.{ Actor, ActorLogging, ActorRef, Cancellable, Props, Status, Terminated }
+import akka.actor.{Actor, ActorLogging, ActorRef, Cancellable, Props, Status, Terminated}
 import akka.event.LoggingReceive
 import org.apache.kafka.clients.consumer._
 import org.apache.kafka.common.TopicPartition
@@ -64,7 +64,7 @@ object KafkaConsumerActor {
 }
 
 private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
-  extends Actor with ActorLogging {
+    extends Actor with ActorLogging {
   import KafkaConsumerActor.Internal._
   import KafkaConsumerActor._
 
@@ -116,7 +116,8 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
     case Stop =>
       if (commitsInProgress == 0) {
         context.stop(self)
-      } else {
+      }
+      else {
         stopInProgress = true
         context.become(stopping)
       }
@@ -127,7 +128,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
   def stopping: Receive = LoggingReceive {
     case Poll =>
       poll()
-    case Stop          =>
+    case Stop =>
     case _: Terminated =>
     case msg @ (_: Commit | _: RequestMessages) =>
       sender() ! Status.Failure(StoppingException())
@@ -183,7 +184,8 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
         i -= 1
       }
 
-    } else {
+    }
+    else {
 
       val rawResult = consumer.poll(pollTimeout().toMillis)
       if (!rawResult.isEmpty) {
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -81,9 +81,11 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
 
   def receive: Receive = LoggingReceive {
     case Assign(tps) =>
+      checkOverlappingRequests("Assign", sender(), tps)
       val previousAssigned = consumer.assignment()
       consumer.assign((tps.toSeq ++ previousAssigned.asScala).asJava)
     case AssignWithOffset(tps) =>
+      checkOverlappingRequests("AssignWithOffset", sender(), tps.keySet)
       val previousAssigned = consumer.assignment()
       consumer.assign((tps.keys.toSeq ++ previousAssigned.asScala).asJava)
       tps.foreach {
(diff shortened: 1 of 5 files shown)


patriknw:
Pushed an improved implementation, and also fixed the restart issue #202

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -74,7 +74,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
   val pollTask: Cancellable =
     context.system.scheduler.schedule(pollInterval(), pollInterval(), self, Poll)(context.dispatcher)
 
-  var requests = Map.empty[TopicPartition, ActorRef]
+  var requests = Map.empty[ActorRef, RequestMessages]
```
drewhk: Use a mutable Map? Immutable Maps are tragically slow...

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -138,20 +137,25 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
     super.preStart()
-    requests = Map.empty[TopicPartition, ActorRef]
     consumer = settings.createKafkaConsumer()
-    commitsInProgress = 0
   }
 
   override def postStop(): Unit = {
     pollTask.cancel()
+
+    // reply to outstanding requests is important if the actor is restarted
```
drewhk: Great!

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -138,20 +137,25 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+    }
+
     consumer.close()
     super.postStop()
   }
 
   def poll() = {
     //set partitions to fetch
-    val partitionsToFetch = requests.keys.toSet
+    val partitionsToFetch: Set[TopicPartition] = requests.values.flatMap(_.topics)(collection.breakOut)
```
drewhk: valuesIterator?

core/src/main/scala/akka/kafka/internal/ExternalSingleSourceLogic.scala
```diff
@@ -55,11 +58,11 @@ private[kafka] abstract class ExternalSingleSourceLogic[K, V, Msg](
 
   val partitionAssignedCB = getAsyncCallback[Iterable[TopicPartition]] { newTps =>
     tps ++= newTps
-    pump()
+    requestMessages()
```
drewhk: Thanks! `pump()` brings back bad memories :D

drewhk:
LGTM, only nitpick is the immutable Map.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: yeah, then I guess I should rewrite to java.util.HashMap and while loops. There are plenty of collection juggling in this actor that can be improved. Not sure I want to do that now. I'll take a look.

13h3r:
Looks like we also need to handle revoke in `KafkaConsumerActor` and remove revoked partitions from `requests`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
13h3r: This map allows to have multiple `ActorRef` requesting for same topics-partition. What about checking for existing requests as part of handling of `RequestMessages`?

drewhk: > Not sure I want to do that now. I'll take a look.

Can be done later, no worries.

patriknw: @13h3r yes, isn't that the point with this shared actor? It's flattened in the `poll` and then the results are sent to the requestors. What do you mean by checking for existing requests?

13h3r: @patriknw with current implementation it is possible to reply to two different actors with same messages. It should not happens from my point of view

patriknw: I think I understand what you mean. Since they all use the same `groupId` they should use unique `TopicPartition`. Then the question is what to do about it if I receive a `RequestMessages` for topics that are already "taken". 

I see two reasons:
1) wrong usage, user has defined things in the wrong way
2) races when switching topics

I see two ways to handle it
a) reject the request, send back an error message, which will fail the stage
b) accept the request, and discard previous request from other stage (send back empty reply, to unblock it)

patriknw: I have added a check, assuming we want to support 2b, but logged as warning. Not sure if it's perfect but we can adjust if someone complains.

13h3r:
@patriknw what about this?

>Looks like we also need to handle revoke in KafkaConsumerActor and remove revoked partitions from requests

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: Then `breakOut` can't be used. For great performance we need to use other collections anyway. We need to get the logic right (and understandable) first, though.

patriknw:
> Looks like we also need to handle revoke in KafkaConsumerActor and remove revoked partitions from requests

@13h3r Not sure what you mean by that. In SingleSourceLogic there is `partitionAssignedCB` and `partitionRevokedCB`. Those are handled as normal `RequestMessages` so I don't see any need for special treatment there.

There is also `Assignment` and `AssignmentWithOffset` messages. Is that what you mean? Should we check for overlapping `requests` for those also?

13h3r:
@patriknw Here it is an order of actions
- consumer sends RequestMessages
- KafkaConsumerActor stores it in requests
- Broker revokes partition
- KafkaConsumerActor calls callback, but do not removes revoked partition from `requests`

Looks like it is not a big problem because we requests partitions based on `consumer.assignment()`, but we should remove revoked partition from `requests` IMO.

patriknw:
thanks for input, I'll take a look at that tomorrow

patriknw:
@13h3r I have added additional checks overlapping topics. See 3a5aaf0
Note that I could not do that for the `Subscribe` and `SubscribePattern` messages, but note that the callback in the stage sends `RequestMessages` immediately when invoked. Feels like there can be races between all those things, but perhaps that doesn't matter.

patriknw:
and the reason for not doing it for `Subscribe` and `SubscribePattern` is that those are only strings of the topic (no partition).

patriknw:
Merging this and releasing RC2, there might be more improvements needed.

Issue #219, opened 2016-08-30
it's blocking


patriknw:
We must re-introduce the dedicated dispatcher because both `send` and `poll` methods are blocking if the broker is not available. `send` returns after timeout (default 60 seconds). `poll` doesn't return at all (at least I had not the patente to wait).

I tried with `sample.scaladsl.AtLeastOnceExample` and `sample.scaladsl.PlainSinkExample` + some printlns in the stages.

The producer dispatcher was removed in https://github.com/akka/reactive-kafka/pull/206

drewhk:
The dispatcher should be on the actor on the consumer side though, but yeah, it looks like the Producer stage must have a dedicated one, too. Sigh, these APIs...

drewhk:
Given that there is already blocking, I still wonder if a spinning-blocking thread approach might be better then...

kciesielski:
@patriknw there's a consumer setting `connections.max.idle.ms` with default value of `540000`  which is 9 minutes. Maybe with lower value the consumer will drop with a timeout exception?

patriknw:
ok, good to know, yeah I didn't wait 9 minutes

drewhk:
Actually we discovered that poll() was blocking (on a NIO selector), but I was under the impression that it is capped by the remaining time from the poll timeout. weeeird.

patriknw:
It is capped to the poll timeout under normal circumstances, but when it can't connect to the broker it is blocking longer. Yeah, weird api.

13h3r:
@patriknw even more, if you lost connection from consumer to broker during metadata request you will wait infinitely. Also we may have check `wakeup` method which aborts current blocking operation.

PR #220, opened 2016-08-30
wakeup consumer in case disconnected from broker, #219

patriknw:
Ignore the first commit, that is from another PR that is not merged yet.

### DIFF
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -13,12 +13,12 @@ import org.apache.kafka.clients.consumer._
 import org.apache.kafka.common.TopicPartition
 import scala.collection.JavaConverters._
 import java.util.concurrent.locks.LockSupport
+import akka.actor.DeadLetterSuppression
 
 object KafkaConsumerActor {
   case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
-  def props[K, V](settings: ConsumerSettings[K, V]): Props = {
+  def props[K, V](settings: ConsumerSettings[K, V]): Props =
     Props(new KafkaConsumerActor(settings)).withDispatcher(settings.dispatcher)
-  }
 
   private[kafka] object Internal {
     //requests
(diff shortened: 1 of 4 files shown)


patriknw:
Refs #219

13h3r:
Thanks! 👍 This is exactly what I want to have here.

Some tests will be nice, but I understand that at the current state of tests it may take too much time

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -35,7 +35,7 @@ object KafkaConsumerActor {
     final case class Messages[K, V](requestId: Int, messages: Iterator[ConsumerRecord[K, V]])
     final case class Committed(offsets: Map[TopicPartition, OffsetAndMetadata])
     //internal
-    private[KafkaConsumerActor] case object Poll
+    private[KafkaConsumerActor] case object Poll extends DeadLetterSuppression
```
drewhk: +1

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -174,31 +177,33 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+        // If it's still not completed it will be tried again after the scheduled Poll.
+        var i = 10
+        while (i > 0 && commitsInProgress > 0) {
+          LockSupport.parkNanos(10 * 1000)
+          val pollTimeout = if (i == 1) 1L else 0L
+          checkNoResult(consumer.poll(pollTimeout))
+          i -= 1
+        }
       }
+      finally wakeupTask.cancel()
```
drewhk: shouldn't it catch the WakeupException and do something with it?

patriknw: I was thinking that normal actor supervision would be enough, restart by default

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: Nice, good to know such trait exists!

kciesielski:
LGTM

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
drewhk: wouldn't that lose all the subscription information?

patriknw: it will notify all stages that have requested messages, that's what I added in postStop elsewhere
for Kafka offsets, it's no different than any other crash/restart

drewhk: Ok

PR #222, opened 2016-09-01
Allow Option/Optional key/value serde in Consumer/Producer Settings #211 #212

glammers1:
This PR is the result of the discussion in #213. The old one does not apply at all.

* Now APIs (Consumer and Producer) allows `apply` with `Option` parameters and `create` with `Optional` parameters. If they are not passed to settings will check configuration and only then throw illegal argument exception.

```scala
def apply[K, V](
      system: ActorSystem,
      keyDeserializer: Option[Deserializer[K]],
      valueDeserializer: Option[Deserializer[V]])

def apply[K, V](
      config: Config,
      keyDeserializer: Option[Deserializer[K]],
      valueDeserializer: Option[Deserializer[V]])

 def apply[K, V](
      system: ActorSystem,
      valueDeserializer: Deserializer[V])

def apply[K, V](
      config: Config,
      valueDeserializer: Deserializer[V])
```
**Similar 4 for java**

* Remove the other variant that has no serializer params.

* Additional two new tests were added for both Consumer/Producer Settings. One of these tests for the Producer reproduce the validation error explained in #211.

* Change the `ProducerSettings` API to resolve the bug.

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -4,6 +4,7 @@
  */
 package akka.kafka
 
+import java.util.Optional
 import java.util.concurrent.TimeUnit
 
 import akka.actor.ActorSystem
@@ -14,8 +15,9 @@ import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.serialization.Deserializer
 
 import scala.annotation.varargs
-import scala.concurrent.duration._
 import scala.collection.JavaConverters._
+import scala.compat.java8.OptionConverters._
+import scala.concurrent.duration._
 
 sealed trait Subscription
 sealed trait ManualSubscription extends Subscription
@@ -97,7 +99,7 @@ object ConsumerSettings {
    * `akka.kafka.consumer`.
    * Key or value deserializer can be passed explicitly or retrieved from configuration.
    */
-  private def apply[K, V](
+  def apply[K, V](
     system: ActorSystem,
     keyDeserializer: Option[Deserializer[K]],
     valueDeserializer: Option[Deserializer[V]]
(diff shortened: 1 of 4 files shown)


kciesielski:
LGTM

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -172,6 +153,7 @@ object ConsumerSettings {
   /**
```
patriknw: regarding above `apply(system, Some(keyDeserializer), Some(valueDeserializer))`
I think we should add
```
require(keyDeserializer != null, "keyDeserializer must be defined")
require(valueDeserializer != null, "valueDeserializer must be defined")
```

Otherwise user can by mistake pass `null` and we end up with `Some(null)`

glammers1: Yes, we could add guard for null values. What is the best way? 

1.Change `Some` to `Option`.

`apply(system, Option(keyDeserializer), Option(valueDeserializer))`

null values would be `None` and it would be validated in the other `apply` method. We have nothing more to change.

2.Add specific validation as you comment.

I vote for 1 because for `create` methods that also calls `apply` using `scala-java8-compat`:

`apply(config, keySerializer.asScala, valueSerializer.asScala)`

`asScala` is converting `null` values to `None`.

glammers1:
PR updated!

* Now validations check `null` values in the main `apply` method (others always calls this one).

```scala
require(
      keySerializer != null &&
        (keySerializer.isDefined || properties.contains(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG)),
      "Key serializer should be defined or declared in configuration"
    )
```

```scala
require(
      valueSerializer != null &&
        (valueSerializer.isDefined || properties.contains(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG)),
      "Value serializer should be defined or declared in configuration"
    )
```

* Change `Some` to `Option` in the overloads of `apply` in order to obtain the same behavior of `scala-java8-compat` `asScala` in `create`.

* Add two new tests to both `ProducerSettingsSpec` and `ConsumerSettingsSpec` to check all the possible calls with `null` values:

```scala
ConsumerSettings(conf, new ByteArrayDeserializer, null)
``` 
uses the following `apply`

```scala
def apply[K, V](
    config: Config,
    valueDeserializer: Deserializer[V]
  ): ConsumerSettings[K, V] = {
    apply(config, Option(keyDeserializer), Option(valueDeserializer))
  }
```

And

```scala
ConsumerSettings(conf, None, null)
```

uses the following `apply`

```scala
def apply[K, V](
    config: Config,
    keyDeserializer: Option[Deserializer[K]],
    valueDeserializer: Option[Deserializer[V]]
  ): ConsumerSettings[K, V] = { ... }
```

Same for `ProducerSettings`

IMHO, all the cases are covered.

patriknw:
LGTM, thanks

Issue #227, opened 2016-09-12
Exception is not propagated  if there is an uncaught error in Sender


nemzsom:
Hi!
I found an issue when tried to use the library (version 0.11) with a kafka 0.9.x server. I got this exception:
```
14:24:13.262 [kafka-producer-network-thread | producer-1] ERROR org.apache.kafka.clients.producer.internals.Sender - Uncaught error in kafka producer I/O thread:
org.apache.kafka.common.protocol.types.SchemaException: Error reading field 'topic_metadata': Error reading array of size 423023, only 34 bytes available
	at org.apache.kafka.common.protocol.types.Schema.read(Schema.java:73)
	at org.apache.kafka.clients.NetworkClient.parseResponse(NetworkClient.java:380)
	at org.apache.kafka.clients.NetworkClient.handleCompletedReceives(NetworkClient.java:449)
	at org.apache.kafka.clients.NetworkClient.poll(NetworkClient.java:269)
	at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:229)
	at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:134)
	at java.lang.Thread.run(Thread.java:745)
```
I know this is caused by the incompatibility between the client (0.10) and the server (0.9) but the issue is that the stream was not terminated. It just hangs. I would expect an exception that a supervisor can handle.

When I run this code:
```
implicit val system = ActorSystem("TEST")
  implicit val materializer = ActorMaterializer(ActorMaterializerSettings(system))

  val supervisor: Supervision.Decider = {
    case e =>
      println(s"Exception in supervisor: ${e.getMessage}")
      e.printStackTrace()
      Supervision.stop
  }

  val producerSettings = ProducerSettings(system, new StringSerializer, new StringSerializer)
    .withBootstrapServers("localhost:9092")

  val done = Source(1 to 100)
    .map(_.toString)
    .map { elem =>
      new ProducerRecord[String, String]("topic1", elem)
    }
    .runWith(Producer.plainSink(producerSettings).withAttributes(ActorAttributes.supervisionStrategy(supervisor)))

  done.onComplete {
    case _ => println("Stream done")
  }
```
neither the `Exception in supervisor` nor the `Stream done` line is printed.

If I switch to a kafka 0.10 server the code works, but the issue is if an error happens in the communication I would like to get an exception.
I attached a little sbt project to reproduce the issue.

Thank you for this great library!

[...]

Issue #224, opened 2016-09-09
improve error handling for WakeupTimeoutException


patriknw:
When I introduced the wakup in https://github.com/akka/reactive-kafka/pull/220 I was thinking that ordinary actor supervision would be enough, but based on feedback in the gitter channel the logging of the exception is causing confusion.

We should implement some error handling for this in the actor itself instead. Then we can have a shorter timeout to let the actor be responsive for other things, but still retry a number of times before finally stopping, and thereby terminating the stream(s).

patriknw:
Let's do this after the 0.11 release. I'll increase the timeout for now to 10 seconds.

nemccarthy:
Hey @patriknw, not sure what was the error report in the gitter chat, but I'm seeing behaviour that sounds familiar from the description;

Start consumer, connects fine and as expected see a subscription is made;
```
16/09/12 18:52:28 INFO utils.AppInfoParser: Kafka version : 0.10.0.1
16/09/12 18:52:28 INFO utils.AppInfoParser: Kafka commitId : a7a17cdec9eaa6c5
16/09/12 18:52:28 INFO internals.AbstractCoordinator: Discovered coordinator 192.168.99.100:32768 (id: 2147482646 rack: null) for group g_segmentActivator.
16/09/12 18:52:28 INFO internals.ConsumerCoordinator: Revoking previously assigned partitions [] for group g_segmentActivator
16/09/12 18:52:28 INFO internals.AbstractCoordinator: (Re-)joining group g_segmentActivator
16/09/12 18:52:28 INFO internals.AbstractCoordinator: Successfully joined group g_segmentActivator with generation 1
16/09/12 18:52:28 INFO internals.ConsumerCoordinator: Setting newly assigned partitions [t_segments1-0] for group g_segmentActivator
```

If I then restart the consumer, it will pause until the timeout is triggered. From going through the message flow it looks like the subscription is never recreated so the system just sits there doing nothing... 

I've tried upping the timeout from 5s to 15s and I still have the same problems. 

Is there a workaround aside from large timeouts?

Logs:
```
16/09/12 18:48:22 INFO consumer.ConsumerConfig: ConsumerConfig values: 
	metric.reporters = []
	metadata.max.age.ms = 300000
	partition.assignment.strategy = [org.apache.kafka.clients.consumer.RangeAssignor]
	reconnect.backoff.ms = 50
	sasl.kerberos.ticket.renew.window.factor = 0.8
[...]

PR #230, opened 2016-09-16
Consumer error handling

kciesielski:
- WakeupException is swallowed until limit is exceeded, then actor stops
- Any other exception stops the actor immediately
- Refs #227 #224

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -52,10 +52,10 @@ akka.kafka.consumer {
   # aborted after this timeout. The KafkaConsumerActor will throw
   # org.apache.kafka.common.errors.WakeupException which will be ignored
   # until max-wakeups limit gets exceeded.
-  wakeup-timeout = 10s
+  wakeup-timeout = 3s
 
   # After exceeding maxinum wakeups the consumer will stop and the stage will fail.
-  max-wakeups = 3
+  max-wakeups = 10
   
   # Fully qualified config path which holds the dispatcher configuration
   # to be used by the KafkaConsumerActor. Some blocking may occur.
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -50,10 +50,12 @@ akka.kafka.consumer {
   
   # If the KafkaConsumer can't connect to the broker the poll will be
   # aborted after this timeout. The KafkaConsumerActor will throw
-  # org.apache.kafka.common.errors.WakeupException, which can be handled
-  # with Actor supervision strategy.
-  # This is subject for change, see issue #224
+  # org.apache.kafka.common.errors.WakeupException which will be ignored
+  # until max-wakeups limit gets exceeded.
   wakeup-timeout = 10s
+
+  # After exceeding maxinum wakeups the consumer will stop and the stage will fail.
+  max-wakeups = 3
   
   # Fully qualified config path which holds the dispatcher configuration
   # to be used by the KafkaConsumerActor. Some blocking may occur.
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -135,9 +135,10 @@ object ConsumerSettings {
     val closeTimeout = config.getDuration("close-timeout", TimeUnit.MILLISECONDS).millis
     val commitTimeout = config.getDuration("commit-timeout", TimeUnit.MILLISECONDS).millis
     val wakeupTimeout = config.getDuration("wakeup-timeout", TimeUnit.MILLISECONDS).millis
+    val maxWakeups = config.getInt("max-wakeups")
     val dispatcher = config.getString("use-dispatcher")
     new ConsumerSettings[K, V](properties, keyDeserializer, valueDeserializer,
-      pollInterval, pollTimeout, stopTimeout, closeTimeout, commitTimeout, wakeupTimeout, dispatcher)
+      pollInterval, pollTimeout, stopTimeout, closeTimeout, commitTimeout, wakeupTimeout, maxWakeups, dispatcher)
   }
 
   /**
@@ -231,6 +232,7 @@ class ConsumerSettings[K, V](
     val closeTimeout: FiniteDuration,
     val commitTimeout: FiniteDuration,
     val wakeupTimeout: FiniteDuration,
+    val maxWakeups: Int,
     val dispatcher: String
 ) {
 
@@ -286,11 +288,12 @@ class ConsumerSettings[K, V](
     closeTimeout: FiniteDuration = closeTimeout,
     commitTimeout: FiniteDuration = commitTimeout,
     wakeupTimeout: FiniteDuration = wakeupTimeout,
+    maxWakeups: Int = maxWakeups,
     dispatcher: String = dispatcher
   ): ConsumerSettings[K, V] =
     new ConsumerSettings[K, V](properties, keyDeserializer, valueDeserializer,
       pollInterval, pollTimeout, stopTimeout, closeTimeout, commitTimeout, wakeupTimeout,
-      dispatcher)
+      maxWakeups, dispatcher)
 
   /**
    * Create a `KafkaConsumer` instance from the settings.
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -278,6 +278,9 @@ class ConsumerSettings[K, V](
   def withDispatcher(dispatcher: String): ConsumerSettings[K, V] =
     copy(dispatcher = dispatcher)
 
+  def withMaxWakeups(maxWakeups: Int): ConsumerSettings[K, V] =
+    copy(maxWakeups = maxWakeups)
+
   private def copy(
     properties: Map[String, String] = properties,
     keyDeserializer: Option[Deserializer[K]] = keyDeserializerOpt,
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -207,8 +207,6 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
           context.stop(self)
           null
       }
-      finally
-        wakeupTask.cancel()
 
     if (requests.isEmpty) {
       try {
@@ -236,8 +234,11 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       finally wakeupTask.cancel()
 
     }
-    else
-      processResult(partitionsToFetch, tryPoll(pollTimeout().toMillis))
+    else {
+      val result = tryPoll(pollTimeout().toMillis)
+      wakeupTask.cancel()
+      processResult(partitionsToFetch, result)
+    }
 
     if (stopInProgress && commitsInProgress == 0) {
       context.stop(self)
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -185,34 +185,31 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       else consumer.pause(java.util.Collections.singleton(tp))
     }
 
-    def tryPoll(timeout: Long, onSuccess: ConsumerRecords[K, V] => Unit): Unit = {
-      Try(consumer.poll(timeout))
-        .transform(
-          rawResult => {
-            wakeupTask.cancel()
-            wakeups = 0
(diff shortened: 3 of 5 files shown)


kciesielski:
There's probably going to be some performance impact, I'll measure it after the design gets accepted and refined.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -77,6 +80,8 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
   var requests = Map.empty[ActorRef, RequestMessages]
   var consumer: KafkaConsumer[K, V] = _
   var commitsInProgress = 0
+  val MaxWakeups = 3
```
13h3r: Should we push this into config?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -181,6 +185,34 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       else consumer.pause(java.util.Collections.singleton(tp))
     }
 
+    def tryPoll(timeout: Long, onSuccess: ConsumerRecords[K, V] => Unit): Unit = {
```
13h3r: what about returning `Try[ConsumerRecords]` here instead of passing `onSuccess`?
This also allow us to drop `recover` and use just one `transform`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -181,6 +185,34 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+            Success(onSuccess(rawResult))
+          },
+          f => {
+            wakeupTask.cancel()
+            Failure(f)
+          }
+        )
+        .recover {
+          case w: WakeupException =>
+            wakeups = wakeups + 1
```
13h3r: should we call another poll immediately?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: yes, it should be added to `ConsumerSettings` and config

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -181,6 +185,34 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       else consumer.pause(java.util.Collections.singleton(tp))
     }
 
+    def tryPoll(timeout: Long, onSuccess: ConsumerRecords[K, V] => Unit): Unit = {
+      Try(consumer.poll(timeout))
```
patriknw: This creates some allocations for each `tryPoll` and we should avoid that in tight loops like this. Can we write this with ordinary `try-catch` even if it might not look as elegant?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -181,6 +185,34 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+          f => {
+            wakeupTask.cancel()
+            Failure(f)
+          }
+        )
+        .recover {
+          case w: WakeupException =>
+            wakeups = wakeups + 1
+            if (wakeups == MaxWakeups) {
+              log.error("WakeupException limit exceeded, stopping")
```
patriknw: include the `w.getMessage`

patriknw: perhaps also log at debug level when wakeups < max

patriknw: ah, already don, as warning

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -181,6 +185,34 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+              context.stop(self)
+            }
+            else
+              log.warning("Consumer interrupted with WakeupException after timeout")
```
patriknw: include the `w.getMessage`, perhaps also mention the name of the timeout config and its current value

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: Yes, sure. I'll rewrite to `try-catch`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: I don't think we need to wait before calling another `poll()`. If the second poll hangs, we have a timeout which will define for how long until next `WakeupException`.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: Yes, but I'll add `w.getMessage` to the warning.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: I left this here to implement later as configurable, if the core algorithm passes review. Will make it configurable.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: I'll change this block to ordinary `try-catch` to avoid unnecessary allocations.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: Ok

kciesielski:
@patriknw @13h3r PR updated, please review changes.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -181,16 +184,42 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
         // no outstanding requests so we don't expect any messages back, but we should anyway
         // drive the KafkaConsumer by polling
 
         def checkNoResult(rawResult: ConsumerRecords[K, V]): Unit =
-          if (!rawResult.isEmpty)
+          if (rawResult != null && !rawResult.isEmpty)
             throw new IllegalStateException(s"Got ${rawResult.count} unexpected messages")
 
-        checkNoResult(consumer.poll(0))
+        checkNoResult(tryPoll(0))
```
patriknw: This will cancel the `wakeupTask`, but it's needed for the below while loop also.
I think you have to cancel outside of `tryPoll`

core/src/main/resources/reference.conf
```diff
@@ -50,10 +50,12 @@ akka.kafka.consumer {
   
   # If the KafkaConsumer can't connect to the broker the poll will be
   # aborted after this timeout. The KafkaConsumerActor will throw
-  # org.apache.kafka.common.errors.WakeupException, which can be handled
-  # with Actor supervision strategy.
-  # This is subject for change, see issue #224
+  # org.apache.kafka.common.errors.WakeupException which will be ignored
+  # until max-wakeups limit gets exceeded.
   wakeup-timeout = 10s
```
kciesielski: Sure

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: Right, done.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -286,11 +288,12 @@ class ConsumerSettings[K, V](
     closeTimeout: FiniteDuration = closeTimeout,
     commitTimeout: FiniteDuration = commitTimeout,
     wakeupTimeout: FiniteDuration = wakeupTimeout,
+    maxWakeups: Int = maxWakeups,
     dispatcher: String = dispatcher
   ): ConsumerSettings[K, V] =
     new ConsumerSettings[K, V](properties, keyDeserializer, valueDeserializer,
       pollInterval, pollTimeout, stopTimeout, closeTimeout, commitTimeout, wakeupTimeout,
-      dispatcher)
+      maxWakeups, dispatcher)
```
patriknw: also add `withMaxWakups`

FYI: So far we don't care about breaking binary compatibility, but later we will and that is the reason why this is not a case class.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -200,42 +227,47 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
-                else acc ++ tpMessages
-            }
-            if (messages.nonEmpty) {
-              ref ! Messages(req.requestId, messages)
-              requests -= ref
-            }
-        }
-      }
+      val result = tryPoll(pollTimeout().toMillis)
+      wakeupTask.cancel()
```
patriknw: would be more safe to cancel in a `try-finally` around the `tryPoll`

patriknw:
@kciesielski only two small nitpicks. Let's fix those and then I can cut a new release.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: Done

core/src/main/scala/akka/kafka/ConsumerSettings.scala
kciesielski: Done. Interesting remark on the case class and binary compatibility, thanks :)

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -200,42 +227,46 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+      processResult(partitionsToFetch, result)
     }
+
     if (stopInProgress && commitsInProgress == 0) {
       context.stop(self)
     }
   }
 
+  private def processResult(partitionsToFetch: Set[TopicPartition], rawResult: ConsumerRecords[K, V]): Unit = {
+    if (rawResult != null & !rawResult.isEmpty) {
```
13h3r: what is the reason to use `&` instead of `&&`?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -200,42 +227,46 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
-                if (acc.isEmpty) tpMessages
-                else acc ++ tpMessages
-            }
-            if (messages.nonEmpty) {
-              ref ! Messages(req.requestId, messages)
-              requests -= ref
-            }
-        }
-      }
+      val result = try tryPoll(pollTimeout().toMillis) finally wakeupTask.cancel()
```
13h3r: Could we have Option here to be sure we did not miss null check

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: right, && it should be

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
patriknw: option means allocations, so it has to be ugly

patriknw: we could throw an exception for the null case and catch it at the end of the `poll` method

```
NoPollResult extends RuntimeException with NoStackTrace
```

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
kciesielski: Snap... thanks, good catch.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
13h3r: I like what Patrick suggest and want to keep scala code nullfree

kciesielski: +1, that's why the first version was `Try`-based. I'll try to go with a `NoPollResult` exception and see if it's better.

kciesielski: @patriknw @13h3r Done and pushed

patriknw:
something wrong with sonatype login right now, I'll try the release tomorrow again

Issue #234, opened 2016-09-25
KafkaConsumerActor checkOverlappingRequests comparing fromStage to requests map value, not the ActorRef key


kyal:
https://github.com/akka/reactive-kafka/blob/master/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala#L142

This comparison will never succeed due to what I believe is a typo and it should be (ref != fromStage && ...) instead.
As such checkOverlappingRequests would never respond to the requesting actor with an empty set of messages and remove the request. No idea if this behavior is acceptable or not but no reason to perform the check if it can never succeed.

kciesielski:
Thanks for reporting @kyal, this indeed looks like a bug.

PR #237, opened 2016-09-28
Fix check for overlapping requests

kciesielski:
- Refs #234


13h3r:
LGTM

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -136,10 +136,10 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
 
   def checkOverlappingRequests(updateType: String, fromStage: ActorRef, topics: Set[TopicPartition]): Unit = {
     // check if same topics/partitions have already been requested by someone else,
-    // which is an indication that something is wrong, but it might be allright when assignments change
+    // which is an indication that something is wrong, but it might be alright when assignments change.
     if (requests.nonEmpty) requests.foreach {
       case (ref, r) =>
-        if (r != fromStage && r.topics.exists(topics.apply)) {
+        if (ref != fromStage && r.topics.exists(topics.apply)) {
```
patriknw: ouch, isn't `equals` great ;-)

Issue #236, opened 2016-09-28
Misconfigured pollTimeout and pollInterval can cause commit starvations


iravid:
Ran into this scenario:
- Configured the following values (mistakenly!): `pollTimeout = 250ms`, `pollInterval = 50ms`
- Ran a graph that consumes messages, does some processing and commits offsets
- Noticed that commit durations steadily got larger until the commit timeout was crossed

After building a version of reactive-kafka with some more logs added, noticed that too much time passes between calling `commitScalaDsl()` and the actual `Commit` message arriving to the consumer actor. The time in between is spent processing increasing amount of `Poll` messages.

It appears that the `Poll` messages kept piling up in the actor's mailbox, starving out other messages.

Perhaps it would be wise to do some of the following:
- Use `scheduleOnce` instead of `schedule` for the `pollTask`, and reschedule it only after `Poll` has been processed
- Use ControlAwareMailbox and mark messages that aren't `Poll` has high priority
- Document this pitfall regarding the values of `pollTimeout` and `pollInterval`

PR #239, opened 2016-10-02
Fix mailbox flooding with Poll messages in KafkaConsumerActor (#236)

iravid:
If the execution time of `poll()` is longer than the configured
`pollInterval`, the actor's mailbox could be flooded with `Poll`
messages, starving out other messages.

This commit changes the scheduling mechanism to use `scheduleOnce`: a
`Poll` message will be scheduled in `preStart()`, and another will be
scheduled once the `Poll` message was processed.

### DIFF
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -76,8 +76,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
   def pollTimeout() = settings.pollTimeout
   def pollInterval() = settings.pollInterval
 
-  val pollTask: Cancellable =
-    context.system.scheduler.schedule(pollInterval(), pollInterval(), self, Poll)(context.dispatcher)
+  var currentPollTask: Cancellable = _
 
   var requests = Map.empty[ActorRef, RequestMessages]
   var consumer: KafkaConsumer[K, V] = _
@@ -117,6 +116,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       consumer.subscribe(Pattern.compile(pattern), new WrappedAutoPausedListener(consumer, listener))
     case Poll =>
       poll()
+      currentPollTask = schedulePollTask()
     case req: RequestMessages =>
       context.watch(sender())
       checkOverlappingRequests("RequestMessages", sender(), req.topics)
@@ -150,6 +150,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
   def stopping: Receive = LoggingReceive {
     case Poll =>
       poll()
+      currentPollTask = schedulePollTask()
     case Stop =>
     case _: Terminated =>
     case msg @ (_: Commit | _: RequestMessages) =>
@@ -160,11 +161,14 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
 
   override def preStart(): Unit = {
     super.preStart()
+
     consumer = settings.createKafkaConsumer()
+    currentPollTask = schedulePollTask()
   }
 
   override def postStop(): Unit = {
-    pollTask.cancel()
+    if (currentPollTask != null)
+      currentPollTask.cancel()
 
     // reply to outstanding requests is important if the actor is restarted
     requests.foreach {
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @iravid,

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -160,11 +161,14 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
     super.preStart()
+
     consumer = settings.createKafkaConsumer()
+    currentPollTask = schedulePollTask()
   }
 
   override def postStop(): Unit = {
-    pollTask.cancel()
+    if (currentPollTask != null)
+      currentPollTask.cancel()
```
patriknw: The problem with this solution is that there is a small (but existing) chance that there is already a `Poll` message in flight in the actor's mailbox when the actor is **restarted**, and that can double the frequency of the polls.

One way to solve that is to include a `target: KafkaConsumer` in the `Poll` message and when receiving `Poll` ignore any `target != this`.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -38,7 +38,7 @@ object KafkaConsumerActor {
     final case class Messages[K, V](requestId: Int, messages: Iterator[ConsumerRecord[K, V]])
     final case class Committed(offsets: Map[TopicPartition, OffsetAndMetadata])
     //internal
-    private[KafkaConsumerActor] case object Poll extends DeadLetterSuppression
+    private[KafkaConsumerActor] case class Poll(target: ActorRef) extends DeadLetterSuppression
```
patriknw: always add `final` to `case class`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
iravid: Nice :-) Hadn't thought of that. I've implemented your suggested solution. `this` would be different across restarts as opposed to `self`, right?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -114,9 +114,16 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       consumer.subscribe(topics.toList.asJava, new WrappedAutoPausedListener(consumer, listener))
     case SubscribePattern(pattern, listener) =>
       consumer.subscribe(Pattern.compile(pattern), new WrappedAutoPausedListener(consumer, listener))
-    case Poll =>
-      poll()
-      currentPollTask = schedulePollTask()
+    case Poll(target) =>
+      if (target == self) {
```
patriknw: no, it must be an instance check, the `ActorRef` will be the same after restart.
`target: KafkaConsumerActor` and `target == this`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -180,7 +194,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
   }
 
   def schedulePollTask(): Cancellable =
-    context.system.scheduler.scheduleOnce(pollInterval(), self, Poll)(context.dispatcher)
+    context.system.scheduler.scheduleOnce(pollInterval(), self, Poll(self))(context.dispatcher)
```
patriknw: to avoid creating a new `Poll` instance all the time:
`val pollMsg = Poll(this)`

iravid:
Let me know if you'd like me to squash after review is done

patriknw:
we can squash from the github ui before merging, thanks

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
iravid: Ah, wait, that cannot work - `this` isn't an ActorRef. Is there a way to implement this when comparing `self`?

patriknw: yes, you want to check that the actor instance (this) is the right one

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
iravid: Added above pollTimeout

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
iravid: Understood now :-)

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
iravid: 👍

iravid:
Thanks for the review

PR #240, opened 2016-10-02
Add API for using an existing producer

iravid:
This continues @13h3r's #197 and incorporates @patriknw and @13h3r's comments.

I rebased this on master and reverted the commit that removes the separate dispatcher again.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/ProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/ProducerStage.scala
@@ -10,6 +10,8 @@ import akka.kafka.ProducerSettings
 import akka.stream._
 import akka.stream.stage._
 import org.apache.kafka.clients.producer.{Callback, KafkaProducer, RecordMetadata}
+
+import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{Future, Promise}
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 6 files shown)


iravid:
Another comment: I continued this branch in the spirit of the original PR, but I find that passing the provider function and a boolean that describes the nature of the provider function to `ProducerStage` somewhat obscure.

Perhaps it would be wiser to create a sealed trait that describes what is being passed:
```
sealed trait ProducerProvider
case class SharedProducer(producer: KafkaProducer) extends ProducerProvider
case class ProducerFactory(factory: () => KafkaProducer) extends ProducerProvider
```

Let me know if you think this is a better approach and I'll adapt the PR.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -19,7 +21,8 @@ import java.util.concurrent.atomic.AtomicInteger
  * INTERNAL API
  */
 private[kafka] class ProducerStage[K, V, P](
-  settings: ProducerSettings[K, V], producerProvider: () => KafkaProducer[K, V]
+  closeTimeout: FiniteDuration,
+  producerProvider: () => KafkaProducer[K, V], closeProducerOnStop: Boolean
```
patriknw: minor subjective nitpick: I'd prefer to have `closeProducerOnStop` after `closeTimeout`

iravid: Agreed - both are configuration parameters. Coming right up...

iravid:
Let me know if you'd like me to squash the commits once review is done

Issue #244, opened 2016-10-13
ProducerMessage.Result should include all RecordMetadata and not just the offset


israel:
KafkaProducer in its onCompletion callback, provides a RecordMetadata for the just written Record, however, the ProducerStage extract only the offset and sets it in ProducerMessage.Result 
I suggest to return the whole RecordMetadata object.

PR #245, opened 2016-10-13
added RecordMetada value to ProducerMessage.Result case class while i…

israel:
…ncluding an offset function for backward compatibility. ProducerStage is now filling in the RecordMetadata it got from the kafka producer onCompletion callback into the ProcuderMessage it returns

### DIFF
--- a/core/src/main/scala/akka/kafka/ProducerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ProducerMessage.scala
@@ -36,7 +36,6 @@ object ProducerMessage {
    */
   final case class Result[K, V, PassThrough](
       metadata: RecordMetadata,
-      //    offset: Long,
       message: Message[K, V, PassThrough]
   ) {
     def offset = metadata.offset()
--- a/core/src/main/scala/akka/kafka/ProducerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ProducerMessage.scala
@@ -31,12 +31,12 @@ object ProducerMessage {
 
   /**
    * Output element of `Consumer#flow`. Emitted when the message has been
-   * successfully published. Includes the original message and the
+   * successfully published. Includes the original message, metadata returned from KafkaProducer and the
    * `offset` of the produced message.
    */
   final case class Result[K, V, PassThrough](
-      metadata: RecordMetadata,
-      message: Message[K, V, PassThrough]
+    metadata: RecordMetadata,
+    message: Message[K, V, PassThrough]
   ) {
     def offset = metadata.offset()
   }
--- a/core/src/main/scala/akka/kafka/ProducerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ProducerMessage.scala
@@ -4,7 +4,7 @@
  */
 package akka.kafka
 
-import org.apache.kafka.clients.producer.ProducerRecord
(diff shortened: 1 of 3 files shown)


kciesielski:
Just out of curiosity - do you need some additional metadata in your case?

israel:
I need the partition number

johanandren:
Refs #244

kciesielski:
@israel Right, good point. Do you have any workaround for current version? I thought it may be possible to use the `DefaultPartitioner` directly to determine partition id, but it takes a parameter of type `Cluster` which is not accessible to our API users.

core/src/main/scala/akka/kafka/ProducerMessage.scala
```diff
@@ -35,8 +35,11 @@ object ProducerMessage {
    * `offset` of the produced message.
    */
   final case class Result[K, V, PassThrough](
-    offset: Long,
-    message: Message[K, V, PassThrough]
-  )
+      metadata: RecordMetadata,
+      //    offset: Long,
```
kciesielski: This can be deleted

israel: or also remove the offset func for backward and let every one access it from the metadata?

israel:
Current workaround for me is a local patched v0.12 I have created.. :-)

core/src/main/scala/akka/kafka/ProducerMessage.scala
kciesielski: Just the commented line, I think it's good idea to have the offset func.

israel: done

core/src/main/scala/akka/kafka/ProducerMessage.scala
```diff
@@ -35,8 +35,10 @@ object ProducerMessage {
    * `offset` of the produced message.
    */
   final case class Result[K, V, PassThrough](
-    offset: Long,
-    message: Message[K, V, PassThrough]
-  )
+      metadata: RecordMetadata,
```
13h3r: is this correctly formatted?

israel: what do you mean?

johanandren: It's indented 4 spaces but should be 2 like the original code.

core/src/main/scala/akka/kafka/ProducerMessage.scala
```diff
@@ -35,8 +35,10 @@ object ProducerMessage {
...
+      message: Message[K, V, PassThrough]
+  ) {
+    def offset = metadata.offset()
```
johanandren: Should we maybe be selective with what we allow access to rather than provide the full `Metadata` to the consumer code (promising we always have an instance of the actual kafka client `Metadata` type may be limiting with regards to evolving the API)?

johanandren: Yes, exactly, but it was more of an open question than an opinion.

kciesielski: Ok :) In my opinion it's better to trust Kafka API as we did with exposing `ConsumerRecord`.

PR #246, opened 2016-10-21
Fix test "consume and commit in batches" 

nafshartous:
Fixed test "consume and commit in batches"

- Increased timeouts due to failures on Travis

- Removed shared mutable state

### DIFF
--- a/core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
+++ b/core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
@@ -10,6 +10,8 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import scala.concurrent.{Await, Future}
 import scala.concurrent.duration._
+import scala.language.postfixOps
+
 import akka.{Done, NotUsed}
 import akka.actor.ActorSystem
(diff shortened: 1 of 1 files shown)


core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -54,25 +56,17 @@ class IntegrationSpec extends TestKit(ActorSystem("IntegrationSpec"))
...
-    group1 = "group1-" + uuid
-    group2 = "group2-" + uuid
-  }
+  val partition0 = 0
 
   val producerSettings = ProducerSettings(system, new ByteArraySerializer, new StringSerializer)
     .withBootstrapServers(bootstrapServers)
 
-  def givenInitializedTopic(): Unit = {
+  def givenInitializedTopic(topic: String): Unit = {
```
patriknw: good catch, `produce` was accessing the vars from the stream thread

nafshartous:
Thanks @patriknw.  I'm going to trigger a bunch of Travis builds just to do a final check that the timing issue is fixed.  Then I can recreate the PR with a single commit in the history if needed.

PR #247, opened 2016-10-22
   Fix test "consume and commit in batches"  

nafshartous:
Fix test "consume and commit in batches" 
- Increased timeouts (was failing intermittently on Travis).
- Verified by triggering multiple Travis builds 
- Removed shared mutable state.


nafshartous:
@patriknw These are the changes you previously reviewed.

Issue #251, opened 2016-10-28
Use Akka's provided StageLogging


ktoso:
Implemented `StageLogging` in Akka, so you can just use it instead of relying on the one in here.
See: https://github.com/akka/akka/issues/18793 

To be removed: https://github.com/akka/reactive-kafka/blob/master/core/src/main/scala/akka/kafka/internal/StageLogging.scala

ktoso:
This is since 2.4.12

jacobtardieu:
I'll take it!

jacobtardieu:
This has been done, the issue can be closed.

ktoso:
Thanks for the ping here :)

PR #261, opened 2016-11-25
Use Akka's provided stage logging, #251

jacobtardieu:
* Update Akka to 2.4.14
* Remove internal StageLogging to use the Akka one


kciesielski:
Thanks :) LGTM

jacobtardieu:
Only once in `ProducerStage` but the wildcard import of akka streams brings Akka's `StageLogging` so no need to import it explicitly.

PR #264, opened 2016-12-20
Adding cross build for Scala 2.12 and upgrading to Kafka 0.10.1.1. Fixes #256.

### LINKED ISSUE #256: Release for Scala 2.12
When would we be able to release this for 2.12?
Some people already are asking so we should look into it.

manub:
* Added resolvers for downstream dependencies.
* Updated travis configuration to cross build.
* Fixed a number of compiler errors caused by introducing Scala 2.12.
* Removed some unused imports.

### DIFF
--- a/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
+++ b/benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
@@ -11,7 +11,7 @@ import org.apache.kafka.clients.consumer.{OffsetCommitCallback, OffsetAndMetadat
 import org.apache.kafka.common.TopicPartition
 
 import scala.annotation.tailrec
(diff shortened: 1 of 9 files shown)


kciesielski:
LGTM, thanks :)

benchmarks/src/main/scala/akka.kafka.benchmarks/KafkaConsumerBenchmarks.scala
```diff
@@ -11,7 +11,7 @@ import org.apache.kafka.clients.consumer.{OffsetCommitCallback, OffsetAndMetadat
 import org.apache.kafka.common.TopicPartition
 
 import scala.annotation.tailrec
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
```
patriknw: very good!

patriknw:
Refs #256

PR #265, opened 2016-12-23
Removing Apache Staging from resolvers.

manub:
I've added this as a workaround until `kafka_2.12` was published on Maven Central. Since it happened, it doesn't need to be in `build.sbt` anymore.


kciesielski:
Thanks!

PR #268, opened 2017-02-02
Feature/poll optimization

agadek:
basing on v0.13,

stream sources pre-buffering,

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -35,7 +35,11 @@ akka.kafka.consumer {
   # Note that non-zero value means that blocking of the thread that
   # is executing the stage will be blocked.
   poll-timeout = 50ms
-  
+
+  # sorces buffer size
+  # adapt accordingly to expected throughput of stream sources
+  source-buffer-size = 100000
+
   # The stage will be await outstanding offset commit requests before
   # shutting down, but if that takes longer than this timeout it will
   # stop forcefully.
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -16,45 +16,23 @@ import scala.collection.JavaConverters._
 import java.util.concurrent.locks.LockSupport
 import akka.actor.DeadLetterSuppression
 import scala.util.control.{NoStackTrace, NonFatal}
-import scala.util.{Failure, Success, Try}
 
 object KafkaConsumerActor {
-  case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
+
   def props[K, V](settings: ConsumerSettings[K, V]): Props =
     Props(new KafkaConsumerActor(settings)).withDispatcher(settings.dispatcher)
-
-  private[kafka] object Internal {
-    //requests
-    final case class Assign(tps: Set[TopicPartition])
-    final case class AssignWithOffset(tps: Map[TopicPartition, Long])
-    final case class Subscribe(topics: Set[String], listener: ConsumerRebalanceListener)
-    final case class SubscribePattern(pattern: String, listener: ConsumerRebalanceListener)
-    final case class RequestMessages(requestId: Int, topics: Set[TopicPartition])
-    case object Stop
-    final case class Commit(offsets: Map[TopicPartition, Long])
-    //responses
-    final case class Assigned(partition: List[TopicPartition])
-    final case class Revoked(partition: List[TopicPartition])
-    final case class Messages[K, V](requestId: Int, messages: Iterator[ConsumerRecord[K, V]])
-    final case class Committed(offsets: Map[TopicPartition, OffsetAndMetadata])
-    //internal
-    private[KafkaConsumerActor] final case class Poll[K, V](target: KafkaConsumerActor[K, V]) extends DeadLetterSuppression
-    private val number = new AtomicInteger()
-    def nextNumber() = {
-      number.incrementAndGet()
-    }
-
-    private[KafkaConsumerActor] class NoPollResult extends RuntimeException with NoStackTrace
-  }
 
   private[kafka] def rebalanceListener(onAssign: Iterable[TopicPartition] => Unit, onRevoke: Iterable[TopicPartition] => Unit) = new ConsumerRebalanceListener {
     override def onPartitionsAssigned(partitions: util.Collection[TopicPartition]): Unit = {
       onAssign(partitions.asScala)
     }
+
     override def onPartitionsRevoked(partitions: util.Collection[TopicPartition]): Unit = {
       onRevoke(partitions.asScala)
     }
   }
+
+  case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
 
   private class WrappedAutoPausedListener(client: KafkaConsumer[_, _], listener: ConsumerRebalanceListener) extends ConsumerRebalanceListener {
     override def onPartitionsAssigned(partitions: util.Collection[TopicPartition]): Unit = {
@@ -66,6 +44,42 @@ object KafkaConsumerActor {
       listener.onPartitionsRevoked(partitions)
     }
   }
+
+  private[kafka] object Internal {
+
+    private val number = new AtomicInteger()
+
+    def nextNumber() = {
+      number.incrementAndGet()
+    }
+
+    //requests
+    final case class Assign(tps: Set[TopicPartition])
+    final case class AssignWithOffset(tps: Map[TopicPartition, Long])
+    final case class Subscribe(topics: Set[String], listener: ConsumerRebalanceListener)
+    final case class SubscribePattern(pattern: String, listener: ConsumerRebalanceListener)
+    final case class RequestMessages(requestId: Int, topics: Set[TopicPartition])
+
+    final case class PauseMessages(requestId: Int, topics: Set[TopicPartition])
+
+    final case class UnpauseMessages(requestId: Int, topics: Set[TopicPartition])
+
+    final case class Commit(offsets: Map[TopicPartition, Long])
+
+    //responses
+    final case class Assigned(partition: List[TopicPartition])
+    final case class Revoked(partition: List[TopicPartition])
+    final case class Messages[K, V](requestId: Int, messages: Iterator[ConsumerRecord[K, V]])
+    final case class Committed(offsets: Map[TopicPartition, OffsetAndMetadata])
+    //internal
+    private[KafkaConsumerActor] final case class Poll[K, V](target: KafkaConsumerActor[K, V]) extends DeadLetterSuppression
+
+    private[KafkaConsumerActor] class NoPollResult extends RuntimeException with NoStackTrace
+
+    case object Stop
+
+  }
+
 }
 
 private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
@@ -74,22 +88,24 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
   import KafkaConsumerActor._
 
   val pollMsg = Poll(this)
-  def pollTimeout() = settings.pollTimeout
-  def pollInterval() = settings.pollInterval
-
-  var currentPollTask: Cancellable = _
-
   var requests = Map.empty[ActorRef, RequestMessages]
   var consumer: KafkaConsumer[K, V] = _
+
+  var currentPollTask: Cancellable = _
   var commitsInProgress = 0
   var wakeups = 0
   var stopInProgress = false
+
+  def pollTimeout() = settings.pollTimeout
+
+  def pollInterval() = settings.pollInterval
 
   def receive: Receive = LoggingReceive {
     case Assign(tps) =>
       checkOverlappingRequests("Assign", sender(), tps)
       val previousAssigned = consumer.assignment()
       consumer.assign((tps.toSeq ++ previousAssigned.asScala).asJava)
+
     case AssignWithOffset(tps) =>
       checkOverlappingRequests("AssignWithOffset", sender(), tps.keySet)
       val previousAssigned = consumer.assignment()
@@ -97,6 +113,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       tps.foreach {
         case (tp, offset) => consumer.seek(tp, offset)
       }
+
     case Commit(offsets) =>
       val commitMap = offsets.mapValues(new OffsetAndMetadata(_))
       val reply = sender()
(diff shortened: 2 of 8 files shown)


core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -120,34 +108,59 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
             case (_, msg: KafkaConsumerActor.Internal.Messages[K, V]) =>
-              requested = false
-              // do not use simple ++ because of https://issues.scala-lang.org/browse/SI-9766
-              if (buffer.hasNext) {
-                buffer = buffer ++ msg.messages
-              }
-              else {
-                buffer = msg.messages
-              }
+              //              requested = false
```
kciesielski: Can this comment be deleted?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -289,7 +319,6 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
           }
           if (messages.nonEmpty) {
             ref ! Messages(req.requestId, messages)
-            requests -= ref
```
kciesielski: Is there no need to remove the request after sending back a response?

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -120,34 +108,59 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
+
+              buffer.enqueue(msg.messages.toSeq: _*)
+              checkBufferAndPause()
               pump()
             case (_, Terminated(ref)) if ref == consumer =>
               failStage(new Exception("Consumer actor terminated"))
           }
+          consumer.tell(requestMessages, self.ref)
```
kciesielski: Could you explain a bit more what's the flow now? Previously `pump()` was repeatedly sending `requestMessages` to the consumer actor. Now this message is sent  in `preStart` and I can't see where else.

13h3r:
Could you please describe your motivation and solution? Are there any limitations comparing to the previous implementation?

agadek:
Did small renaming, now should be more clear. 

Lately was playing with reactive-kafka. I put a lot of attention on message delivery latency. I've concentrated on committable subSources as our use case looks after solution that helps avoid blocking in case of partial system failure. (in case of one partition downstream fails, others can run while it recovers).  

My proposition resolve major problem subsources partitioning usage. For multiply subsources latency increase in linear proportion to number of partition. All of them are requesting to single consumer. As each subsource sends request for new bench of messages - waiting occurs. Currently each request results in poll. Practically each subsource polls independently (thats blocking for others). What is more some polls actually goes empty(!) without any request in queue (i now its needed as heartbeat, but still ineffective). 

My proposition is to change model from pulling to pushing with backpressure strategy. Stream sources got buffer with adoptable size buffer (u can adopt accordingly to your setup memory and throughput). This way downstream always has messages ready to pull. Buffer overflown situation results with sending pause/unpause messages to consumer (backpressure strategy). 

Consequences: subsources needs to be observed, in case of failure assign partitions are paused and seek to last committed offset called. 

Funny think setting poll interval bigger then 0ms adds additional 30-40ms to each msg latency (in case of my mac)

please look at it, needs testing in other use cases and maybe on scaled up machine. 
some discussion over when should happen pausing/unpausing would be also helpfull
what you think?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
agadek: requests now become subscription, those are valid for life time of down stream

13h3r:
Sorry, right now I have no time to perform detailed review, but have a quick question: why this is better than simple `buffer` stage right after the source?

agadek:
with stream buffer sources will still need to sends request for next messages batch, waiting still occurs.

13h3r:
> with stream buffer sources will still need to sends request for next messages batch, waiting still occurs.
Why? Lets say we have an internal source buffer for 1k messages and explicit `buffer` stage for 1k elements. We still have 1k message for downstream available. In case of very quick consumption you still need to fetch additional messages. What is the difference? Am I miss something?

I think we should agree on the idea first. WDYT @kciesielski @patriknw @ktoso ?

drewhk:
I think we should measure. If buffer achieves the same result, then there is no reason for change.

agadek:
> I think we should measure. If buffer achieves the same result, then there is no reason for change.

For my case latency is the key, for multiple sources this way has better numbers. Thats why our project stays with plain consumer.

13h3r:
@agadek still can not get why buffer will be slower. You introduce an internal buffer which fetch some additional messages to be ready to push it downstream immediately. But introducing `buffer` should do the same.

Are two stages against one is an issue?

kciesielski:
@agadek can you tell more about your benchmarks?

agadek:
> Why? Lets say we have an internal source buffer for 1k messages and explicit buffer stage for 1k elements. We still have 1k message for downstream available. In case of very quick consumption you still need to fetch additional messages. What is the difference? t Am I miss something?

You are right in case of single source. When single source sends request, it results in instant call of poll method. Quick and responsive, my change won't beat performance of this solution. However this approach works well for single source, but if you consider multiple sub-sources things are not performing that much. As I wrote earlier each request results in call of poll method. Each request contains only one partition so polls are made sequentially for each partition independently. More partition more waiting O(n). Stream buffering wont help for this issue. 

> till can not get why buffer will be slower
It depends what slower means for you. My case requires reducing request processing as much as possible. Current implementation in my opinion is single source oriented. Sub-sources are thing that need some change. 

@kciesielski 
Benchmark contains: 
- generator, simple timestamps pubished on kafka topic
- consumer, reading msg, counting throughput and time delta

Downstream processes msgs asynchronously, commits are batched afterwords (groupedWithin). 
What you want to know more? 



Maybe building separated strategies for single and multipleosurces are the think.

13h3r:
> Stream buffering wont help for this issue.

What if you put `buffer` in `SubFlow`? In this case you will have per partition buffer. The only issue with this is poll per partition, but I think it will not be a problem together with the buffer stage.

agadek:
> What if you put buffer in SubFlow? In this case you will have per partition buffer. The only issue with this is poll per partition, but I think it will not be a problem together with the buffer stage.

If it is a problem or not, depends on your latency tolerance and number of sub-sources. if single poll takes 1ms and you have 50 partitions, last sub-source have to wait at least 50ms. For my 50ms is like twice as much as whole request processing should take.

core/src/main/resources/reference.conf
```diff
@@ -35,7 +35,11 @@ akka.kafka.consumer {
   # Note that non-zero value means that blocking of the thread that
   # is executing the stage will be blocked.
   poll-timeout = 50ms
-  
+
+  # sorces buffer size
```
analytically: typo

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -74,29 +88,32 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
 
   var currentPollTask: Cancellable = _
-
-  var requests = Map.empty[ActorRef, RequestMessages]
-  var consumer: KafkaConsumer[K, V] = _
   var commitsInProgress = 0
   var wakeups = 0
   var stopInProgress = false
 
+  def pollTimeout() = settings.pollTimeout
```
ktoso: val is not necessarily better though, def is fine.
val would be worth checking if we still have space in the object without further extending its allocated size (due to alignment), and the def will likely be inlined anyway to that val so I would not worry about it unless proven to have impact

pkolodziejczyk-soi: ok, inlined def being exactly val from settings seems reasonable
thanks

patriknw: `import settings.pollTimeout`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -109,48 +126,65 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
           else reply ! Committed(offsets.asScala.toMap)
         }
       })
-      //right now we can not store commits in consumer - https://issues.apache.org/jira/browse/KAFKA-3412
-      poll()
```
patriknw: not needed any more?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -109,48 +126,65 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
-      if (target == this) {
-        poll()
-        currentPollTask = schedulePollTask()
-      }
+      if (target == this) poll()
       else {
         // Message was enqueued before a restart - can be ignored
         log.debug("Ignoring Poll message with stale target ref")
       }
+      schedulePollTask()
```
patriknw: why this change? there is a purpose with the `target`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -109,48 +126,65 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
     case Stop =>
-      if (commitsInProgress == 0) {
-        context.stop(self)
-      }
-      else {
-        stopInProgress = true
-        context.become(stopping)
-      }
+      stopInProgress = true
+      context.become(stopping)
```
patriknw: why change?

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -12,22 +12,46 @@ import akka.stream.stage.{GraphStageLogic, OutHandler}
...
 
+  val bufferSize = settings.sourceBufferSize
+  val buffer: mutable.Queue[ConsumerRecord[K, V]] = mutable.Queue()
+  var paused = false
+  val pauseMessage = KafkaConsumerActor.Internal.PauseMessage(0, tps)
+  val unpauseMessage = KafkaConsumerActor.Internal.UnpauseMessage(0, tps)
+
+  def checkBufferAndPause(): Unit =
+    if (!paused && buffer.size >= bufferSize) {
+      consumer.tell(pauseMessage, self.ref)
```
patriknw: so it's not a strict max buffer size, the consumer can continue pushing

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -109,48 +126,65 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
         log.debug("Ignoring Poll message with stale target ref")
       }
+      schedulePollTask()
 
-    case req: RequestMessages =>
+    case sub: SubscriptionMessage =>
       context.watch(sender())
-      checkOverlappingRequests("RequestMessages", sender(), req.topics)
-      requests = requests.updated(sender(), req)
-      poll()
```
patriknw: Wouldn't a simple and efficient solution be to not perform the `poll` immediately and instead send an `ExtraPoll` message to self that will do the `poll`. In that way additional `RequestMessages` that are pending in the mailbox will be received first and the `poll` will process all of them. A short delay could also be used to collect more `RequestMessages` before the first `poll`.

It could even be adaptive so that it still performs it immediately if there is only one known subscriber.

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
agadek: just a trigger to send pause, buffer can overflow but no harm

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
agadek: i was experimenting with solution basing on polls executed only with set interval, lag still was present. Im assuming that in your approach would also, but just guessing. Worth to check

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
agadek: fair enough

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
agadek: is it necessary to poll after commit event if there is no request in buffer ?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
agadek: good point, may lead to overlapping scheduled polls

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
agadek: cant tell right now, so worth to revert ;-)

agadek:
> to be honest I don't know if this will break anything else
unrelated refactoring doesn't make it easier to review

real performance test would help evaluate if it works 

treat this PR as an feedback to my rective-kafka experience and reason why i wont use it in current version. probably there is a better solution for latency issue.

graphex:
This is a pretty painful blocker for me too, as I am stuck with consuming from almost 1000 partitions. KStreams works really fast but is locked to a single kafka cluster to consume and produce from. Akka streams is such a great API I would really like to see the kafka consumer perform better.

patriknw:
@agadek @graphex Could one of you provide an example/test (code) that illustrates the scenario that you would like to improve. Then I can turn that into a benchmark and see what the bottleneck is for myself.

agadek:
not sure if a still have that, its some time since I was on it. I'll try to dig it out, give me day or two, please

agadek:
@patriknw primitive simple banchmark you can find here https://github.com/agadek/reactiveKafkaBenchmark
To get comperison you will need kafka with topic with setup for ex. 100 partitions and local build of my PR.
You should be able to see difference between latency in both cases.

patriknw:
Thanks @agadek 
I will take it for a spin today.

patriknw:
@agadek Here is my take on this: https://github.com/akka/reactive-kafka/pull/293
Your benchmark was a good starting point, but I did some [adjustments](https://github.com/agadek/reactiveKafkaBenchmark/pull/1).

agadek:
@patriknw as I said simple as it could be

agadek:
@patriknw  PR is ready to be closed, what do you think ?

patriknw:
superseded by #293

PR #270, opened 2017-02-08
Fix commit batching from more than one stage

fdimuccio:
Hello,

I hit this particular bug: I have two Consumer.committableSource that are merged in one flow and at the end of the stream I commit messages using a single CommittableOffsetBatch, but only messages from one stage (belonging to one groupId) are committed. 

I did some investigations and it turns out that is not enough to pass the full offsets map to one single KafkaConsumerActor, it will commit the offsets only for one groupId (org.apache.kafka.clients.consumer.internals.ConsumerCoordinator: 602) skipping the others. There is also a side effect with the current implementation: it could commit offsets belonging to a topic for which a given groupId wasn't subscribed.

My patch fixes both issues but what bothers me is the global ExecutionContext that I had to use. I couldn't find a better way to import an execution context without breaking the API, I'm open to any suggestion to improve it.

I hope it helps and that I didn't miss anything.


typesafehub-validator:
Hi @fdimuccio,

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -190,7 +192,8 @@ private[kafka] object ConsumerStage {
       if (offsets.isEmpty)
         Future.successful(Done)
       else {
-        stages.head._2.commit(this)
+        implicit val ec = akka.dispatch.ExecutionContexts.global()
+        Future.sequence(stages.map(_._2.commit(this))).map(_ => Done)
```
kciesielski: As far as I understand, the `sequence` call does a serie of `flatMap`s on futures and concatenates results into a sequence, which is a safe operation to be handled by the global EC. Can someone verify this? :)

patriknw: `ExecutionContext.global` should not be used, we should stay in Akka EC land. Hmm, I think I did this in one of the earlier implementations, but it might have been lost in some refactoring. I'll try to find it.

patriknw: Perhaps this could still do `stages.head._2.commit(this)` and then the implementation of the `Committer` can delegate to the rest of the stages, since there I think we have `materializer.executionContext` or similar.

fdimuccio: It can be done if `CommittableOffsetBatch` exposes the stages alongside the offsets, and once done the Committer doesn't have to be a `KafkaAsyncConsumerCommitterRef` anymore since the owned `ActorRef` is redundant.

kciesielski: @fdimuccio Could you code this suggestion?

fdimuccio: Sure, but I had to close this PR because by mistake I made it from the master branch. Before posting a new one I'll create the issue and post the proposed solution there.

patriknw:
Ping me when there is a new PR, I'm interested in that we fix this.

fdimuccio:
I just opened a new PR: https://github.com/akka/reactive-kafka/pull/285

PR #273, opened 2017-02-24
Issue 272: Multiple Message CommitBatch Update and No Offset Result Conversion

dbcfd:
https://github.com/akka/reactive-kafka/issues/272

Provide method to update commitbatch with multiple values (e.g. when grouping kafka messages), and provide a version of the KafkaConsumerActor that does not convert the commit result to a map.

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerMessage.scala
@@ -90,6 +90,13 @@ object ConsumerMessage {
     def updated(offset: CommittableOffset): CommittableOffsetBatch
 
     /**
+     * Add/overwite an offset position for the given groupId, topic, partition.
+     * @param offsets Sequence of offsets to add/overwrite offset position for the given groupId, topic, partition.
+     * @return
+     */
+    def updated(offsets: Seq[CommittableOffset]): CommittableOffsetBatch
+
+    /**
      * Scala API: Get current offset positions
      */
     def offsets(): Map[GroupTopicPartition, Long]
--- a/core/src/main/scala/akka/kafka/ConsumerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerMessage.scala
@@ -11,6 +11,7 @@ import akka.Done
 import akka.kafka.internal.ConsumerStage.CommittableOffsetBatchImpl
 import org.apache.kafka.clients.consumer.ConsumerRecord
 
+import scala.collection.immutable
 import scala.concurrent.Future
 
 /**
@@ -81,7 +82,7 @@ object ConsumerMessage {
   /**
    * For improved efficiency it is good to aggregate several [[CommittableOffset]],
    * using this class, before [[Committable#commitScaladsl committing]] them. Start with
-   * the [[CommittableOffsetBatch$#empty empty] batch.
+   * the [[CommittableOffsetBatch#empty empty]] batch.
    */
   trait CommittableOffsetBatch extends Committable {
     /**
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -7,21 +7,36 @@ package akka.kafka
 import java.util
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.regex.Pattern
+
 import akka.actor.{Actor, ActorLogging, ActorRef, Cancellable, Props, Status, Terminated}
 import akka.event.LoggingReceive
 import org.apache.kafka.clients.consumer._
 import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.errors.WakeupException
+
 import scala.collection.JavaConverters._
 import java.util.concurrent.locks.LockSupport
+
+import akka.Done
 import akka.actor.DeadLetterSuppression
+
 import scala.util.control.{NoStackTrace, NonFatal}
 import scala.util.{Failure, Success, Try}
 
 object KafkaConsumerActor {
   case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
-  def props[K, V](settings: ConsumerSettings[K, V]): Props =
-    Props(new KafkaConsumerActor(settings)).withDispatcher(settings.dispatcher)
+
+  def propsWithCommitResult[K, V](
+    settings: ConsumerSettings[K, V]
+  ): Props = {
+    Props(new KafkaConsumerActor(settings, m => Internal.OffsetCommitResult(m.asScala.toMap))).withDispatcher(settings.dispatcher)
+  }
+
+  def props[K, V](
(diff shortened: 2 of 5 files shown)


core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -153,12 +153,11 @@ private[kafka] object ConsumerStage {
...
       extends CommittableOffsetBatch {
 
-    override def updated(committableOffset: CommittableOffset): CommittableOffsetBatch = {
-      val partitionOffset = committableOffset.partitionOffset
-      val key = partitionOffset.key
-
-      val newOffsets = offsets.updated(key, committableOffset.partitionOffset.offset)
-
+    private def updatedStages(
+      existing: Map[String, Committer],
```
patriknw: this is only used with `updatedStages(stages` so the `existing` parameter is not needed (`stages` can be used)

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -167,14 +166,37 @@ private[kafka] object ConsumerStage {
...
+
+      val newOffsets = offsets.updated(key, committableOffset.partitionOffset.offset)
+
+      new CommittableOffsetBatchImpl(newOffsets, updatedStages(stages, key, committableOffset))
+    }
+
+    override def updated(committableOffsets: Seq[CommittableOffset]): CommittableOffsetBatch = {
+      val (newStages, newOffsets) = committableOffsets.groupBy(_.partitionOffset.key).foldLeft((stages, offsets)) {
+        case ((aggStages, aggOffsets), (key, committableOffsetsForKey)) =>
+          val commitToUse = committableOffsetsForKey.sortBy(_.partitionOffset.offset).head
```
patriknw: this is picking the smallest offset, what is the reason for that?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -167,14 +166,37 @@ private[kafka] object ConsumerStage {
...
+          val partitionOffset = commitToUse.partitionOffset
+          val key = partitionOffset.key
+          val offset = commitToUse.partitionOffset.offset
+
+          val offsetToUpdateTo = Math.max(aggOffsets.getOrElse(key, offset), offset)
```
patriknw: but here it's only using that offset if it's greater than existing in the batch
I'm not sure I understand what this method is intended to do.

patriknw: ScalaDoc describing semantics will be needed.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -97,13 +97,13 @@ private[kafka] object ConsumerStage {
       val offsets = Map(
         new TopicPartition(offset.key.topic, offset.key.partition) -> (offset.offset + 1)
       )
-      (ref ? Commit(offsets)).mapTo[Committed].map(_ => Done)
+      (ref ? Internal.Commit(offsets)).mapTo[Internal.CompletedCommit.type].map(_ => Done)
```
patriknw: how can we know that it's `CompletedCommit` and not `OffsetCommitResult`?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -7,21 +7,36 @@ package akka.kafka
...
 
 object KafkaConsumerActor {
   case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
-  def props[K, V](settings: ConsumerSettings[K, V]): Props =
-    Props(new KafkaConsumerActor(settings)).withDispatcher(settings.dispatcher)
+
+  def propsWithCommitResult[K, V](
+    settings: ConsumerSettings[K, V]
+  ): Props = {
+    Props(new KafkaConsumerActor(settings, m => Internal.OffsetCommitResult(m.asScala.toMap))).withDispatcher(settings.dispatcher)
```
patriknw: I don't understand this, what is the purpose of this `OffsetCommitResult`? I don't see it used anywhere.

core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -90,6 +90,13 @@ object ConsumerMessage {
     def updated(offset: CommittableOffset): CommittableOffsetBatch
 
     /**
+     * Add/overwite an offset position for the given groupId, topic, partition.
+     * @param offsets Sequence of offsets to add/overwrite offset position for the given groupId, topic, partition.
+     * @return
+     */
+    def updated(offsets: Seq[CommittableOffset]): CommittableOffsetBatch
```
dbcfd: Will add that and the corresponding java api.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
dbcfd: Further discussion on your comment about use.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
dbcfd: Think I had originally meant for this to be an aggregation and not use stages, but it makes more sense just to use stages as implemented.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
dbcfd: Error, should be last for largest offset. I would like to add additional tests around this functionality. Best to do a test similar to https://github.com/akka/reactive-kafka/blob/master/core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala#L140?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
dbcfd: Meant to select largest offset both above and here, which should be the same behavior as a single offset.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
dbcfd: I had originally added this to handle cases where the commit result is used, but after adding it, found that no users of this actor seem to use the commit result. Am I missing the use of the commits somewhere? Does it make sense to expose the committed offsets in the api (e.g. commitableSourceWithOffsets) or should this actor just not return the committed offsets?

ennru:
I looked into this PR now and can see a slight chance that the current code in certain situations could overwrite a higher commit offset with a lower one. So the logic to take the max of the current value in the offsets map and the incoming one could make sense.
For the rest of the PR I don't get enough motivation in this issue description and the code has moved forward a lot since you opened it.
I'll close both the PR and this issue, please open a new one if you have any updates on this.

Issue #229, opened 2016-09-16
Document how to restart a Kafka stream based processing pipeline


drewhk:
I.e. how to wait on previous instance to properly stop and rematerialize the stream, etc.

nafshartous:
I'd like to help and have a few questions.  How does one build and preview the docs ?

patriknw:
@nafshartous Sorry for embarrassing long delay of this reply. Use `sbt paradox`, and then you can open `docs/target/paradox/site/index.html` in browser.

nafshartous:
Does something like this capture the intent ?

```
//restart kafka consumer
object RestartKafkaConsumeExample extends ConsumerExample {
  def main(args: Array[String]): Unit = {
    // #restartConsumer
    def createAndRun(): Consumer.Control =
      Consumer.committableSource(consumerSettings, Subscriptions.topics("topic1"))
        .toMat(Sink.ignore)(Keep.left).run()

    val control = createAndRun()

    control.shutdown()

    Await.result(control.isShutdown, 5.seconds)

    // Start the stream again
    createAndRun()
    // #restartConsumer
  }
}
```

nafshartous:
Will submit a new PR based on the discussion in

  https://github.com/akka/reactive-kafka/pull/275

jacobg:
Thanks for documenting an approach to restarting the stream. If this approach is attempting to deal with failure in the Kafka consumer source, then I'm wondering why one of approaches documented for handling failures in Akka streams up-stream sources by just restarting the failed stage is not used?
http://doc.akka.io/docs/akka/2.5.4/scala/stream/stream-error.html

Is not sufficient to just restart the Kafka source stage using RestartSource or a supervision strategy that restarts?

Are there cases where an Akka Kafka source's underlying implementation can fail without propagating the failure into the stream?

kciesielski:
@jacobg Looks like `RestartSource` is a pretty new thing, added in Akka 2.5. I think we may need both usages in the example, for Akka 2.4 and 2.5.

kciesielski:
@nafshartous restarting is already documented in https://github.com/akka/reactive-kafka/blob/master/docs/src/main/paradox/errorhandling.md. I'm not sure why it's not visible in the main docs. @patriknw does that require some kind of redeployment to the docs?

jacobg:
[...]

PR #275, opened 2017-02-26
Document how to restart a Kafka stream based processing pipeline #229

nafshartous:
Issue #229 - Document how to restart a Kafka stream based processing pipeline

### DIFF
--- a/docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
+++ b/docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
@@ -347,7 +347,8 @@ object RestartKafkaConsumeExample extends ConsumerExample {
     // #restartConsumer
     val runnableGraph: RunnableGraph[Consumer.Control] =
       Consumer.committableSource(consumerSettings, Subscriptions.topics("topic1"))
-        .map { message => { println(message.record.value()); message } }
+        .map(msg => { println(msg.record.value()); msg.committableOffset })
+        .map(_.commitScaladsl())
         .toMat(Sink.ignore)(Keep.left)
     val control1: Consumer.Control = runnableGraph.run()
     val shutdownFuture = control1.shutdown()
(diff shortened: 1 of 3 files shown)


docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
```diff
@@ -341,3 +341,29 @@ object ExternallyControlledKafkaConsumer extends ConsumerExample {
...
+        .toMat(Sink.ignore)(Keep.left)
+    val control1: Consumer.Control = runnableGraph.run()
+    val shutdownFuture = control1.shutdown()
+
+    shutdownFuture onComplete {
+      case Success(Done) =>
+        println("Stream was shutdown")
+
+        // Start the stream again
+        val control2: Consumer.Control = runnableGraph.run()
```
kciesielski: I am not sure if this is something we'd recommend. How would you return this `control2` object for further error handling? Maybe it would be better to manage the graph in a an actor and make it self-restart when the stream gets shut down or fails?

nafshartous: @kciesielski  Yes the example is not realistic.  How about an actor with two states (Processing and NotProcessing).  In the NotProcessing state when receiving a Start message the state changes to Processing and the stream is started.  Conversely in the Processing state, receiving the Stop message results in stopping the stream and transitioning to the NotProcessing state.  

The two states could represent whether or not a downstream dependency is available to receive data.

kciesielski: @nafshartous Sounds good. It would also be good to include `future.onFailure` with handling that restarts the whole actor or the whole processing somehow (resend `Start` to itself?).
This is a very frequent question - how to deal with failing stream so souch example would be great.

nafshartous: Closing and will submit a new PR based on the above discussion.

PR #283, opened 2017-03-24
Fix shared kafka consumer in 0.14

yurikpanic:
Any Consumer.plainExternalSource will fail with "java.lang.Exception: Consumer actor terminated" if it is created after KafkaConsumerActor manages to run pull().

E.g. original example for sharing KafkaConsumer from documentation might run successfully, but it will fail if we insert Thread.sleep(1000L) between actor creation and stream materialization.

### DIFF
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -87,12 +87,12 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
 
   def receive: Receive = LoggingReceive {
     case Assign(tps) =>
-      trySchedulePollTask()
+      scheduleFirstPollTask()
       checkOverlappingRequests("Assign", sender(), tps)
       val previousAssigned = consumer.assignment()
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @yurikpanic,

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -194,6 +197,9 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
     super.postStop()
   }
 
+  def trySchedulePollTask(): Unit =
```
patriknw: I'd find it easier to understand if this was named `scheduleFirstPollTask`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -178,7 +182,6 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
     super.preStart()
 
     consumer = settings.createKafkaConsumer()
-    currentPollTask = schedulePollTask()
```
patriknw: ok, so it's deferring the first poll until after the first Subscribe (or similar) message

Issue #284, opened 2017-03-26
Fix commit batching from more than one stage


fdimuccio:
Hello,

I hit this particular bug: I have two Consumer.committableSource that are merged in one flow and at the end of the stream I commit messages using a single CommittableOffsetBatch, but only messages from one stage (belonging to one groupId) are committed.

fdimuccio:
Following the discussion of PR [https://github.com/akka/reactive-kafka/pull/270](https://github.com/akka/reactive-kafka/pull/270) I changed the code to avoid the usage of the global ExecutionContext, here is the commit: [https://github.com/fdimuccio/reactive-kafka/commit/79e4fc975e5af4f8f8178fc4eaa6fb772bbf2516](https://github.com/fdimuccio/reactive-kafka/commit/79e4fc975e5af4f8f8178fc4eaa6fb772bbf2516).

The only drawback (that I see) of this solution is that the `KafkaAsyncConsumerCommitterRef` is tied to `CommittableOffsetBatchImpl`. To decouple them the `CommittableOffsetBatch` trait should be changed to return also the `Committer` for each `GroupPartitionTopic`, but since the `Committer` is an internal API I didn't want to go too far with the changes.

PR #285, opened 2017-04-04
Issue 284: Fix commit batching from more than one stage

fdimuccio:
https://github.com/akka/reactive-kafka/issues/284

Change `KafkaAsyncConsumerCommitterRef` to send the `PartitionOffset` to the right stage when batch committing.

One drawback of this solution is that the `KafkaAsyncConsumerCommitterRef` is tied to `CommittableOffsetBatchImpl`. To decouple them the `CommittableOffsetBatch` trait should be changed to return also the `Committer` for each `GroupPartitionTopic`, but since the `Committer` is part of the internal API I didn't want to make it public by exposing it.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
@@ -22,6 +22,7 @@ import org.apache.kafka.common.TopicPartition
 
 import scala.compat.java8.FutureConverters.FutureOps
 import scala.collection.JavaConverters._
+import scala.collection.immutable
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future, Promise}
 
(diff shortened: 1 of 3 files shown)


kciesielski:
@fdimuccio Thanks for your effort :) I think it's the best solution so far. I think it's OK to merge. @patriknw can you give it a look?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -141,12 +156,13 @@ private[kafka] object ConsumerStage {
...
       extends CommittableOffset {
-    override def commitScaladsl(): Future[Done] = committer.commit(partitionOffset)
+    override def commitScaladsl(): Future[Done] = committer.commit(Seq(partitionOffset))
     override def commitJavadsl(): CompletionStage[Done] = commitScaladsl().toJava
   }
 
   trait Committer {
-    def commit(offset: PartitionOffset): Future[Done]
+    // Commit all offsets (of different topics) belonging to the same stage
+    def commit(offsets: Seq[PartitionOffset]): Future[Done]
```
patriknw: `Seq` can be mutable, e.g. `Array`, therefore we explicitly use the `immutable` one by:
```scala
import scala.collection.immutable

def commit(offsets: immutable.Seq[PartitionOffset]): Future[Done]
```

fdimuccio:
@kciesielski you are welcome. 

If everything is ok, tell me if I have to squash those commits.

kciesielski:
@fdimuccio LGTM, I guess @patriknw can approve the changes and we're good to merge.

patriknw:
Refs #284

PR #287, opened 2017-04-13
Fixes for Akka 2.5 compatibility

kciesielski:
This PR removes dependency on akka-stream-contrib and fixes some callback handling to make the library work well with Akka 2.5.0.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,8 +9,7 @@ val kafkaVersion = "0.10.2.0"
 val kafkaClients = "org.apache.kafka" % "kafka-clients" % kafkaVersion
 
 val commonDependencies = Seq(
-  "com.typesafe.akka" %% "akka-testkit" % akkaVersion % Test,
-  "com.typesafe.akka" %% "akka-stream-contrib" % "0.6" % Test
+  "com.typesafe.akka" %% "akka-testkit" % akkaVersion % Test
 )
 
 val coreDependencies = Seq(
--- a/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
@@ -14,7 +14,7 @@ import akka.kafka.scaladsl.Consumer._
 import akka.kafka.{javadsl, scaladsl, _}
 import akka.stream._
 import akka.stream.scaladsl.Source
-import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue}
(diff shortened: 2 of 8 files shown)


core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -211,8 +211,13 @@ private[kafka] trait MessageBuilder[K, V, Msg] {
...
 
-private[kafka] trait PromiseControl extends Control {
-  this: GraphStageLogic =>
+private[kafka] sealed trait ControlOperation
+
+private[kafka] case object ControlStop extends ControlOperation
+
+private[kafka] case object ControlShutdown extends ControlOperation
+
+private[kafka] trait PromiseControl extends GraphStageLogic with Control with CallbackWrapper[ControlOperation] {
```
johanandren: This is not just 2.5.0 updates though?

kciesielski: This is related because Akka 2.5.0 complains that `java.lang.IllegalStateException: not yet initialized: only setHandler is allowed in GraphStageLogic constructor` in some tests. Indeed, we expose callback to the outside, so if someone stops the stage before it gets fully initialized, it may throw such error. I guess 2.4.x wasn't checking this to such extent?

johanandren: Ah, yes, it is more verbal when there are graph stage issues now.

build.sbt
```diff
@@ -3,14 +3,13 @@ import de.heikoseeberger.sbtheader.HeaderPattern
 
 name := "akka-stream-kafka"
 
-val akkaVersion = "2.4.17"
+val akkaVersion = "2.5.0"
```
patriknw: This is not possible. Akka being backwards compatible means that libraries built with Akka 2.4.x can be used in applications running Akka 2.5.x, but not the opposite.

I haven't looked at what the reason was why akka-stream-kafka could not be used with Akka 2.5.0, but I would guess we accidentally used some internal APIs in akka-stream-kafka. We should try to fix that instead, and release a new version, built with 2.4.17, that can be used with 2.5.0.

kciesielski: @patriknw So if we change 2.5.0 back to 2.4.17 in this PR then this version will be working well with 2.5.0, right?

patriknw: yes, it's supposed to work ;-)

kciesielski: Ok, I switched back to 2.4 and changed PR name/description.

kciesielski:
Is this OK to merge?

Issue #290, opened 2017-04-24
Sub source of plainPartitionedSource doesn't complete when its partition gets revoked


elkozmon:
Hi,

I am experiencing an issue when auto-rebalance kicks in and my plainPartitionedSource consumer "loses" some partitions, their sub sources don't complete as they are supposed to. Later with the help of the debugger I learned that when the `shutdown()` method is called upon the SubSourceStage, its control callback is not initialized and therefore won't shutdown. SubSourceStage lacks superclass `preStart()` method call which initializes this control callback, so I added it and it seemed to do the trick.

patriknw:
@elkozmon could you create a pull request?

elkozmon:
@patriknw [sure](https://github.com/akka/reactive-kafka/pull/294)

PR #294, opened 2017-04-26
Issue 290:  Sub source of plainPartitionedSource doesn't complete when its partition gets revoked

elkozmon:
[#290](https://github.com/akka/reactive-kafka/issues/290)

* Sub sources of plainPartitionedSource didn't shutdown
  when their partitions got revoked due to control callback
  not being initialized.
* By adding superclass preStart() method call in SubSourceStage
  the control callback initalizes properly.


patriknw:
verified that CLA has been signed, don't know why the automatic check didn't run

PR #296, opened 2017-04-26
Document error handling patterns

kciesielski:


docs/src/test/java/sample/javadsl/StreamWrapperActor.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.kafka.common.serialization.StringDeserializer;
+import scala.concurrent.duration.Duration;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+
+public class StreamWrapperActor extends UntypedActor {
+
+    final ActorSystem system = ActorSystem.create("example");
```
patriknw: creating a new ActorSystem from the Actor :)

docs/src/test/java/sample/javadsl/StreamWrapperActor.java
```diff
@@ -0,0 +1,80 @@
...
+
+    protected final Materializer materializer = ActorMaterializer.create(system);
```
patriknw: this can be `ActorMaterializer.create(getContext())`, which will make the materializer a child of the actor, i.e. have the same lifecycle, otherwise you should stop the materializer in the Actor's `postStop`.

docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
```diff
@@ -341,3 +342,51 @@ object ExternallyControlledKafkaConsumer extends ConsumerExample {
...
+    //#errorHandlingStop
+    val done =
+      Consumer.plainSource(consumerSettings, Subscriptions.topics("topic1"))
+        .mapAsync(1)(processMsg)
+        .runWith(Sink.ignore)
+
+    done.onComplete {
+      case Failure(ex) =>
+        log.error(ex, "Stream failed, stopping the actor.")
+        context.stop(self)
```
patriknw: never use `context` from another thread like that, you can only send a message from here

kciesielski: Thanks, I was explaining the same to someone and decided to write this docs, then copied the bad code here :)

kciesielski:
@patriknw thanks for the review, fixes applied.

docs/src/test/java/sample/javadsl/StreamWrapperActor.java
```diff
@@ -0,0 +1,80 @@
...
+                    .withGroupId("group1")
+                    .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+
+    public void onReceive(Object message) throws Exception {
+        unhandled(message);
+    }
+
+    CompletionStage<String> process(ConsumerRecord<byte[], String> msg) {
+        return CompletableFuture.completedFuture("processed");
+    }
```
johanandren: Calling a method of the actor from another thread here, which is unsafe, make it static or turn it into a message exchange with ask (the latter probably better shows a good example).

docs/src/test/java/sample/javadsl/StreamWrapperActor.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+
+public class StreamWrapperActor extends UntypedActor {
+
+    final ActorSystem system = getContext().system();
+        
+    protected final Materializer materializer = ActorMaterializer.create(getContext());
```
patriknw: it shouldn't leak, with your latest fixes Johan, should it?
A more graceful stop in postStop might still be better because of the scary logging that will happen when abruptly stopping the materializer (or it's actor)

johanandren: Ah, it's created with the `context` not the `context.system`, yes then it will not leak, my bad! Just the scary error message then, which may not matter as much for an example I guess.

docs/src/test/java/sample/javadsl/StreamWrapperActor.java
```diff
@@ -0,0 +1,80 @@
...
+import akka.pattern.BackoffSupervisor;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import scala.concurrent.duration.Duration;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+
+public class StreamWrapperActor extends UntypedActor {
```
kciesielski: We're still on Akka 2.4, with 2.5 compatibility implemented in #287

Jayasagar:
I was trying to use the StreamWrapperActor approach to build resilient reactive Kafka consumer and it works fine with the linear stream as shown in sample example (docs/src/test/java/sample/javadsl/StreamWrapperActor.java )

However, we are unable to make it work with Closed Runnable Graphs. Is anyone successful to handle 'consumer fails due to Wakeup exception' using the closed runnable graphs? One of the graphs looks like 
CommitableSource -> Partition ->(0) -> Flow -> Sink.ignore
                                                      ->(1) -> Partition  Flow -> (0)-> Flow -> Sink.ignore
                                                                                             -> (1) -> Flow-> CommitableProducerSink

From our understanding, right now, the streams supervisor strategy does not catch exceptions from certain stages/operators from closed graph, i.e. supervision strategy only works with linear stream. 

We have tried watchTermination/WithsupervisionStrategy on Source and WithsupervisionStrategy on RunnableGrapgh, but it never calls.

Some References:
http://stackoverflow.com/questions/35631754/why-is-akka-streams-swallowing-my-exceptions
https://github.com/akka/akka/issues/19950

Jayasagar:
I was trying to use the StreamWrapperActor approach to build resilient reactive Kafka consumer and it works fine with the linear stream as shown in sample example (docs/src/test/java/sample/javadsl/StreamWrapperActor.java )

However, we are unable to make it work with Closed Runnable Graphs. Is anyone successful to handle 'consumer fails due to Wakeup exception' using the closed runnable graphs? One of the graphs looks like
CommitableSource -> Partition ->(0) -> Flow -> Sink.ignore
->(1) -> Partition Flow -> (0)-> Flow -> Sink.ignore
-> (1) -> Flow-> CommitableProducerSink

From our understanding, right now, the streams supervisor strategy does not catch exceptions from certain stages/operators from closed graph, i.e. supervision strategy only works with linear stream.

We have tried watchTermination/WithsupervisionStrategy on Source and WithsupervisionStrategy on RunnableGrapgh, but it never calls.

Some References:
http://stackoverflow.com/questions/35631754/why-is-akka-streams-swallowing-my-exceptions
akka/akka#19950

docs/src/test/java/sample/javadsl/StreamWrapperActor.java
kciesielski: I updated the examples to use `ask`.

kciesielski:
@Jayasagar Are you sure you are keeping the correct materialized value? Just like in this answer: https://stackoverflow.com/questions/35631754/why-is-akka-streams-swallowing-my-exceptions#comment58993826_35632148 which points out to important distinction between `Keep.right`, `Keep.left` and `Keep.both`.

Jayasagar:
@kciesielski I am not sure is it possible to do on ClosedGraph. On RunnableGraph, there is only one method available is 'run'. 

Here is how my code looks like

// Materialize
`ActorMaterializer materializer = ActorMaterializer.create(ActorMaterializerSettings
                .create(actorSystem)
                .withDebugLogging(true)
                .withSupervisionStrategy(new TestDecider()), actorSystem);
                
Graph<ClosedShape, NotUsed> closedShapeGraph = GraphDSL.create(builder -> {
  // Build your graph....
  return ClosedShape.getInstance();
});

// Run graph here
RunnableGraph.fromGraph(closedShapeGraph)
          .withAttributes(ActorAttributes.withSupervisionStrategy(new SelfKillActorDecider()))
          .run(materializer);`

Jayasagar:
I also tried Graph with the stage I want the materialized value of i.e. Source I think. Not working.

```java
Graph<ClosedShape, Consumer.Control> completionStageGraph = GraphDSL.create(source, (builder, sourceShape) -> {
// builder...
}

// Run it
        Consumer.Control completionControl = RunnableGraph.fromGraph(completionStageGraph)
                .run(materializer);

        CompletionStage<Done> completionStage = completionControl.isShutdown();
        completionStage.exceptionally(e -> {
            getSelf().tell(PoisonPill.getInstance(), getSelf());
            return Done.getInstance();
        });
```

Jayasagar:
Solved! trick is to use GraphDSL.create(source, (builder, sourceShape) ,  instead GraphDSL.create(builder) , basically, we need to pass the stage we want the materialized value of , in our case it is Source!
Thanks @kciesielski !!

kciesielski:
@Jayasagar Interesting. Yes, sounds like a case worth documenting. How about we merge this PR and then you'll submit yours? @patriknw wdyt? I think this PR is ready to merge as I covered last change request by @johanandren in 549d2bc.

Jayasagar:
@kciesielski Yep, I will extend the example with a bit more complex stream using Source, Sink(Ignore vs ProducerSink), Flows, Partitions, ClosedGraphs!! Please tag with my name once it is merged :) Thanks.

patriknw:
@Jayasagar merged this, great if you can add more

Jayasagar:
@patriknw I made a pull request https://github.com/akka/reactive-kafka/pull/321 
Can you have a look? Thanks! 
It is about 'Document how to handle consumer fails while using closed graphs'

Thanks.

Issue #267, opened 2017-01-23
Expose offsetsForTimes function from kafkaConsumer


jvvasques:
At Talkdesk we're using Kafka and we need to have access to a kafkaConsumer primitive to get [offsets given a timestamp.](https://kafka.apache.org/0101/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#offsetsForTimes(java.util.Map))

I've discovered this project and I want to integrate it in our Kafka pipeline. Are you open to receive a PR to add support to this Kafka low-level primitive?

kciesielski:
@jvvasques of course, PRs are most welcome :) BTW: does this call to `consumer.offsetsForTimes()` have to be synchronized with `poll` and `commit`?

jvvasques:
I believe no. According to the [code and documentation](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1426) the call to `consumer.offsetsForTimes()` is blocking and it does not interfere with `pool` and `commit`

13h3r:
How do you use such offsets? Are they used once during stream start to setup initial offsets or they are used on regular basis during data processing?

jvvasques:
@13h3r we plan to use those offsets to setup the initial position in kafka when starting the stream

ennru:
Looks like #300 which was released in 0.17 implemented this.
Please re-open if something is missing.

PR #300, opened 2017-05-04
Issue 267: Expose offsetsForTimes function from kafkaConsumer

nsphung:
You can use this to get topic/partitions offsets from a timestamp (i.e Kafka timestamp https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message) to setup the initial position in kafka when starting the stream. This will work without auto-rebalance if you plan to manage your offset manually. 

Discussion on #267

### DIFF
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -26,6 +26,7 @@ object KafkaConsumerActor {
     //requests
     final case class Assign(tps: Set[TopicPartition])
     final case class AssignWithOffset(tps: Map[TopicPartition, Long])
+    final case class AssignOffsetsForTimes(timestampsToSearch: Map[TopicPartition, Long])
     final case class Subscribe(topics: Set[String], listener: ConsumerRebalanceListener)
     final case class SubscribePattern(pattern: String, listener: ConsumerRebalanceListener)
     final case class RequestMessages(requestId: Int, topics: Set[TopicPartition])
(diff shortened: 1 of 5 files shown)


searler:
Note the typo: Assign**e**mentOffsetsForTimes

That particular error appears through out the code.

nsphung:
Thanks. It's fixed

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -103,6 +104,19 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       tps.foreach {
         case (tp, offset) => consumer.seek(tp, offset)
       }
+    case AssignOffsetsForTimes(timestampsToSearch) =>
+      scheduleFirstPollTask()
+      checkOverlappingRequests("AssignOffsetsForTimes", sender(), timestampsToSearch.keySet)
+      val previousAssigned = consumer.assignment()
+      consumer.assign((timestampsToSearch.keys.toSeq ++ previousAssigned.asScala).asJava)
+      val topicPartitionToOffsetAndTimestamp = consumer.offsetsForTimes(timestampsToSearch.mapValues(long2Long(_)).asJava)
```
patriknw: note that `offsetsForTimes` is a blocking call, javadoc is even mentioning: "Notice that this method may block indefinitely if the partition does not exist." We already use a dedicated dispatcher for the `KafkaConsumerActor` so it might be alright, but somewhat scary.

PR #303, opened 2017-05-17
Fix the "At-Least-Once Delivery" link

TimMoore:
Right now, the link on http://doc.akka.io/docs/akka-stream-kafka/current/consumer.html points to http://doc.akka.io/docs/akka-stream-kafka/current/atleastonce.md, which returns 404.

Paradox requires the `@ref:` prefix when linking to a local Markdown document, so that it will rewrite the file extension to `.html` when building the site.

The new page is great. Thanks, @patrick-premont!


kciesielski:
Thanks!

PR #308, opened 2017-05-28
Using tail recursion to collect Kafka client properties

arpanchaudhury:


core/src/main/scala/akka/kafka/internal/ConfigSettings.scala
```diff
@@ -4,32 +4,31 @@
...
+
+import scala.collection.JavaConverters.asScalaSetConverter
 
 /**
  * INTERNAL API
  */
 private[kafka] object ConfigSettings {
 
   def parseKafkaClientsProperties(config: Config): Map[String, String] = {
-    def collectKeys(c: ConfigObject, prefix: String, keys: Set[String]): Set[String] = {
```
patriknw: Did you see the issue for real, or is this just a nice cleanup?
Wouldn't it have to be a really deep properties structure to be an issue.

arpanchaudhury: @patriknw I didn't faced any issue as such but considering the number of properties we can pass to the producers and consumers I refactored it as a nice to have. WDYT?

patriknw: I think it's good to improve it, thanks.

To clarify, for the archives, the old grows stack per prop level, not per property. E.g. aaa.bbb.ccc.ddd.eee.fff is 6 levels.

arpanchaudhury:
@patriknw thank you for the explanation.  updated the PR.

kciesielski:
LGTM, thanks @arpanchaudhury!

Issue #322, opened 2017-07-06
Support for Kafka version 0.11


raboof:
Kafka 0.11.0.0 was released in June, can we support it side-by-side 0.10?

abesanderson:
I'm currently testing this on a fork, and it was pretty straightforward to change.  There were a couple deprecated apis and configs, but there weren't a lot of tests in reactive-kafka which used them.  The main thing to note is that the 0.11.0 client will only support back to version 0.10.0 brokers.  Anything earlier would require an upgrade to brokers before clients can be upgraded.

seglo:
@abesanderson I'm curious if you've pushed your work anywhere.  I'm interested in experimenting with it.  I don't see any changes in your remote GitHub fork.

https://github.com/abesanderson/reactive-kafka

abesanderson:
@seglo It's just on another branch: https://github.com/abesanderson/reactive-kafka/tree/upgrade-kafka-0.11.0.0

seglo:
Of course!  Thanks.

agemooij:
I just read about the new [Record Headers feature](https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers) that was released with 0.11. 

That could be really useful to us. Are there any plans for implementing this new feature in the near future?

seglo:
There's a ton of great stuff in 0.11.0.0.  My favourite is the exactly-once consume->transform->produce transactional support.  I have a little demo available here.

https://github.com/seglo/exactly-once-streams/blob/master/src/main/scala/seglo/apps/ConsumeTransformProduce.scala

I'm thinking about working on a PR for this, unless someone beats me to it.

gabriele83:
Hi, how much time is needed to have the Kafka 0.11 integration?
Thanks

albio:
I need also. Any news?

albertobisello:
I have the same needs. Could you add support to Kafka 11? 
many thanks guys!

ktoso:
Hi @kciesielski, what do you think about this?
Usually we keep 2 versions "alive" / "maintained", so this if supporting 0.11 means enough changes that we can't trivially do so in reactive-kafka 0.11 and it means calling an 0.12 we should likely call 0.10 end-of-life soon. 

- Do you know or have time to investigate how much change there is? 
- Do you think we can stop supporting reactive-kafka 0.10 soon?

ktoso:
[...]

PR #327, opened 2017-07-20
Update Kafka version and ScalaTest-Embedded-Kafka version to support 0.11

jeqo:
I have updated dependencies to support Kafka 0.11.

### DIFF
--- a/core/src/test/scala/akka/kafka/internal/ProducerTest.scala
+++ b/core/src/test/scala/akka/kafka/internal/ProducerTest.scala
@@ -4,12 +4,12 @@
  */
 package akka.kafka.internal
 
+import java.util.Date
 import java.util.concurrent.{CompletableFuture, TimeUnit}
 
 import scala.concurrent.{Await, ExecutionContext, Future, Promise}
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
-
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.kafka.ProducerMessage._
@@ -20,23 +20,15 @@ import akka.stream.scaladsl.Flow
 import akka.stream.testkit.scaladsl.{TestSink, TestSource}
 import akka.testkit.TestKit
 import akka.kafka.test.Utils._
-
-import org.apache.kafka.clients.producer.{
-  Callback,
-  KafkaProducer,
-  ProducerRecord,
-  RecordMetadata
-}
+import org.apache.kafka.clients.producer.{Callback, KafkaProducer, ProducerRecord, RecordMetadata}
 import org.apache.kafka.common.TopicPartition
-import org.apache.kafka.common.record.Record
 import org.apache.kafka.common.serialization.StringSerializer
-
 import org.mockito.Matchers._
-import org.mockito.Mockito, Mockito._
+import org.mockito.Mockito
+import Mockito._
 import org.mockito.invocation.InvocationOnMock
 import org.mockito.stubbing.Answer
 import org.mockito.verification.VerificationMode
-
 import org.scalatest.{FlatSpecLike, Matchers}
 import org.scalatest.BeforeAndAfterAll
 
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:
Hi @jeqo,

core/src/test/scala/akka/kafka/internal/ProducerTest.scala
```diff
@@ -61,7 +53,7 @@ class ProducerTest(_system: ActorSystem)
 
   def recordAndMetadata(seed: Int) = {
     new ProducerRecord("test", seed.toString, seed.toString) ->
-      new RecordMetadata(new TopicPartition("test", seed), seed.toLong, seed.toLong, Record.NO_TIMESTAMP, -1, -1, -1)
+      new RecordMetadata(new TopicPartition("test", seed), seed.toLong, seed.toLong, new Date().getTime, java.lang.Long.valueOf(-1), -1, -1)
```
johanandren: Use -1L instead or if it for some weird reason needs to be manually constructed, cache the instance, Long is immutable.

Also, don't allocate one `Date` for every record, use `System.currentTimeMillis` instead.

johanandren:
Refs #322

core/src/test/scala/akka/kafka/internal/ProducerTest.scala
johanandren: Actually, I wonder if the date should really be set here, see the upgraded fork by @abesanderson https://github.com/abesanderson/reactive-kafka/commit/f378e50fdaf85b6a6fd7f4ee28988620ab018f7a

abesanderson: In 0.11, most of the constants that were in `Record` were moved to `RecordBatch`.

jeqo: Yes, I could update it to your commit referred by @johanandren

jeqo:
@johanandren thanks for your feedback. I just found the discussion in issues. Since 0.10.2 there is a client compatibility that enable clients to be updated keeping older broker versions: https://www.confluent.io/blog/upgrading-apache-kafka-clients-just-got-easier/ Not sure if thats what you're referring in #322  

I think for using Headers this changes are be enough, but to support Exactly Once Semantics, some changes should be added at the Stage level. 

IMO we can work on EOS later, and have 0.11 exposed as a SNAPSHOT

patriknw:
What was the conclusion? Will this still work with older Kafka server?

ijuma:
Kafka 0.11.0 clients support the exact same set of broker versions as Kafka 0.10.2 clients: 0.10.0, 0.10.1, 0.10.2, 0.11.0. Kafka 0.11.0 clients have additional functionality if interacting with 0.11.0 brokers. Does that clarify it?

core/src/test/scala/akka/kafka/internal/ProducerTest.scala
```diff
@@ -54,14 +46,16 @@ class ProducerTest(_system: ActorSystem)
...
+
   type K = String
   type V = String
   type Record = ProducerRecord[K, V]
   type Msg = Message[String, String, NotUsed.type]
 
   def recordAndMetadata(seed: Int) = {
     new ProducerRecord("test", seed.toString, seed.toString) ->
-      new RecordMetadata(new TopicPartition("test", seed), seed.toLong, seed.toLong, Record.NO_TIMESTAMP, -1, -1, -1)
+      new RecordMetadata(new TopicPartition("test", seed), seed.toLong, seed.toLong, RecordBatch.NO_TIMESTAMP, checksum, -1, -1)
```
ijuma: I suggest not relying on the internal `RecordBatch` class. Set your own constant or hardcode it since it's a test.

jeqo: Done :+1:

PR #331, opened 2017-08-03
Fixed issue with partitioned consumer getting in a bad state.

charles-crain:
Fixes https://github.com/akka/reactive-kafka/issues/330

SubSourceLogic had a few issues.  It used a queue to track partitions Kafka had assigned, which were then turned into sources as the downstream demanded them.  This had a few issues:

1. If a partition was revoked before a source was created, the source would be created anyway for the revoked partition.
2. If a partition was revoked and reassigned multiple times before being processed, it could appear in the queue many times.
3. If a partition was revoked (and possibly reassigned) after a source was created but before creation was complete, this source would never be shut down.

In this new code I explicitly track which TopicPartitions are in the "assigned but not started" and "started but not done starting" states.  I am no longer able to reproduce the issue under conditions that would reliably reproduce it before.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
@@ -26,7 +26,10 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
 ) extends GraphStageLogic(shape) with PromiseControl with MessageBuilder[K, V, Msg] {
   var consumer: ActorRef = _
   var self: StageActor = _
-  var buffer: immutable.Queue[TopicPartition] = immutable.Queue.empty
+  // Kafka has notified us that we have these partitions assigned, but we have not created a source for them yet.
+  var pendingPartitions: immutable.Set[TopicPartition] = immutable.Set.empty
+  // We have created a source for these partitions, but it has not started up and is not in subSources yet.
+  var partitionsInStartup: immutable.Set[TopicPartition] = immutable.Set.empty
   var subSources: Map[TopicPartition, Control] = immutable.Map.empty
 
   override def preStart(): Unit = {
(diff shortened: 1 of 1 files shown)


DevinSRose:
Note that this fixes the error where a message like this is printed repeatedly: 

RequestMessages from topic/partition already requested by other stage

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -55,22 +58,35 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
+    partitionsInStartup --= tps
+    tps.flatMap(subSources.get).foreach(_.shutdown())
+    subSources --= tps
   }
 
   val subsourceCancelledCB = getAsyncCallback[TopicPartition] { tp =>
     subSources -= tp
-    buffer :+= tp
+    partitionsInStartup -= tp
+    pendingPartitions += tp
```
johanandren: Why does it go back in pending when it is cancelled?

charles-crain: I was actually just copying what the code did before.  pendingPartitions is effectively the same thing as "buffer" used to be, but it is a Set instead of a Queue to prevent processing the same TopicPartition twice under conditions when partitions are revoked and reassigned quickly.

The old code put the TopicPartition back in buffer in this case, so I figured I should do the same with pendingPartitions to preserve the same behavior.

johanandren: Ah ok.

charles-crain:
@patriknw I gave some thought to how to test this but you are correct it is difficult.  Ultimately this is a timing issue that results from Kafka rapidly assigning and revoking partitions before the substreams we create to serve them can progress through their various states.  If an assigned partition progresses from assigned -> created -> started before it is revoked then everything works fine.  If OTOH it is revoked during either of those transitions, things got bad.  

To test it we'd have to be able to simulate a quick assign-then-revoke while possibly forcing the system to wait to perform the transitions until fake-Kafka had revoked the just-recently-assigned partition.

What I can say is we now have nearly a week of runtime on an internal build of this change and have not seen the issue (or any others), whereas we saw it many times daily with the old code.

patriknw:
Ok, thanks

kciesielski:
@charles-crain Hi, are you still running the internal build without seeing the error reoccuring?

charles-crain:
@kciesielski Yes we have a lot of runtime on this now, in production with no issues.

kciesielski:
@charles-crain Thanks! Sounds great, merging.

PR #332, opened 2017-08-11
Add small note that it's 2.5 compatible

ktoso:
Should also be added in docs you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,14 @@ Supports Kafka 0.9.0.x
 ## Akka versions compatibility
 
 Please note that while the library depends on Akka 2.4.x is is fully compatible to be used with Akka 2.5.x.
-This is because of Akka's strict [backwards compatibility guarantees](http://doc.akka.io/docs/akka/2.5.3/scala/common/binary-compatibility-rules.html). If you want to use reactive-kafka with Akka 2.5 simply include Akka **and** Akka Streams dependencies using the latest 2.5 version.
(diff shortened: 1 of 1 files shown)


ktoso:
Refs https://github.com/akka/reactive-kafka/issues/316 
and :point_up: [August 11, 2017 4:48 AM](https://gitter.im/akka/reactive-kafka?at=598cb88d329651f46e0aa79a)

README.md
```diff
@@ -21,6 +21,11 @@ The documentation for the old API can be found in [OLD_README.md](https://github
 
 Supports Kafka 0.9.0.x
 
+## Akka versions compatibility
+
+Please note that while the library depends on Akka 2.4.x is is fully compatible to be used with Akka 2.5.x.
+This is because of Akka's strict [backwards compatibility guarantees](http://doc.akka.io/docs/akka/2.5.3/scala/common/binary-compatibility-rules.html). If you want to use reactive-kafka with Akka 2.5 simply include Akka Streams / Akka dependencies using the latest 2.5 version.
```
patriknw: it's important to define akka-streams, e.g. akka-actor would not be enough. Could we clarify that somehow. Perhaps replace "/" with "and" ?

ktoso: Yeah, I made it: `include Akka **and** Akka Streams dependencies using the latest 2.5 version.`

ktoso:
@kciesielski do you think we should add it to docs too?

PR #337, opened 2017-09-04
rename whitesource stable name to include minor version

patriknw:


raboof:
We should now also be able to update to sbt-whitesource 0.1.6, perhaps do that in one go?

PR #338, opened 2017-09-07
Expose metrics via Control

ktoso:
We'd like to be able to expose metrics to users when they use the Sharing the Consumer Actor pattern; I thought this style would likely be good enough - I assume the same could be done in the Publisher?

### DIFF
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -27,9 +27,24 @@ import scala.collection.JavaConverters._
 import scala.concurrent.duration._
 
 object KafkaConsumerActor {
-  case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
+  final case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
+
   def props[K, V](settings: ConsumerSettings[K, V]): Props =
     Props(new KafkaConsumerActor(settings)).withDispatcher(settings.dispatcher)
+
+  // ------------ public protocol ------------
+  /** DO NOT INHERIT */
+  sealed class RequestMetrics
+  case object RequestMetrics extends RequestMetrics {
+    /** Java DSL */
+    def getInstance: RequestMetrics = this
+  }
+  /** Contains a snapshot of metrics obtained from the underlying kafka consumer */
+  final case class ConsumerMetrics(metrics: Map[MetricName, Metric]) {
+    /** Java API */
+    def getMetrics: util.Map[MetricName, Metric] = metrics.asJava
+  }
+  // ------------ end of public protocol ------------
 
   private[kafka] object Internal {
     sealed trait SubscriptionRequest
@@ -61,17 +76,7 @@ object KafkaConsumerActor {
 
     private[KafkaConsumerActor] class NoPollResult extends RuntimeException with NoStackTrace
   }
-  
-  /** Public protocol which can be used to interact with the Actor */
-  object Protocol {
-    // requests
-    final case class RequestMetrics()
-    
-    // responses
-    /** Contains a snapshot of metrics obtained from the underlying kafka consumer */
-    final case class ConsumerMetrics(metrics: Map[MetricName, Metric])
-  }
-  
+
   private[kafka] def rebalanceListener(onAssign: Iterable[TopicPartition] => Unit, onRevoke: Iterable[TopicPartition] => Unit) = new ConsumerRebalanceListener {
     override def onPartitionsAssigned(partitions: util.Collection[TopicPartition]): Unit = {
       onAssign(partitions.asScala)
@@ -208,10 +213,10 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
         context.become(stopping)
       }
 
-    case Protocol.RequestMetrics() =>
+    case _: RequestMetrics =>
       val unmodifiableYetMutableMetrics: java.util.Map[MetricName, _ <: Metric] = consumer.metrics()
-      sender() ! Protocol.ConsumerMetrics(unmodifiableYetMutableMetrics.asScala.toMap)
-      
+      sender() ! ConsumerMetrics(unmodifiableYetMutableMetrics.asScala.toMap)
+
     case Terminated(ref) =>
       requests -= ref
       requestors -= ref
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -11,40 +11,25 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.regex.Pattern
 
-import akka.actor.{ Actor, ActorLogging, ActorRef, Cancellable, Props, Status, Terminated }
+import akka.actor.{Actor, ActorLogging, ActorRef, Cancellable, DeadLetterSuppression, NoSerializationVerificationNeeded, Props, Status, Terminated}
 import akka.event.LoggingReceive
 import org.apache.kafka.clients.consumer._
-import org.apache.kafka.common.{ Metric, MetricName, TopicPartition }
+import org.apache.kafka.common.{Metric, MetricName, TopicPartition}
 import org.apache.kafka.common.errors.WakeupException
 
-import scala.collection.JavaConverters._
 import java.util.concurrent.locks.LockSupport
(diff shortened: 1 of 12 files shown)


ktoso:
Please review is this general idea could make it in @patriknw @kciesielski @13h3r :)

13h3r:
Yes, I think it could be implemented this way

kciesielski:
Thanks, LGTM. As for producer, there's `Producer.flow(settings: ProducerSettings, producer: KafkaProducer)`, so users can provide their own producers which can be directly asked for metrics.

ktoso:
thanks, will finish this PR Then soon

On Sep 10, 2017 11:47 AM, "Krzysiek Ciesielski" <notifications@github.com>
wrote:

Thanks, LGTM. As for producer, there's Producer.flow(settings:
ProducerSettings, producer: KafkaProducer), so users can provide their own
producers which can be directly asked for metrics.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/akka/reactive-kafka/pull/338#issuecomment-328331691>,
or mute the thread

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -48,7 +52,17 @@ object KafkaConsumerActor {
 
     private[KafkaConsumerActor] class NoPollResult extends RuntimeException with NoStackTrace
   }
-
+  
+  /** Public protocol which can be used to interact with the Actor */
+  object Protocol {
+    // requests
+    final case class RequestMetrics()
```
patriknw: strange with case class without params, case object?

then it need accessor for Java API

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -48,7 +52,17 @@ object KafkaConsumerActor {
...
+    
+    // responses
+    /** Contains a snapshot of metrics obtained from the underlying kafka consumer */
+    final case class ConsumerMetrics(metrics: Map[MetricName, Metric])
```
patriknw: Java API: getMetrics: java.util.Map

v-gerasimov:
What about this PR? Seems there is no way to get consumer metrics now.

ktoso:
Hah, got reminded of this one via discuss.akka.io Fill fix it

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
ktoso: ok

ktoso:
Addressed feedback, checked Java API, added docs (also for producer where it's trivial however added it so it's all explained symmetrically).

Please review @kciesielski @patriknw

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -61,6 +77,15 @@ object KafkaConsumerActor {
...
   }
 
+  private[kafka] def rebalanceListener(onAssign: Iterable[TopicPartition] => Unit, onRevoke: Iterable[TopicPartition] => Unit) = new ConsumerRebalanceListener {
+    override def onPartitionsAssigned(partitions: util.Collection[TopicPartition]): Unit = {
+      onAssign(partitions.asScala)
+    }
+    override def onPartitionsRevoked(partitions: util.Collection[TopicPartition]): Unit = {
+      onRevoke(partitions.asScala)
+    }
+  }
```
patriknw: what does this have to do with metrics?

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -11,25 +11,41 @@ import java.util.concurrent.TimeUnit
...
+  case object RequestMetrics extends RequestMetrics {
+    /** Java DSL */
+    def getInstance: RequestMetrics = this
+  }
+  /** Contains a snapshot of metrics obtained from the underlying kafka consumer */
+  final case class ConsumerMetrics(metrics: Map[MetricName, Metric]) {
+    /** Java API */
+    def getMetrics: util.Map[MetricName, Metric] = metrics.asJava
+  }
+  // ------------ end of public protocol ------------
```
patriknw: This works when the KafkaConsumer actor is created and passed in, but what about the other cases where the actor is created by the library? Wouldn't it be better to expose the metrics via the `Control` materialized value?

Could be a Future and using ask internally.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
ktoso: Rebase goone weird :\ Thanks

ktoso:
Yeah you're right that a `def metrics` on the `Control` should be added as well; will do so

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
ktoso: Yeah good point, will add that

patriknw:
Perhaps `Control` can be the only public API?

ktoso:
> Perhaps Control can be the only public API?

Yeah, tried doing that. It's hard to make it `def metrics: Map[MetricName, Metric]` sadly, due to the SubSourceLogic and friends... This would be optimal API for users.

It is easier to make it a `Future[Map` which probably is good enough

ktoso:
Done, now the only API is to get metrics from the control value

ktoso:
Please give it a look @patriknw when you have some time

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -6,16 +6,21 @@
...
+import akka.kafka.KafkaConsumerActor.Internal.{ConsumerMetrics, RequestMetrics}
 import akka.kafka.Subscriptions._
 import akka.kafka.{ConsumerFailed, ConsumerSettings, KafkaConsumerActor, Subscription}
 import akka.stream.stage.GraphStageLogic.StageActor
 import akka.stream.stage.{GraphStageLogic, OutHandler, StageLogging}
 import akka.stream.{ActorMaterializerHelper, SourceShape}
-import org.apache.kafka.clients.consumer.ConsumerRecord
-import org.apache.kafka.common.TopicPartition
+import akka.util.Timeout
+import org.apache.kafka.clients.consumer.{Consumer, ConsumerRecord}
```
chbatey: Consumer is unused

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -9,6 +9,8 @@ import java.util.concurrent.TimeUnit
 
 import akka.NotUsed
 import akka.actor.{ActorRef, ExtendedActorSystem, Terminated}
+import akka.dispatch.ExecutionContexts
```
chbatey: Clean up imports, some of the new unused

docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
```diff
@@ -7,20 +7,23 @@ package sample.scaladsl
 import akka.pattern.ask
 import akka.kafka.ConsumerMessage.{CommittableMessage, CommittableOffsetBatch}
 import akka.kafka._
-import akka.actor.{Props, ActorRef, Actor, ActorSystem, ActorLogging, PoisonPill}
+import akka.actor.{Actor, ActorLogging, ActorRef, ActorSystem, PoisonPill, Props}
```
chbatey: Some of these new imports are unused

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -187,6 +192,11 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
         stopInProgress = true
         context.become(stopping)
       }
+
+    case RequestMetrics =>
+      val unmodifiableYetMutableMetrics: java.util.Map[MetricName, _ <: Metric] = consumer.metrics()
```
chbatey: This map goes through a lot of conversions for the Java case.

I guess there's no point trying to avoid this. Looking at the Kafka code they give us the raw map wrapped in an unmodifiable so we shouldn't pass that around.

ktoso: Yeah it is converted a lot indeed... I figured to leave it along until reported as a problem

ktoso:
Thanks! Exciting one :)

PR #349, opened 2017-10-02
Add withProperties(Map[String,String]) method to Kafka client settings

makubi:


core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -275,6 +275,13 @@ class ConsumerSettings[K, V](
   def withProperty(key: String, value: String): ConsumerSettings[K, V] =
     copy(properties = properties.updated(key, value))
 
+  /**
+    * The raw properties of the kafka-clients driver, see constants in
+    * `org.apache.kafka.clients.consumer.ConsumerConfig`.
+    */
+  def withProperties(properties: Map[String, String]): ConsumerSettings[K, V] =
```
patriknw: would also need Java API:
```
def withProperties(properties: java.util.Map[String, String]): ConsumerSettings[K, V]
```

makubi: Done

patriknw:
thanks

PR #350, opened 2017-10-03
Pass akka.kafka.consumer.close-timeout to KafkaConsumer.close

chbatey:
Helps fix https://github.com/akka/reactive-kafka/issues/252

Now users can set a short timeout for closing consumers


patriknw:
BTW, why are the defaults so high, 60s for producer and 20s for consumer? Shall we reduce them?

chbatey:
I don't know but the defaults on the kafka side is 30s for consumer and forever for the producer.

kciesielski:
Nice, thanks @chbatey, merging.

PR #351, opened 2017-10-04
 Use specfic exception for Kafka commit timeout rather exposing AskException

chbatey:
* To encourage users to look at their kafka cluster before assuming it
is a bug in the library
* Add a warning threshhold at which we start logging slow kafka commits

### DIFF
--- a/core/src/main/scala/akka/kafka/CommitTimeoutException.scala
+++ b/core/src/main/scala/akka/kafka/CommitTimeoutException.scala
@@ -1,0 +1,13 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.kafka
+
+import java.util.concurrent.TimeoutException
+
+/**
+ * Calls to `commitJavadsl` and `commitScaladsl` will be failed with this exception if
+ * Kafka doesn't respond within `commit-timeout`
+ */
+class CommitTimeoutException(msg: String) extends TimeoutException(msg) {}
--- a/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
@@ -127,9 +127,14 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       val commitMap = offsets.mapValues(new OffsetAndMetadata(_))
       val reply = sender()
       commitsInProgress += 1
+      val startTime = System.nanoTime()
       consumer.commitAsync(commitMap.asJava, new OffsetCommitCallback {
         override def onComplete(offsets: util.Map[TopicPartition, OffsetAndMetadata], exception: Exception): Unit = {
           // this is invoked on the thread calling consumer.poll which will always be the actor, so it is safe
+          val duration = System.nanoTime() - startTime
+          if (duration > settings.commitTimeWarning.toNanos) {
+            log.warning("Kafka commit took longer than `commit-time-warning`: {} ms", duration)
+          }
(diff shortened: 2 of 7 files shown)


chbatey:
Rebased

core/src/main/scala/akka/kafka/CommitTimeoutException.scala
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.kafka
+
+/**
+ * Calls to `commitJavadsl` and `commitScaladsl` will be failed with this expcetion if
+ * Kafka doesn't respond within `commit-timeout`
+ */
+class CommitTimeoutException(msg: String) extends Exception(msg) {}
```
johanandren: If we have it extend `TimeoutException` it would be source compatible.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -84,23 +85,30 @@ private[kafka] object ConsumerStage {
...
   }
 
   // This should be case class to be comparable based on ref and timeout. This comparison is used in CommittableOffsetBatchImpl
-  case class KafkaAsyncConsumerCommitterRef(ref: ActorRef, timeout: FiniteDuration)(implicit ec: ExecutionContext) extends Committer {
+  case class KafkaAsyncConsumerCommitterRef(ref: ActorRef, commitTimeout: FiniteDuration)(implicit ec: ExecutionContext) extends Committer {
     import scala.collection.breakOut
     import akka.pattern.ask
-    implicit val to = Timeout(timeout)
+    implicit val to = Timeout(commitTimeout)
+    private val commitTimeoutException = new CommitTimeoutException(s"Kafka commit took longer than: $commitTimeout")
```
johanandren: Nice to save the allocation, but, this will give a misleading stack trace, which isn't very nice.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -127,9 +127,14 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
       val commitMap = offsets.mapValues(new OffsetAndMetadata(_))
       val reply = sender()
       commitsInProgress += 1
+      val startTime = System.currentTimeMillis()
       consumer.commitAsync(commitMap.asJava, new OffsetCommitCallback {
         override def onComplete(offsets: util.Map[TopicPartition, OffsetAndMetadata], exception: Exception): Unit = {
           // this is invoked on the thread calling consumer.poll which will always be the actor, so it is safe
+          val duration = System.currentTimeMillis() - startTime
```
johanandren: `currentTimeMillis` can't be trusted to be monotonic (system clock adjustments etc.) use `nanoTime` for relative time calculations (although maybe not super important here)

patriknw: nanoTime 👍

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
chbatey: :+1:

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
chbatey: :+1: will update

core/src/main/scala/akka/kafka/CommitTimeoutException.scala
chbatey: :+1:

chbatey:
Ready for re-review @johanandren

PR #360, opened 2017-10-23
Add specific exception type for when the Kafka Consumer Actor is terminated

rgcase:
When the `KafkaConsumerActor` dies the stream stage would fail with `new Exception("Consumer actor terminated")`. This is more DRY and gives an exception type that can be matched on, rather than having to do something more awkward like 
```
case ex: Exception if ex.getMessage.contains("Consumer actor terminated") => ...
```

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerActorTerminatedException.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerActorTerminatedException.scala
@@ -1,0 +1,7 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.kafka
+
+class ConsumerActorTerminatedException extends Exception("Consumer actor terminated")
--- a/core/src/main/scala/akka/kafka/ConsumerFailed.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerFailed.scala
@@ -4,4 +4,4 @@
  */
 package akka.kafka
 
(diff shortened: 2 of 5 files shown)


core/src/main/scala/akka/kafka/ConsumerActorTerminatedException.scala
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.kafka
+
+class ConsumerActorTerminatedException extends Exception("Consumer actor terminated")
```
raboof: I agree a raw exception is a bit crude, on the other hand this seems rather specific and 'leaks implementation details' - allowing you to match on this specific exception makes it harder for us to refactor safely later.

Could we come up with a broader category of problems that we could put this exception under? What was the context in which you wanted to treat the "Consumer actor terminated" problem differently from any other problems?

patriknw: I agree with both. We should have a more specific exception, but still name it less implementation specific.
`ConsumerFailed`?

It should be a `RuntimeException`.

rgcase: A more general and less implementation specific name makes sense to me. Unfortunately I think we have to leave the message the same because there's probably already plenty of code out there with `if ex.getMessage.contains("Consumer actor terminated") => ...`. I like `ConsumerFailed`.

rgcase:
Should I squash commits or does someone do that on merge?

raboof:
We can have github take care of that on merge :+1:

PR #367, opened 2017-11-03
Allow storage of offsets outside Kafka, using Kafka to automatically assign partitions to consumers

rgcase:
Work in progress looking for feedback. Still needs tests and more robust live testing.


seglo:
Already mentioned this on the fork, but now that this is a legit PR I'll regurgitate it here.  This is a really great addition to reactive-kafka by exposing hooks to let users easily handle consumer group rebalancing.  I'm looking forward to using this in the Fast Data akka streams sample apps and recommending its use as part of the docs.  Thanks @rgcase

patriknw:
Refs https://github.com/akka/reactive-kafka/issues/328

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -99,6 +99,15 @@ object Consumer {
     Source.fromGraph(ConsumerStage.plainSubSource[K, V](settings, subscription))
 
   /**
+   * The `plainPartitionedManualOffsetSource` is similar to [[#plainPartitionedSource]] but allows the use of an offset store outside
+   * of Kafka, while retaining the automatic partition assignment. When a topic-partition is assigned to a consumer, the `loadOffsetsOnAssign`
+   * function will be called to retrieve the offset, followed by a seek to the correct spot in the partition. The `onRevoke` function gives
+   * the consumer a chance to store any uncommitted offsets, and do any other cleanup that is required.
+   */
+  def plainPartitionedManualOffsetSource[K, V](settings: ConsumerSettings[K, V], subscription: AutoSubscription, loadOffsetOnAssign: Set[TopicPartition] => Map[TopicPartition, Long], onRevoke: Set[TopicPartition] => Unit = _ => ()): Source[(TopicPartition, Source[ConsumerRecord[K, V], NotUsed]), Control] =
```
patriknw: For the record, we discussed the possibility to let `loadOffsetOnAssign` return a `Future[Map[TopicPartition, Long]]` instead to better support loading the offset asynchronously from an external place. The reason that doesn't work is because `seek` must be called before the `onPartitionsAssigned` callback is returned. Also note that `onPartitionsAssigned` is called on the same thread as called `poll`, i.e. the KafkaConsumerActor.

patriknw: btw, it should be renamed to `getOffsetOnAssign`

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -54,23 +54,18 @@ object KafkaConsumerActor {
...
 
-  private[kafka] def rebalanceListener(onAssign: Iterable[TopicPartition] => Unit, onRevoke: Iterable[TopicPartition] => Unit) = new ConsumerRebalanceListener {
-    override def onPartitionsAssigned(partitions: util.Collection[TopicPartition]): Unit = {
-      onAssign(partitions.asScala)
-    }
-    override def onPartitionsRevoked(partitions: util.Collection[TopicPartition]): Unit = {
-      onRevoke(partitions.asScala)
-    }
-  }
+  private[kafka] case class ListenerCallbacks(onAssign: (KafkaConsumer[_, _], Iterable[TopicPartition]) => Unit, onRevoke: Iterable[TopicPartition] => Unit)
```
patriknw: scala.Iterable might be mutable, would feel safer to use something that for sure is immutable, such as scala.collection.immutable.Seq, or `Set`. I think it's converted to `Set` later anyway.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -57,16 +59,24 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
     }
   }
 
-  val partitionAssignedCB = getAsyncCallback[Iterable[TopicPartition]] { tps =>
-    pendingPartitions ++= tps.filter(!partitionsInStartup.contains(_))
-    pump()
+  def partitionAssignedCB(kafkaConsumer: KafkaConsumer[_, _], tps: Iterable[TopicPartition]) = {
```
patriknw: As I understand it, those async callbacks must be removed. KafkaConsumerActor calls `poll`, `onPartitionsAssigned` is called, `seek` must be called before returning from `onPartitionsAssigned`. That means that all that must be in the KafkaConsumerActor without any asyncCallback.

rgcase: I actually think it's okay to put the call to `pump()` inside the async callback. If the `KafkaConsumer` has been assigned partitions, but `pump()` hasn't been called yet, it'll just mean that no downstream `Source`s will send a request for messages to the `KafkaConsumerActor` yet which is fine, it'll just pause the partition until it does receive a request.

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
rgcase: I've spent some more time digging into the `KafkaConsumer` code and I think I have a better understanding of what's happening now. I think we will actually be able to go back to accepting a `Future[Map[TopicPartition, Long]]` from `getOffsetOnAssign`. The `WrappedAutoPausedListener` pauses a partition as soon as it is assigned, so as long as that is done synchronously, we're free to do the `seek` and then `pump` asynchronously as long as we do them in that order, since a partition will only be `resume`d when a request comes to the `KafkaConsumerActor` from a downstream `Source`, which is created in `pump()`.

schrepfler:
What does outside mean? Any kind of external storage?

rgcase:
@schrepfler Yes that's the idea. You can store the offsets in some other database. It can be a little more flexible and it allows you to do a sort of transactional commit if you store the outcome of your processing, and the offset of a message in the same database in one transaction.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -57,16 +64,29 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
-    pendingPartitions --= tps
-    partitionsInStartup --= tps
-    tps.flatMap(subSources.get).foreach(_.shutdown())
-    subSources --= tps
+  private implicit val askTimeout = Timeout(5000, TimeUnit.MILLISECONDS)
+  def partitionAssignedCB(tps: Set[TopicPartition]) = {
+    implicit val ec = materializer.executionContext
+    getOffsetsOnAssign.fold(pumpCB.invoke(tps)) { getOffsets =>
+      getOffsets(tps).flatMap { offsets =>
+        consumer.ask(KafkaConsumerActor.Internal.Seek(offsets)).map(_ => pumpCB.invoke(tps))
```
patriknw: what should happen if ask timeout? ok to ignore?

rgcase: I think we'll have to fail the stream if the ask times out. We can't proceed correctly if we don't know if the `seek` happened. We could maybe retry once before failing. Sending this message won't result in any remote network calls, it's just a local message send to the `KafkaConsumerActor` and a call to `KafkaConsumer.seek` which doesn't trigger a call to the broker, so it shouldn't really ever time out anyway. If it does time out, something probably went wrong with the `KafkaConsumerActor`.

I'll add some logic to stop the stream if it happens.

patriknw: KafkaConsumerActor is sometimes blocking on the `poll`. Perhaps increase the timeout further and if it happens fail the stream. No need for retry.

rgcase:
Added tests and docs.

Issue #368, opened 2017-11-17
Remove usage of CallbackWrapper


patriknw:
In Akka 2.5.7 the async callbacks were improved so the internal CallbackWrapper can be removed. 

Note that the CallbackWrapper was removed so akka-streams-kafka will not work with Akka 2.5.7 until this issue is fixed.

See https://github.com/akka/akka/pull/23185

patriknw:
This is probably a good reason for stop supporting 2.4.x and build with 2.5.x by default.

patriknw:
I'm fixing this, and releasing

PR #371, opened 2017-11-21
update to Akka 2.5.7, #368

patriknw:
* and remove CallbackWrapper, since that internal class was
  removed in 2.5.7
* the AsyncCallback is now safe to use immediately as materialized
  value

Refs #368

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ If you have questions or are working on a pull request or just curious, please f
 Created and maintained by
 [<img src="https://softwaremill.com/images/header-main-logo.3449d6a3.svg" alt="SoftwareMill logo" height="25">](https://softwaremill.com)
 
-## Documentation for version 0.11 or later
+## Documentation for version 0.18 or later
 
 [Documentation](http://doc.akka.io/docs/akka-stream-kafka/current/) and [API](http://doc.akka.io/api/akka-stream-kafka/current/)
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -23,13 +23,7 @@ Supports Kafka 0.9.0.x
...
 ## Akka versions compatibility
 
-Please note that while the library depends on Akka 2.4.x is fully compatible to be used with Akka 2.5.x.
-This is because of Akka's strict [backwards compatibility guarantees](http://doc.akka.io/docs/akka/2.5.3/scala/common/binary-compatibility-rules.html). If you want to use reactive-kafka with Akka 2.5 simply include Akka **and** Akka Streams dependencies using the latest 2.5 version, for example like this:
-
-```
-libraryDependencies += "com.typesafe.akka" %% "akka-stream" % "2.5.x" // pick the latest version here
-libraryDependencies += "com.typesafe.akka" %% "akka-actor" % "2.5.x"
-```
+Please note that while the library depends on Akka 2.5.x. It can't be used with Akka 2.4.x. Version 0.17 and earlier can be used with Akka 2.4.x.
```
raboof: 'while' can be dropped.

ktoso: Link to http://doc.akka.io/docs/akka/current/scala/common/binary-compatibility-rules.html to clarify that it's easy for people to upgrade?

Issue #301, opened 2017-05-08
Feature Request: periodically re-submit the previously committed offset 


samidalouche:
The main problem with kafka consumer is explained here: https://issues.apache.org/jira/browse/KAFKA-3806

Apparently, one of the answers suggests to periodically re-submit the previously committed offset, even no activity occurred. 

So, it would be handy if reactive-kafka provided some logic that could easily allow to re-submit the previously committed offsets to make sure they never disappear from kafka.

njhartwell:
:+1:

asflierl:
Running into the same issue. It's important to note that a consumer also needs to periodically commit the current offset *before* having received the first message in the current session (there might be an offset stored in Kafka for this client/group from a previous session and we do not want that to reset when there are no messages for a day).

PR #375, opened 2017-12-06
Allow committed offsets refresh

rtimush:
This pull request solves the problem described in https://issues.apache.org/jira/browse/KAFKA-3806 and #301.

In some workloads, it is possible that no events are sent to a given topic/partition for 24 hours. By default, Kafka evicts committed offsets after this period, so if the consumer is then restarted it has no stored offset to start with. Depending on the `auto.offset.reset` it may result in re-processing of the whole history, losing some recent events or a failure.

This pull request allows a consumer to refresh the committed offset on a regular basis even if there are no new messages in the topic. The following logic applies:
- If the last commit has been successfully completed, the offsets from this commit are sent.
- If there were no messages at all since the consumer started, the original offset for a partition is sent.
- If the last commit has failed, no offset refresh for the partitions affected happens.

The feature should be explicitly enabled by calling `withCommitRefreshInterval` on the `ConsumerSettings`, so existing client won't be affected.

There is an integration test for this feature, but it takes a few minutes to run because the eviction timeout is configured in minutes and cannot be fractional.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -62,6 +62,11 @@ akka.kafka.consumer {
   # After exceeding maxinum wakeups the consumer will stop and the stage will fail.
   max-wakeups = 10
 
+  # If enabled the consumer will re-send last committed offsets periodically
+  # for all assigned partitions. See https://issues.apache.org/jira/browse/KAFKA-4682.
+  commit-refresh-enabled = false
+  commit-refresh-interval = 1m
+
   # If enabled log stack traces before waking up the KafkaConsumer to give
   # some indication why the KafkaConsumer is not honouring the poll-timeout
   wakeup-debug = true
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -62,10 +62,9 @@ akka.kafka.consumer {
   # After exceeding maxinum wakeups the consumer will stop and the stage will fail.
   max-wakeups = 10
 
-  # If enabled the consumer will re-send last committed offsets periodically
+  # If set to a finite duration the consumer will re-send last committed offsets periodically
   # for all assigned partitions. See https://issues.apache.org/jira/browse/KAFKA-4682.
-  commit-refresh-enabled = false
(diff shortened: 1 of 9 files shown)


bsideup:
@rtimush @ktoso @patriknw do you have any updates on the status of this PR?

ktoso:
We'll soon have @ennru join us and spend considerable time on Alpakka projects, which includes this, which means quicker turnaround in merging things very soon :-) Thanks for your patience, Alpakka has proven it self and is going to get the attention it deserves very soon :-)

fchaillou:
@ktoso no worries, I'm interested in these features so i'm doing what i can to help you in my own way !
Welcome and good luck to @ennru then :+1:

ennru:
Good work and welcome to the Kafka connector project!

Could you rebase, please?

rtimush:
Thanks, @ennru. Rebased.

ennru:
Great, thank you.
I'd like to see some documentation for this feature. We'd at least need a section in `reference.conf`.

rtimush:
Makes sense, I added parameters to the `reference.conf` and also added a small note in the consumer docs.

core/src/main/resources/reference.conf
```diff
@@ -62,6 +62,11 @@ akka.kafka.consumer {
   # After exceeding maxinum wakeups the consumer will stop and the stage will fail.
   max-wakeups = 10
 
+  # If enabled the consumer will re-send last committed offsets periodically
+  # for all assigned partitions. See https://issues.apache.org/jira/browse/KAFKA-4682.
+  commit-refresh-enabled = false
+  commit-refresh-interval = 1m
```
ktoso: would it not be nicer to have one value, and it being `infinite` being "disabled"? That's usually how we have such settings in Akka and Akka HTTP

rtimush: It is indeed much nicer, but I didn't want to introduce a custom config behavior. As you say it is already done like this in akka, I will change the code here to use a single config value too.

ennru:
Thank you for your contribution!

Issue #336, opened 2017-09-01
Consumer might skip offsets


kodemaniak:
We are facing a strange behaviour using reactive-kafka. 

We are consuming a single topic with 256 partitions and about 2.5 Million messages. From that log we build a view on our data and strongly require at-least-once message semantics in the consumer in order to correctly build the view. 

We experienced a lot of exceptions resulting from events that have been skipped, e.g., an update message coming in for a given entity without the corresponding create message creating that entity in the first place. We wrote some tooling to check whether the events are in the correct partition and are stored in the right order and were able to confirm that the log (i.e. the topic/partition) is in the expected state.

After a lot of debugging we think we have reduced the problem to reactive-kafka. We use a partitioned commitable source and consume each partition separately. Due to some internal issues the processing of the messages was rather slow on our side. This triggered a couple of timeouts on our side but also on the kafka consumer side, in turn leading to a lot of rebalances. Regarding the Kafka documentation this is expected and can be fine tuned using timeout settings and the `ConsumerConfig.MAX_POLL_RECORDS_CONFIG`. However, the experienced behaviour that messages are skipped should not occur.

It seems a topic that is being rebalanced onto the same node reuses the `KafkaConsumer` and skips exactly `ConsumerConfig.MAX_POLL_RECORDS_CONFIG` records. Our assumption is that as long as the consumer is the same, it will never reset it's internal state but request new messages from the broker (`ConsumerConfig.MAX_POLL_RECORDS_CONFIG` records in one reqeust to the broker). We assume that the `KafkaConsumer` relies on a restart/recreation of the consumer when something fails during processing in order to restart from the last committed offset. When we change the `MAX_POLL_RECORDS_CONFIG' to different values, the number of skipped messages is always consistent with that setting. 

In our setting we also have messages that can be applied to the view independently of the previous messages, so that we actually successfully commit offsets after messages have been skipped. Therefore, we are basically ending up in a totally incomplete and irrecoverable state of the view. 

Correct behaviour would be, IMO, that after a failure during the processing, the consumer needs to be reset to the last committed offset. This must probably be done by the `KafkaConsumerActor` by recreating the consumer.

patriknw:
Interesting, would you be able to create a pull request?

kodemaniak:
Yes, I will work on that and try to produce a pull request.

seglo:
[...]

Issue #357, opened 2017-10-17
Using takeWhile terminates CommittableSource while commits are in progress


mina-mishriky-ck:
We are having an issue when using a `takeWhile`operation on a `CommittableSource` to consume messages up to a certain offset.
The graph simply looks like:
`CommittableSource` ~~> `takeWhile(...)` ~~> `Flow` (performs logic and commits offsets) ~~> `SinkActor`

The issue that we are facing is once the `takeWhile` condition is reached, the source is canceled immediately, which results in commits failing for offsets that had originally passed the `takeWhile` condition. The error message is `"akka.kafka.KafkaConsumerActor$Internal$Commit"`

I believe the expected behavior should be to wait for outstanding offsets to be committed before shutting down. We are open to alternative approaches for consuming up to a certain offset, if better ones exist.

mina-mishriky-ck:
Any thoughts on this?

pierangeloc:
I encountered a similar issue when using simple `take`. I have a simple flow that consumes from `sourceTopic`, transforms the string payloads and puts them in `targetTopic`.

```scala
Consumer.committableSource(consumerSettings, Subscriptions.topics(sourceTopic))
    .take(10)
    .map { msg =>
      ProducerMessage.Message(new ProducerRecord[Array[Byte], String](
        targetTopic,
        msg.record.value + "-processed"
      ), msg.committableOffset
      )
    }
    .via(Producer.flow(producerSettings))
    .map(_.message.passThrough)
    .batch(max = 20, first => CommittableOffsetBatch.empty.updated(first)) { (batch, elem) =>
      batch.updated(elem)
    }
    .mapAsync(3)(_.commitScaladsl())
    .runWith(Sink.ignore)
```
After the run, if I run the graph again it restarts from the same offset. By inspecting the offset in Zookeeper for the consumer group, I see that offset is not committed.
Removing the `take` and letting it consume all available elements from source topic solves the problem.

ennru:
Is this solved by the `commitsInProgress` logic in `KafkaConsumerActor`?
Or is this different to #166 
See test cases in #490

2m:
Closing this ticket due to inactivity. Please re-open this ticket with more information if you still find the original issue relevant.

PR #377, opened 2017-12-07
Implement Consumer handling directly in GraphStages

kodemaniak:
Triggered by #336 I started to implement a `PlainSource` and a `CommittingStage` as GraphStages. Instead of using the `KafkaConsumerActor` they handle the `KafkaConsumer` directly. Each stream uses its own consumer thus preventing the issues from #336 (which were mainly due to inconsistent state within the actor).

I think that this approach is much simpler to undestand and maintain than the `KafkaConsumerActor. Furhermore the benchmarks show that it performs better.

I implemented additional factory methods to use the new graph stages and to be able to compare them to the existing ones. We currently use these two GraphStages embedded in our system in production and did not encounter any issues so far. I have not yet moved the tests to the reactive-kafka code base, and integration tests are missing. But I wanted to put this implementation up for discussion to see whether it might be alternative approach to handling consumers in reactive-kafka.

Specifically, the API of the CommittingStage is different to the original committableSource API. This is mainly due to the fact that the stage commits on its own and the user does not have to commit on the records himself. This is closer to Kafka Streams. Furthermore, I assume some polishing has to be applied to be usable in a generic library, e.g., I am not sure if the subscriptions are currently handled correctly, since we don't use them in our code base (i.e. we have a slightly simplified version of the stages).

So please consider this more as a proposal than a complete pull request. Any feedback is welcome.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/CommittingStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/CommittingStage.scala
@@ -1,0 +1,192 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.kafka.internal
+
+import java.util
+import java.util.regex.Pattern
+
+import akka.NotUsed
+import akka.kafka.Subscriptions._
+import akka.kafka.{ConsumerSettings, Subscription}
+import akka.stream.stage._
+import akka.stream._
+import akka.stream.scaladsl.{Flow, GraphDSL, Source}
+import org.apache.kafka.clients.consumer._
+import org.apache.kafka.common.TopicPartition
+
+import scala.annotation.unchecked.uncheckedVariance
+import scala.collection.JavaConverters._
+import scala.collection.immutable
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+/**
+ * @author carsten
+ */
+final case class BidiSourceShape[-FO, +FI, -O](flowOut: Outlet[FO @uncheckedVariance], flowIn: Inlet[FI @uncheckedVariance], out: Outlet[O @uncheckedVariance]) extends Shape {
+  override val inlets: immutable.Seq[Inlet[_]] = flowIn :: Nil
+  override val outlets: immutable.Seq[Outlet[_]] = flowOut :: out :: Nil
+
+  override def deepCopy(): BidiSourceShape[FO, FI, O] = BidiSourceShape(flowOut.carbonCopy(), flowIn.carbonCopy(), out.carbonCopy())
+
+  override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape = {
+    require(inlets.size == 1, s"proposed inlets [${inlets.mkString(", ")}] do not fit BidiSourceShape")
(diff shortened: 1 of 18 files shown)


raboof:
(looks like that travis failure was unrelated, #381)

edrevo:
I think this is a great PR! I would not only solve #336, but also #382. I think having one consumer per source makes things much simpler.

A small piece of feedback: it would be great to use the same model as the current `commitableSource` to manage commits. That approach has some problems (#357), but I think that can be fixed by sending the KafkaConsumer with the CommitabfOffset and counting the number of alive references to the consumer. Closing the KafkaConsumer would be responsibility of whoever drops the refCount to zero.

I am very interested in this code, since my scenario requires _at-least-once_ semantics, which is currently broken. I am happy to evolve this PR as needed if necessary so we can merge this. @NeQuissimus, can this PR be merged or is there any necessary change still pending?

NeQuissimus:
@edrevo I just reviewed the code, I do not have merge privileges :) IMO it is good to go.

edrevo:
can anyone from the akka team take a look at this (cc @kciesielski, @patriknw which are the top committers to this project)

kodemaniak:
I do also have tests for the code which I would need to port to reactive-kafka from our internal version that we are currently using. Furthermore, I have added the `Control` materialized value to the stages, so that one can close the consumer from the outside and stop the stream.

Regarding the `commitableSource` model, I also think that it can be implemented. I did not go for that approach since I think explicitly committing the offset in the stream has no real advantage, it is just boiler plate code for me. But maybe providing both alternatives would be an option.

I will try to port my latest changes to this pull request ASAP. Should this be merged, I would prefer to switch back to reactive kafka anyway.

mbknor:
@edrevo I've added a different approach to solve #357 
Please have a look

core/src/main/scala/akka/kafka/internal/CommittingStage.scala
```diff
@@ -0,0 +1,204 @@
...
+    apply(consumer, flow, commitBatch)
+  }
+
+  def apply[K, V](consumer: Consumer[K, V], flow: Flow[ConsumerRecord[K, V], ConsumerRecord[K, V], NotUsed], commitBatch: Int): Source[ConsumerRecord[K, V], Control] = {
+    Source.fromGraph(GraphDSL.create(new CommittingStage(consumer, commitBatch)) { implicit b => stage =>
+      import GraphDSL.Implicits._
+
+      val handlerFlow = b.add(flow)
+
+      stage.flowOut ~> handlerFlow ~> stage.flowIn
```
seglo: It wasn't clear to me at first why the Consumer commit and poll implementation were implemented in the `CommitStage`, but now I understand that you're building the graph internally here (or at least the segment of graph where the record is consumed, transformed, and committed).  I like this approach because it prevents the possibility of using a `committableSource` without actually committing the records, which doesn't make sense, but is currently possible

Injecting user flow in this manner also greatly simplifies the problem of keeping a reference to the Consumer before the user's flow and after.  The current implementation is quite complicated and leaves a lot to be desired IMO.

ennru:
We've worked through the `KafkaConsumerActor` quite a bit for today's release of 1.0-M1, some issues have been resolved, for others we need the feedback from our power users to understand if they are solved in real-world scenarios.

Thank you for putting this PR together in the first place. For now, we do not consider to have a second way to use Kafka's consumer within Alpakka Kafka.

PR #380, opened 2017-12-15
Replace calls to deprecated `expectNoMsg`-method

Philippus:
Applying this PR will remove 5 warnings.

### DIFF
--- a/core/src/test/scala/akka/kafka/internal/ConsumerTest.scala
+++ b/core/src/test/scala/akka/kafka/internal/ConsumerTest.scala
@@ -120,7 +120,7 @@ class ConsumerTest(_system: ActorSystem)
 
       probe
         .request(1)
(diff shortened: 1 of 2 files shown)


raboof:
The fix for the travis failure is in #381

core/src/test/scala/akka/kafka/internal/ConsumerTest.scala
```diff
@@ -120,7 +120,7 @@ class ConsumerTest(_system: ActorSystem)
 
       probe
         .request(1)
-        .expectNoMsg()
+        .expectNoMessage(200.millis)
```
raboof: I wonder if we should provide a version of this method with a default timeout here after all (though independent from and smaller than the default expect timeout, see https://github.com/akka/akka/pull/23381#discussion_r140767414)

PR #385, opened 2017-12-25
Make ConsumerSettings return Consumer (over KafkaConsumer)

fmovlex:
Make `ConsumerSettings` return the base `Consumer[K,V]` interface from here: `def createKafkaConsumer(): Consumer[K, V]`.

It allows us to replace/decorate the underlying Consumer implementation, e.g.
- Replace `kafka.KafkaConsumer` with `kafka.MockConsumer`
- Decorate the consumer with `brave.zipkin.TracingConsumer`

Edit: The producer already allows for this behavior, as the `Sink`s accept the base interface (`org.apache.kafka.clients.producer.Producer`)


raboof:
Thanks @fmovlex , sorry that this seems to have fallen off the radar somehow!

PR #389, opened 2018-01-09
exclude KafkaConsumer messages from serialization verification

ktoso:
Since these are known to be sent only locally, let's exclude them for serialization verification, and also enable it such that we make sure that others, who use the library can enable `akka.actor.serialize-messages`.

Mostly a quality of life update for downstream users of the library :)


johanandren:
Wouldn't it be better to just not have serialize-messages enabled when running things that will only ever run locally?

ktoso:
Tell that to users of this library which do want to enable this setting? ;-)

It's not about *this* library, it is about *users* of this library which want to enable that setting. Yet they can not, because the lib breaks then. By not having the setting enabled in tests here, we allow sneaking in such errors again, and the library will again become a blocker for users who want to verify their serialization works, since it will always blow up, because this library.

I also proposed a ticket how we could provide a way for users to self check and add such exclusions here: https://github.com/akka/akka/issues/24273

ktoso:
Conflict resolved; Can I merge?

ktoso:
Since responses follow this pattern already I'll merge this. Was discovered as an issue in a customer interaction as well

PR #390, opened 2018-01-11
Emit read elements on SubSourceLogic before closing

### LINKED ISSUE #382: Lost messages in plainPartitionedSource on partition reassignment
I am currently seeing the following behavior, using reactive-kafka 0.18:

1. I am sending 2 messages to a Kafka topic, which fall in two different partitions (partition 0 and partition 3)
2. A Kafka `plainPartitionedSource` is created and joins the consumer group at 22:51:04
3. A second Kafka `plainPartitionedSource` is created and joins the consumer group at 22:51:13
4. A partition rebalance happens between 22:51:28 and 22:51:34 (I see several `RequestMessages from topic/partition X already registered by other stage`)
5. Partition 0 stays with the first consumer
6. Partition 3 moves to the second consumer
7. The message to partition 0 is never emitted by the Source
8. The message to partition 3 is correctly emitted by the second consumer

Upon inspecting the reactive-kafka's source code, here is what I think is happening in the first consumer:

1. All partitions are assigned to the consumer, which triggers `partitionAssignedCB` (SubSourceLogic:78) which in turns `pump`s and emits a Source for each partition
2. The Source for partition 0 receives a `pull`, which sends a RequestMessages to the kafka consumer actor (SubSourceLogic:235), _but_ still hasn't received the messages associated with that request
3. The new consumer joins the consumer group, and Kafka first revokes all partitions and then assigns the new ones.
4. Revoking of all partitions causes the Source for partition 0 to be cancelled (SubSourceLogic:99)
5. The Source for partition 0 calls `completeStage` (SubSourceLogic:222), regardless of whether it was mid-request (`requested = true`) or not.
6. The Kafka actor performs a poll for partition 0, since it has a request for it.  And sends the resulting messages to the Source that was just closed (KafkaConsumerActor:346)

This causes the Kafka consumer to mark the message in partition 0 as received, but nobody actually received it.

Does this sound like a reasonable explanation? cc @13h3r, @elkozmon, @patriknw, @rgcase since you are authors of the SubSouceLogic file.

I am happy to provide a PR to fix this, but I'm not sure of what the best solution is (I'm pretty new to Kafka and Akka Streams). Is it reasonable to emit all pending messages before closing the Source? Or is that a big no-no?

edrevo:
**Description** An attempt to fix #382. I don't have much experience with akka-streams or reactive-kafka, so it is entirely possible that my fix is sub-optimal or just plain wrong. Any feedback is welcome.
**Testing** I have a test that (sometimes) reproduces the issue in https://github.com/edrevo/reactive-kafka/commit/51a7b08d581d2b952c41d99be30b368815bd0250 and after the fix it seems to work consistently. The test is very low-quality, so I am not submitting it in this PR.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
@@ -147,6 +147,7 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
         val shape = stage.shape
         val requestMessages = KafkaConsumerActor.Internal.RequestMessages(0, Set(tp))
         var requested = false
+        var shuttingDown = false
         var self: StageActor = _
         var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
 
@@ -177,7 +178,7 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
 
         setHandler(out, new OutHandler {
           override def onPull(): Unit = {
-            pump()
+            if (!shuttingDown) pump()
           }
 
           override def onDownstreamFinish(): Unit = {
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @edrevo,

edrevo:
Closing this PR. I just noticed that I am not calling `.invoke` on the callback. Once I add that, the code fails again... :/

PR #398, opened 2018-01-27
Travis build:  add Scala 2.12.4 to the build matrix

sullis:


raboof:
(the Travis failure was unrelated, logged that as #400)

sullis:
ready for review.

.travis.yml
```diff
@@ -1,13 +1,14 @@
...
 scala:
-  - "2.11.11"
+  - "2.11.12"
+  - "2.12.4"
 jdk:
   - oraclejdk8
 script:
   # make 'git branch' work again
   - git branch -f "$TRAVIS_BRANCH" && git checkout "$TRAVIS_BRANCH"
-  - sbt -jvm-opts .travis-jvmopts +test
```
raboof: The `+` in `+test` makes sure the tests are ran across multiple Scala versions (see the [sbt docs](http://www.scala-sbt.org/1.x/docs/Cross-Build.html#Cross-Building+a+Project) and [our config](https://github.com/akka/reactive-kafka/blob/master/build.sbt#L43)) - so as far as I can see we're already testing against both Scala versions, and this change does not appear to be needed?

sullis: I just pushed an update.   I  changed "+test" to "test"

raboof:
What's the benefit of splitting the build from one 5m build (that tests both 2.11 and 2.12) into two separate 4m builds (for 2.11 and 2.12 separately)?

I'm not sure it's an improvement, and it might make deploying/releasing from Travis in the future harder.

raboof:
Sorry, not sure it's an improvement, so closing this PR for now - feel free to re-open it when we run into a situation where this could help!

PR #407, opened 2018-02-07
Fixing #357 - committableSource now delays consumer shutdown if it has un-committed messages inflight

mbknor:
Delays the consumer-shutdown either when all inflight messages are committed OR by timing out.

The "when all inflight messages are committed" can be kind of tricky since akka stream is buffering messages.

When using .take( x ) close to the source, we will have no buffered messages.
If you know how many messages are buffered, you can choose to configure 'max-inflight-commits-on-shutdown'.

See test.

It would be great to get some feedback on this PR.

I have tested it in my internal app, and it DO make the take/takeWhile problem go away (https://github.com/akka/reactive-kafka/issues/357)

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -50,6 +50,18 @@ akka.kafka.consumer {
 
   # If commits take longer than this time a warning is logged
   commit-time-warning = 1s
+
+  # The max time we wait for inflight commits after commitableSource shutdown.
+  # Also see max-inflight-commits-on-shutdown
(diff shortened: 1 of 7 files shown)


mbknor:
Any comments on this one, @kciesielski @patriknw?

ennru:
We've just released 0.21 which includes the change to use of Scalafmt. Please check it out - you're more than welcome to revive this PR.

ennru:
Today we're releasing 1.0-M1. As this PR seems abandoned, I will close it now.

PR #408, opened 2018-02-09
Fix wrong unit in log message, closes #406

### LINKED ISSUE #406: Wrong units on "Kafka commit took linger than ..." log message
I noticed log messages like:
`Kafka commit took longer than commit-time-warning: 1328712428 ms`

That translates to many days which seemed wrong, so I looked for the code responsible for this. Looks like the correct unit should be nanoseconds:
https://github.com/akka/reactive-kafka/blob/f3d6ce9d35ae1680c943ab5a195fca7eb739f13c/core/src/main/scala/akka/kafka/KafkaConsumerActor.scala#L139

Philippus:


core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -136,7 +136,7 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
           // this is invoked on the thread calling consumer.poll which will always be the actor, so it is safe
           val duration = System.nanoTime() - startTime
           if (duration > settings.commitTimeWarning.toNanos) {
-            log.warning("Kafka commit took longer than `commit-time-warning`: {} ms", duration)
+            log.warning("Kafka commit took longer than `commit-time-warning`: {} ns", duration)
```
johanandren: Not quite convinced ns granularity is useful here, wouldn't truncating to ms, or pretty-printing to biggest possible unit (I think we have some util for that in Akka) be better?

Philippus: do you mean `toCoarsest`? I think that will not help here, because that will always return nanos in this case. I've changed the code so it will convert to millis now.

PR #414, opened 2018-02-27
Wakeup counter reset after certain amount of time

xbaran:
Added **wakeup-window** property default: -1

- -1 -> never reset wakup counter
- any time period e.g. 24h -> will reset wakeup counter every 24h


chbatey:
Wakeups are reset to 0 every time we successfully call the KafkaConsumer poll method and a poll is scheduled every 50ms by default.

What scenario are you addressing here?

xbaran:
Well, good point!

I've just missed the wakeup reset after poll I'm afraid 🤦‍♂️

Issue #369, opened 2017-11-18
Transactional support for Kafka consume-transform-produce workflows


seglo:
Kafka announced several new features in its 0.11.0.0 release that collectively allow for exactly-once semantics, or "effective-once" depending on your opinions on the subject.  Semantics aside, there are several new Kafka client features that would be nice to expose in reactive-kafka's API.

Exactly-once delivery on the Kafka producer was impossible to guarantee before the release of 0.11.0.0 because there was a possibility that a producer could send a duplicate message in a retry and fail before it received an acknowledgement.

Kafka achieves exactly-once by making a number of changes to Kafka broker and client code.

* Idempotent producers. When `ENABLE_IDEMPOTENCE` is enabled in the producer config it will coordinate with the broker to ensure that retried messages aren't duplicated.
* Transactions have been introduced to guarantee exactly once message delivery in a `consume->transform->produce` workflow. This is a common operation involving consuming from a Kafka topic, processing, and producing to a new topic. To make transactions a reality in Kafka 0.11 the following changes were made:
  * Idempotent producers
  * Transaction Coordinator: New logical component on broker 
  * Message schema changes: New concepts such as Transaction ID, Producer ID, and transaction sequence numbers.

Supporting these features will help close the gap between Kafka streams and Akka streams w/ reactive-kafka for stateless streaming/transformation workflows.  I would like to work on a PR to implement these new features of the Kafka client.

Kafka Improvement Proposals (KIP) for Transaction Support
* [KIP-98: Exactly Once Delivery and Transactional Messaging](https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging)
* [KIP-129: Streams Exactly-Once Semantics](https://cwiki.apache.org/confluence/display/KAFKA/KIP-129%3A+Streams+Exactly-Once+Semantics)

ktoso:
I wonder how much it can be incorporated -- looking forward to a PR then @seglo !

seglo:
Great, thanks! I've used these features with the underlying Kafka client libs.  I'm optimistic it can be incorporated.  I'll start working on it this week.

mjkrumlauf:
@seglo how is this progressing?

seglo:
[...]

PR #420, opened 2018-03-26
Kafka Transactions

seglo:
This PR is my first draft of Kafka transactional support to allow for "exactly once" message _processing_ semantics.  I originally proposed this feature in #369 .  I've been tinkering for some time now and I'm eager for feedback from the maintainers and users of this project.  This implementation is inspired by the `exactly_once` setting for the `processing.guarantee` configuration found in Kafka Streams ([KIP-129](https://cwiki.apache.org/confluence/display/KAFKA/KIP-129%3A+Streams+Exactly-Once+Semantics), [design document](https://docs.google.com/document/d/1pGZ8xtOOyGwDYgH5vA6h19zOMMaduFK1DAB8_gBYA2c/edit#heading=h.vkrkjfth3p8p)).  I've included [documentation](https://github.com/seglo/reactive-kafka/blob/transactions/docs/src/main/paradox/transactions.md) which describes the implementation and usage in detail.  Some excerpts are copied into the PR description below for convenience.

> # Transactions 
> Kafka Transactions provide guarantees that messages processed in a consume->transform->produce workflow (consumed from a source topic, transformed, and produced to a destination topic) are processed exactly once or not at all.  This is achieved through coordination between the Kafka consumer group coordinator, transaction coordinator, and the consumer and producer clients used in the user application.  The Kafka producer marks messages that are consumed from the source topic as "committed" only once the transformed messages are successfully produced to the sink.   
> 

...

> ## Consume Transform Produce Workflow
> Kafka transactions are handled transparently to the user.  The `Consumer.transactionalSource` will enforce that a consumer group id is specified and the `Producer.transactionalFlow` or `Producer.transactionalSink` will enforce that a transactional id is specified.  All other Kafka consumer and producer properties required to enable transactions are overridden.
> 
> Transactions are committed on an interval which can be controlled with the producer config `akka.kafka.producer.eos-commit-interval-ms`, similar to how exactly once works with Kafka Streams.  The default value is `100ms`.  The larger commit interval is the more records will need to be reprocessed in the event of failure and the transaction is aborted.
> 
> When the stream is materialized the producer will initialize the transaction for the provided `transactionalId` and a transaction will begin.  Every commit interval (`eos-commit-interval-ms`) we check if there are any offsets available to commit.  If offsets exist then we suspend backpressured demand while we drain all outstanding messages that have not yet been successfully acknowledged (if any) and then commit the transaction.  After the commit succeeds a new transaction is begun and we re-initialize demand for upstream messages.
> 
> To gracefully shutdown the stream and commit the current transaction you must call `shutdown()` on the `Control` materialized value to await all produced message acknowledgements and commit the final transaction.  

An oversimplified example using `RestartableSource` to recover from transient errors encountered in a stream.  For a more detailed example see the [`"successful consume transform produce transaction with transient failure causing an abort with restartable source"` test in `IntegrationSpec`](https://github.com/seglo/reactive-kafka/blob/transactions/core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala#L465).

```
    var innerControl = null.asInstanceOf[Control]

    val stream = RestartSource.onFailuresWithBackoff(
      minBackoff = 1.seconds,
      maxBackoff = 30.seconds,
      randomFactor = 0.2
    ) { () =>
      Consumer.transactionalSource(consumerSettings, Subscriptions.topics("source-topic"))
        .via(business)
        .map { msg =>
          ProducerMessage.Message(
            new ProducerRecord[Array[Byte], String]("sink-topic", msg.record.value), msg.partitionOffset)
        }
        // side effect out the `Control` materialized value because it can't be propagated through the `RestartSource`
        .mapMaterializedValue(innerControl = _)
        .via(Producer.transactionalFlow(producerSettings, "transactional-id"))
    }

    stream.runWith(Sink.ignore)

    // Add shutdown hook to respond to SIGTERM and gracefully shutdown stream
    Runtime.getRuntime.addShutdownHook(new Thread { Await.result(innerControl.shutdown(), 10.seconds) })
```


core/src/main/scala/akka/kafka/scaladsl/Producer.scala
```diff
@@ -72,6 +74,35 @@ object Producer {
...
+   * emits a [[ConsumerMessage.TransactionalMessage]].  The flow requires a unique `transactional.id` across all app
+   * instances.  The flow will override producer properties to enable Kafka exactly once transactional support.
+   */
+  def transactionalFlow[K, V](settings: ProducerSettings[K, V], transactionalId: String): Flow[Message[K, V, ConsumerMessage.PartitionOffset], Result[K, V, ConsumerMessage.PartitionOffset], NotUsed] = {
+    require(transactionalId != null && transactionalId.length > 0, "You must define a Transactional id.")
+
+    val txProducerSettings = settings.withProperties(
+      ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG -> true.toString,
+      ProducerConfig.TRANSACTIONAL_ID_CONFIG -> transactionalId,
+      ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION -> "1"
```
maasg: `1.toString` ?  (Or even better, consider defining default values  as constants)

core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -418,5 +420,109 @@ class IntegrationSpec extends TestKit(ActorSystem("IntegrationSpec"))
 
       }
     }
+
+    "successful consume transform produce transaction" in {
```
maasg: `consume-transform-produce` ?

core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -418,5 +420,109 @@ class IntegrationSpec extends TestKit(ActorSystem("IntegrationSpec"))
...
+        val restartSource = RestartSource.onFailuresWithBackoff(
+          minBackoff = 0.1.seconds,
+          maxBackoff = 1.seconds,
+          randomFactor = 0.2
+        ) { () =>
+          restartCount += 1
+          Consumer.transactionalSource(consumerSettings, TopicSubscription(Set(sourceTopic)))
+            .filterNot(_.record.value() == InitialMsg)
+            .map { msg =>
+              if (msg.record.value().equals("500") && restartCount < 2) {
```
maasg: the `producer` above looks like it produces integers (`1 to 1000`) but this check is against a `String` value. (maybe there's some .toString transformation in between, but it's not obvious from the test)

docs/src/main/paradox/transactions.md
```diff
@@ -0,0 +1,72 @@
...
+
+Only use this source if you have the intention to connect it to `Producer.transactionalFlow` or `Producer.transactionalSink`.
+
+## Transactional Sink and Flow
+
+The `Producer.transactionalSink` is similar to the `Consumer.commitableSink` in that messages will be automatically committed as part of a transaction.  The `Producer.transactionalSink` or `Producer.transactionalFlow` are required when connecting a consumer to a producer to achieve a transactional workflow.
+
+They override producer properties `enable.idempotence` to `true` and `max.in.flight.requests.per.connection` to `1` as required by the Kafka producer to enable transactions.
+
+A transactional ID must be defined and unique for each instance of the application.
```
maasg: 'transactional ID' - I'd certainly look for another name. I find it confusing with the ubiquitous  `transactionId`

docs/src/test/scala/sample/scaladsl/TransactionsExample.scala
```diff
@@ -0,0 +1,59 @@
...
+    // #transactionalSink
+
+    terminateWhenDone(control)
+  }
+}
+
+class TransactionsFailureRetryExample extends ConsumerExample {
+  def main(args: Array[String]): Unit = {
+    // #transactionalFailureRetry
+    var innerControl = null.asInstanceOf[Control]
```
maasg: consider 
```
var innerControl : Control = _
```  
to avoid the noisy `asInstanceOf`

docs/src/test/scala/sample/scaladsl/TransactionsExample.scala
```diff
@@ -0,0 +1,59 @@
...
+        }
+        // side effect out the `Control` materialized value because it can't be propagated through the `RestartSource`
+        .mapMaterializedValue(innerControl = _)
+        .via(Producer.transactionalFlow(producerSettings, "transactional-id"))
+    }
+
+    stream.runWith(Sink.ignore)
+
+    // Add shutdown hook to respond to SIGTERM and gracefully shutdown stream
+    Runtime.getRuntime.addShutdownHook(new Thread { Await.result(innerControl.shutdown(), 10.seconds) })
```
maasg: consider: 
```
sys.ShutdownHookThread {
  Await.result(innerControl.shutdown(), 10.seconds)
}
```

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -6,126 +6,284 @@
...
       }
+    })
 
-      setHandler(out, new OutHandler {
-        override def onPull() = {
-          tryPull(in)
+    def produce(msg: Message[K, V, P]): Unit = {
+      val r = Promise[Result[K, V, P]]
+      producer.send(msg.record, new Callback {
+        override def onCompletion(metadata: RecordMetadata, exception: Exception) = {
```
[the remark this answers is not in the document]
seglo: Agreed, but I'm not sure there's much value in doing it in this one use case where we handle a producer [`Callback`](https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/producer/Callback.html).  The `produce` method's code was moved, but otherwise unchanged from its original implementation.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -6,126 +6,284 @@
...
-            // we do not have to check if producer was already closed in send-callback as `flush()` and `close()` are effectively no-ops in this case
-            producer.flush()
-            producer.close(closeTimeout.toMillis, TimeUnit.MILLISECONDS)
-            log.debug("Producer closed")
-          }
-          catch {
-            case NonFatal(ex) => log.error(ex, "Problem occurred during producer close")
-          }
+    private def onCommitInterval(): Unit = {
+      if (!commitInProgress) {
```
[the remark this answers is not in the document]
seglo: `onCommitInterval` is called on the normal commit interval and when draining producer ack's.  The guard is just to prevent re-assigning `commitInProgress = true` in the latter case, but it doesn't really matter since its threadsafe anyway.  I could have created a separate schedule, but since the callback is the same (except to assign `commitInProgress`) I just re-used it.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -6,126 +6,284 @@
...
+        commitInProgress = true
+      }
+      val awaitingConfirmationCount = awaitingConfirmation.get
```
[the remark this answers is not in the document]
seglo: Yes, this was a leftover from earlier troubleshooting.

+1

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -6,126 +6,284 @@
...
+  }
+}
+
+private[kafka] trait ProducerCompletionState {
+  def onCompletionSuccess(): Unit = {}
+  def onCompletionFailure(ex: Throwable): Unit = {}
+}
+
+private[kafka] trait MessageCallback[K, V, P] {
+  def awaitingConfirmation: AtomicInteger
```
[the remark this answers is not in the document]
seglo: +1

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -6,126 +6,284 @@
...
+  def onMessageAckCb: AsyncCallback[Message[K, V, P]]
+}
+
+private[kafka] class TransactionOffsetBatch() {
+  private val offsets = mutable.Map[GroupTopicPartition, Long]()
```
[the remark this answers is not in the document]
seglo: We don't need to be concerned with concurrent modifications because this Map is updated from the `GraphStateLogic` which ensures threadsafety for us.  We could still make `TransactionOffsetBatch` immutable, but we effectively just push that mutation down `TransactionProducerStageLogic` which may be a better place for it anyway.  I'll consider this some more.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -6,126 +6,284 @@
...
+
+  def nonEmpty: Boolean = offsets.nonEmpty
+  def clear(): Unit = offsets.clear()
+
+  def offsetMap(): (String, JMap[TopicPartition, OffsetAndMetadata]) = {
+    val groups = offsets.groupBy {
+      case (groupTopicPartition, _) => groupTopicPartition.groupId
+    }.keys
+
+    require(groups.size == 1, "Transactional batch must contain messages from exactly 1 consumer group. " +
```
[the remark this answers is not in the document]
seglo: Agreed.  I'm not really happy with having to check this at runtime at all, but at the time I was thinking I could share code with the `CommittableMessage` which allows for multiple groups.  It also provides a way to capture the upstream consumer group ID without having to provide it to a `transactionalSource` or transactionalFlow` explicitly.

For now I'll take your suggestion and add a check to the write path as well (`update` method).

core/src/main/scala/akka/kafka/javadsl/Producer.scala
```diff
@@ -73,6 +73,26 @@ object Producer {
       .mapMaterializedValue(_.toJava)
       .asJava
 
+  /**
+   * Sink that is aware of the [[ConsumerMessage.TransactionalMessage#PartitionOffset]] from a [[Consumer]].  It will
+   * initialize, begin, produce, and commit the consumer offset as part of a transaction.
+   */
+  def transactionalSink[K, V](
+    settings: ProducerSettings[K, V],
+    transactionalId: String
```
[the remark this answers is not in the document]
seglo: It is weird, but that's what Kafka defined the `Producer` config key as :)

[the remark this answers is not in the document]
seglo: `transactional.id`

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
seglo: This will never change (at least not with current releases of Kafka) and is only used with `transactionalSink` and `transactionalFlow`, so I think it makes sense to keep it close to the implementation.

seglo: 1.toString +1

core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
seglo: Yes.  That's how it's referred to in KIP-98.  I'll update all references +1

core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
seglo: Yes, the existing integration test infrastructure uses key, value serdes of `Array[Byte]`, `String`.

docs/src/main/paradox/transactions.md
seglo: This is in reference to the `transactional.id` kafka producer config.

docs/src/test/scala/sample/scaladsl/TransactionsExample.scala
seglo: That's actually not possible (results in a compiler error) because `innerControl` is a variable.  See this SO answer for a summary as to why: https://stackoverflow.com/a/36955530/895309

docs/src/test/scala/sample/scaladsl/TransactionsExample.scala
seglo: Nice +1

seglo:
Thanks for the thorough review @maasg !  I've addressed much of it in a commit I just pushed.  I'm going to tackle some of the re-factorings you suggested offline tomorrow (FTR: unit tests, encapsulate transaction state with an ADT).

docs/src/test/scala/sample/scaladsl/TransactionsExample.scala
maasg: Interesting. I didn't know about the scope difference. Thanks for the link.
How about:
```
var innerControl = null.asInstanceOf[Control]
// vs
var innerControl:Control = null
```
?

seglo: Yes, that works.

docs/src/main/paradox/transactions.md
maasg: OK if it's known in the "local dialect" :)

seglo:
@patriknw @ktoso Hey.  I've been tinkering with this PR and I think it's ready for more eyes.  I've squashed and rebased.  I realise this is a bit of a monster so I'm not looking for any kind of quick turnaround, but I just wanted your advice on who might be an ideal reviewer so I can poke the right people.

core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -28,6 +28,15 @@ object ConsumerMessage {
...
   )
 
+  /**
+   * Output element of `transactionalSource`.
+   * The offset is automatically committed as by the Producer
+   */
+  final case class TransactionalMessage[K, V](
+      record: ConsumerRecord[K, V],
+      partitionOffset: PartitionOffset
+  )
```
fchaillou: Did you forget to make this class implement `PartitionOffsetMessage` ?

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -6,126 +6,299 @@
...
+    override def updated(partitionOffset: PartitionOffset): TransactionBatch = new NonemptyTransactionBatch(partitionOffset)
+  }
+
+  class NonemptyTransactionBatch(
+      head: PartitionOffset,
+      tail: Map[GroupTopicPartition, Long] = Map[GroupTopicPartition, Long]())
+    extends TransactionBatch {
+    private val offsets = tail + (head.key -> head.offset)
+
+    def group: String = offsets.keys.head.groupId
```
fchaillou: why not only `head.key.groupId` ?

core/src/main/scala/akka/kafka/ProducerSettings.scala
```diff
@@ -146,7 +148,8 @@ final class ProducerSettings[K, V](
     val valueSerializerOpt: Option[Serializer[V]],
     val closeTimeout: FiniteDuration,
     val parallelism: Int,
-    val dispatcher: String
+    val dispatcher: String,
+    val eosCommitIntervalMs: Long
```
fchaillou: I don't know if that makes sense but in some use case the user might want to be able to decide to commit a transaction after a specific number of input message were fully processed.
I think it should be doable by tracking an `awaitingTransactionCommit` the same way the current `awaitingConfirmation` works.
This way you could support the use case where the user want to make sure each input message generate a transaction (or more than one if wanted)

core/src/main/scala/akka/kafka/ConsumerMessage.scala
seglo: This is an artifact of an earlier implementation attempt where I tried to generalize consumer group committable messages and transactional messages by extracting `PartitionOffsetMessage` into a trait.  I didn't pursue this option because I thought it would be more appropriate for the `ProducerStage` to control the transaction commits, like they're done in Kafka Streams.  For now I'll remove the `PartitionOffsetMessage` since it's not being used.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
seglo: :+1:

fchaillou: thinking of it you could probably simplify the NonemptyTransactionBatch by directly storing the offsets and groupId :

```
class EmptyTransactionBatch extends TransactionBatch {
    override def updated(partitionOffset: PartitionOffset): TransactionBatch = new NonemptyTransactionBatch(partitionOffsetkey.groupId, Map(partitionOffset.key -> partitionOffset.offset))
}

 class NonemptyTransactionBatch(val group : String, offsets : Map[GroupTopicPartition, Long]) {
         def offsetMap(): Map[TopicPartition, OffsetAndMetadata] = offsets.map {
           case (gtp, offset) => new TopicPartition(gtp.topic, gtp.partition) -> new OffsetAndMetadata(offset + 1)
        }

override def updated(partitionOffset: PartitionOffset): TransactionBatch = {
     require(
          group == partitionOffset.key.groupId,
        s"Transaction batch must contain messages from exactly 1 consumer group. $group != ${partitionOffset.key.groupId}")
 
         val updatedOffsets = offsets + (partitionOffset.key -> partitionOffset.offset)

      new NonemptyTransactionBatch(group, updateOffsets)
    }
}


}

File: core/src/main/scala/akka/kafka/ProducerSettings.scala (hunk 16 above)
Comment (seglo) [reply core/src/main/scala/akka/kafka/ProducerSettings.scala]: That's a fair point.  There are several reasons I didn't implement this approach, but they may not be too compelling.  I welcome a challenge here.

1. [KIP-129](https://cwiki.apache.org/confluence/display/KAFKA/KIP-129%3A+Streams+Exactly-Once+Semantics) nor the current Kafka Streams implementation (AFAICT) exposes a way to batch a transaction in this manner.
2. In a Consume-Transform-Produce workflow we have strong guarantees that the source message is already persisted and possibly replicated, so it's not critical we must process the message successfully the first time.  If the batch fails for some transient reason then it can be processed the next time.  I acknowledge this reason is more of an opinion than fact since there are other things that can go wrong, such as the retention period for that source message elapsing, etc.
3. It would be a little more implementation work and I wanted to see what people thought of this approach (which mirrors what's done in Kafka Streams) before I extended the implementation.

**Comment by seglo** [comment]
Thanks a lot for the review @fchaillou !  I provided feedback and added a commit incorporating some of your suggestions.

File: core/src/main/scala/akka/kafka/ProducerSettings.scala (hunk 16 above)
Comment (fchaillou) [reply core/src/main/scala/akka/kafka/ProducerSettings.scala]: Your points are completely valid.
Let's get this implementation merged and after I could take a look at adding this feature if I really feel it important !

File: core/src/main/scala/akka/kafka/internal/ProducerStage.scala (hunk 15 above)
Comment (seglo) [reply core/src/main/scala/akka/kafka/internal/ProducerStage.scala]: Passing in the group name may be a good idea.  I want to give the `NonemptyTransactionBatch` the property of never being empty, so I was inspired by the [`NonEmptyList` implementation in cats](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/NonEmptyList.scala) to provide the `head` element in the default constructor.

File: docs/src/main/paradox/transactions.md [hunk 17]
```diff
@@ -0,0 +1,72 @@
...
+Java
+: @@ snip [transactionalFailureRetry](../../test/java/sample/javadsl/TransactionsExample.java) { #transactionalFailureRetry }
+
+## Caveats
+
+There are several scenarios that this library's implementation of Kafka transactions does not automatically account for.
+
+All of the scenarios covered in the @ref[At-Least-Once Delivery documentation](atleastonce.md) (Multiple Effects per Commit, Non-Sequential Processing, and Conditional Message Processing) are applicable to transactional scenarios as well.
+
+Only one application instance per `transactional.id` is allowed.  If two application instances with the same `transactional.id` are run at the same time then the instance that registers with Kafka's transaction coordinator second will throw a `ProducerFencedException` so it doesn't interfere with transactions in process by the first instance.  To distribute multiple transactional workflows for the same subscription the user must manually subdivide the subscription across multiple instances of the application.  This may be handled internally in future versions.
```
rafalmag: In KafkaStreams they add a sufix with source topic name and partition number to transactional.id . It was the most challenging thing that I tried to explain in [comment](https://github.com/akka/reactive-kafka/issues/369#issuecomment-364974971) as in my opinion in most use cases there would be multiple "processors" that would handle the same work (same input topics). We cannot generate new transactional.id for new processing instance as then zombie fencing would not work either.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
```diff
@@ -73,6 +73,26 @@ object Producer {
...
+  ): Sink[Message[K, V, ConsumerMessage.PartitionOffset], CompletionStage[Done]] =
```
rafalmag: There are 2 other common use cases - for single input message:
- someone would like to do the message processing and only commit the input message, so there would be no output kafka message
- someone would like to emit multiple messages (possible for different output topics)

This could be modeled as `Sink[Seq[Message[K, V, ConsumerMessage.PartitionOffset]], CompletionStage[Done]]`

docs/src/main/paradox/transactions.md
rafalmag: This might still break exactly once processing. Simple example:
- single group name
- single topic T with 3 partitions: p1, p2, p3. Each partition has a lot of messages, p1 has p1m1,p1m2, ... p2 has p2m1,p2m2, etc. -  offset 0
- processing on 2 hosts/nodes/processes: A,B

After a system start (deployment/restart, etc.):
- node A registers, kafka assigns p1,p2,p3 partitions and the p1m1, p2m1, p3m1 messages are pulled.
- node B registers, kafka would reassign the partitions, for example A to p1,p2 and B to p3.
Process B could potentially pull the message from p3 - so it would process p3m1.

And now the race between A and B happens - the outcome should be that the processing of p3m1 on one of the nodes should finish successfully and the other should fail (InvalidProducerEpoch or ProducerFenced). 
In order to make this mechanism work both processes should use the same transactional.id - so during the partition attach and initTransaction the transactional.id epoch is bumped up. 

Because of that Transactional.id cannot be host related. 
See also https://www.confluent.io/blog/transactions-apache-kafka/ - but that does not give any solution... I cannot find out better solution than just coping the one from kafka streams...

This is _extremely_ important as when this bit is broken then we still have _at least once_ processing instead of _exactly once_.

rafalmag: Please see also [my comment](https://github.com/akka/reactive-kafka/issues/369#issuecomment-380473894) here.

seglo: Thanks a lot for joining this discussion @rafalmag , I was hoping you would provide your feedback!  Distributed transactional workflows are a hard problem to solve.  The implementation in KafkaStreams is complex.  I intentionally avoided this use case because I wanted to get the basic workflow for a single application instance working correctly.  I'm on the fence about implementing distributed transactional support in reactive-kafka because I think it can be handled by the client configuring appropriate consumer group subscriptions and the transaction.id themselves. This path loses out on the advantages of consumer group rebalancing, but does satisfy distributed transaction workflows.  Do you agree?

That said, I agree that it would be wonderful to have some way to support the distributed use case, so let's study our options.

In KafkaStreams the task's Producer's `transactional.id` is defined in [`StreamThread`](https://github.com/apache/kafka/blob/2e5d4af83f6b9f87fc84c0ff10462a5414ff8cb4/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java#L425).  It's a concatenation of the KafkaStreams config `application.id` and the `TaskId`. If you review [KIP-129](https://cwiki.apache.org/confluence/display/KAFKA/KIP-129%3A+Streams+Exactly-Once+Semantics) we see the design specifies one `KafkaProducer` per task.  So what is a task?

In Kafka Streams a task has an assigned set of topic partitions from the consumer group.  A task can contain multiple `StreamThread`'s to distribute work internally on the host machine.  A `PartitionGrouper` is used to assign tasks to partitions and the [`DefaultPartitionGrouper`](https://github.com/apache/kafka/blob/f26fbb9adcd66a31740da5c99f14a108bbc24304/streams/src/main/java/org/apache/kafka/streams/processor/DefaultPartitionGrouper.java) assigns each `TaskId` to a subset of topic partitions.  

How the `TaskId`'s are created is not clear to me, but it seems to be related to how many `Source`'s are available for the stream.  I think this may translate to the number of active consumers there are for a consumer group.  If that's the case then somehow Kafka Streams is aware of other consumers and assigns them ID's, so the assignment can happen, probably thru some metadata Kafka topic.

Obviously more research needs to be done here, but I agree that the simplest approach is what you already suggested by assigning a producer `transaction.id` per consumed topic partition.  I'll do a little more digging in the Kafka Streams codebase to see how `TaskId`'s are generated, but if it involves out of band communication to other app instances then I think the solution maybe too high level and complex for the reactive-kafka project.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
seglo: Interesting.  I didn't consider this use case.  I'll think about how to support it.

docs/src/main/paradox/transactions.md
rafalmag: @seglo I really like the idea of `TransactionalMessage` in
```  
  def transactionalSource[K, V](settings: ConsumerSettings[K, V], subscription: Subscription): Source[TransactionalMessage[K, V], Control]
```
and
```
def transactionalSink[K, V](settings: ProducerSettings[K, V], transactionalId: String): 
Sink[Message[K, V, ConsumerMessage.PartitionOffset], Future[Done]] 
```

Rebalancing can be detected by consumer in source, transaction id should be set for producer in sink, but this approach really decouples source and sink. That is why I suggested in [comment](https://github.com/akka/reactive-kafka/issues/369#issuecomment-364974971) to implement something that receives the logic as a flow and does the creation of consumer, producer and necessary feedback loop in its internals, eg:
```
def withTransactions(consumerSettings: ConsumerSettings, producerSettings: ProducerSettings, flow: Flow[ConsumerRecord, Seq[ProducerRecord], Mat]): RunnableGraph[Mat]
```

Previously I thought that method could be implemented using:

```  
  def plainPartitionedSource[K, V](settings: ConsumerSettings[K, V], subscription: AutoSubscription): Source[(TopicPartition, Source[ConsumerRecord[K, V], NotUsed]), Control]
```	

so here we get a new substream for each topic-partition pair, so we can attach the flow and then some lazy producer (that would be created taking as an argument the TopicPartition). 

BTW: Maybe a corresponding method based on `TransactionalMessage` could be useful for that:
```
  def transactionalSource[K, V](settings: ConsumerSettings[K, V], subscription: AutoSubscription): Source[(TopicPartition, Source[TransactionalMessage[K, V], NotUsed]), Control]
```

In the sourcecode of `plainPartitionedSource` you will find a mechanism of emitting a new substream after rebalancing and a closing the old stream (in case of detach). 

Regarding:
> I'll also do a little more digging in the Kafka Streams codebase to see how TaskId's are generated, but if it involves out of band communication to other app instances then I think the solution maybe too high level and complex for the reactive-kafka project.

[TaskId](https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/processor/TaskId.java) in the Kafka Streams is just topicGorupId and partition in a form (`t_p`)

For setting the task Id no extra communication is required. The [org.apache.kafka.clients.consumer.Consumer](https://github.com/apache/kafka/blob/23d01c805bef7504abfa83ecac7e384d121a583a/clients/src/main/java/org/apache/kafka/clients/consumer/Consumer.java#L56) has a nice method:
```
public void subscribe(Collection<String> topics, ConsumerRebalanceListener callback);
```
and [ConsumerRebalanceListener](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.java) could be used to listen to the rebalancing events to know what partitions are revoked and what new are created.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
seglo: @rafalmag Couldn't this use case be satisfied with non-Transactional consumer group offset commits after message is processed?  What benefit does the user get when they actually use a transaction?

docs/src/test/java/sample/javadsl/ConsumerExample.java
```diff
@@ -388,4 +388,20 @@ public void demo() {
...
 
+  public void demo() {
+    // #transactionalSource
+    Consumer
+      .transactionalSource(consumerSettings, Subscriptions.topics("source-topic"))
+      .via(business())
+      .map(msg ->
+        new ProducerMessage.Message<byte[], String, ConsumerMessage.PartitionOffset>(
+          new ProducerRecord<>("sink-topic", msg.record().value()), msg.partitionOffset()))
+      .runWith(Sink.ignore(), materializer);
```
ennru: Does it make sense to show a `transactionalSource` without a `transactionalSink`? Is this snippet even referenced?

core/src/main/resources/reference.conf
```diff
@@ -16,6 +16,9 @@ akka.kafka.producer {
   # will be used.
   use-dispatcher = "akka.kafka.default-dispatcher"
 
+  # The time interval to commit a transaction when using the `transactionalSink` or `transactionalFlow`
+  eos-commit-interval-ms = 100
```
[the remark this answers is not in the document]
seglo: :+1:

docs/src/main/paradox/transactions.md
```diff
@@ -0,0 +1,72 @@
+# Transactions
+
+Kafka Transactions provide guarantees that messages processed in a consume-transform-produce workflow (consumed from a source topic, transformed, and produced to a destination topic) are processed exactly once or not at all.  This is achieved through coordination between the Kafka consumer group coordinator, transaction coordinator, and the consumer and producer clients used in the user application.  The Kafka producer marks messages that are consumed from the source topic as "committed" only once the transformed messages are successfully produced to the sink.
```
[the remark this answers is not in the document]
seglo: I'll add a sentence in the introduction paragraph referencing the KIP for full details.  This link was already included in the Further Reading section, but there's value in referencing it early on too.

docs/src/test/java/sample/javadsl/ConsumerExample.java
seglo: You're right, this wasn't used and wouldn't make much sense anyway.  It was a remnant from an earlier documentation attempt (documenting it across existing consumer and producer doc pages) that I later abandoned.

seglo:
Thanks for the review @ennru.  I've incorporated your feedback.

core/src/main/scala/akka/kafka/javadsl/Consumer.scala
```diff
@@ -104,6 +104,15 @@ object Consumer {
...
       .asJava
 
+  /**
+   * Transactional source to setup a stream for Exactly Only Once (EoS) kafka message semantics.  To enable EoS it's
+   * necessary to use the [[Producer#transactionalSink]] or [[Producer#transactionalFlow]] (for passthrough).
+   */
+  def transactionalSource[K, V](consumerSettings: ConsumerSettings[K, V], subscription: Subscription): Source[TransactionalMessage[K, V], Control] =
+    scaladsl.Consumer.transactionalSource(consumerSettings, subscription)
+      .mapMaterializedValue(new WrappedConsumerControl(_))
+      .asJava
```
ennru: I expect a type error here as we upgraded to Akka 2.5.12.
Could you rebase, please.

seglo:
Thanks @ennru .  I rebased and successfully ran the testsuite.  I did not see the type error in `javadsl.Consumer` that you anticipated.  Let me know if I missed something here.

On an unrelated note I noticed that the `resume consumer from committed offset after retention period` test in `IntegrationSpec` takes over 2 minutes to run because of the `Thread.sleep`'s, which significantly bloats the time it takes to run the testsuite especially when targeting other JVM/Scala versions in Travis.  I assume this is an intentional test to show how to recover from an offset after consumer group offset information is lost, but maybe we can tag it differently so the `IntegrationSpec` can be run more quickly during development?  Another option may be to decrease the retention period for the internal consumer offsets topic so we don't need to wait as long.

ennru:
Great, thank you. If the compiler is happy, I'm happy.

Yes, that test is a bit in the way. The retention period in Kafka can't be less than one minute, that's why it takes so long time. But you're right, we should move it to a separate class.

ennru:
Thank you for this great new functionality!

seglo:
Great!  Thank you and all the reviewers :)

core/src/main/scala/akka/kafka/ProducerSettings.scala
SemanticBeeng: Also think we need extra support on top of this current Kafka transactions semantics to implement things like `sagas`, especially in the context of "data pipelines" and not just "microservices".

https://microservices.io/patterns/data/saga.html

cc @seglo @fchaillou

Issue #421, opened 2018-04-03
Actually expose ability to set rebalance listener


ktoso:
Seems this was attempted in 2016 in https://github.com/akka/reactive-kafka/issues/93 however something was not quite done "all the way" it seems.

Always the hardcoded 

```
private[kafka] def rebalanceListener(onAssign: Set[TopicPartition] => Unit, onRevoke: Set[TopicPartition] => Unit): ListenerCallbacks =
    ListenerCallbacks(onAssign, onRevoke)
```

```
    def rebalanceListener =
      KafkaConsumerActor.rebalanceListener(tps => partitionAssignedCB.invoke(tps), partitionRevokedCB.invoke)
```

is used instead of one provided by the user.

It seems like it was intended to be able to pass it in a Subscription, but was never done?

ktoso:
cc @kciesielski @patriknw 

I'll work on this as got someone requesting it

PR #423, opened 2018-04-04
Add ability to set rebalance listeners from user code #421

ktoso:
Resolves https://github.com/akka/reactive-kafka/issues/421

Let me know if I missed anything.
Will add docs now

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -9,6 +9,7 @@ import java.util.Optional
 import java.util.concurrent.TimeUnit
 
 import akka.actor.ActorSystem
+import akka.japi.function.Procedure
 import akka.kafka.KafkaConsumerActor.ListenerCallbacks
 import akka.kafka.internal.ConfigSettings
 import com.typesafe.config.Config
@@ -23,7 +24,11 @@ import scala.concurrent.duration._
 
 sealed trait Subscription {
   def rebalanceListenerCallbacks: Option[ListenerCallbacks]
+  /** Sets callbacks ti be invoked when the consumer related to this subscription gets rebalancing events */
   def withRebalanceListenerCallbacks(onAssign: Set[TopicPartition] => Unit, onRevoke: Set[TopicPartition] => Unit): Subscription
+  /** Java API: Sets callbacks ti be invoked when the consumer related to this subscription gets rebalancing events */
+  final def withRebalanceListenerCallbacksJavadsl(onAssign: Procedure[java.util.Set[TopicPartition]], onRevoke: Procedure[java.util.Set[TopicPartition]]): Subscription =
+    withRebalanceListenerCallbacks(s ⇒ onAssign(s.asJava), s ⇒ onRevoke(s.asJava))
 }
 sealed trait ManualSubscription extends Subscription {
   def withRebalanceListenerCallbacks(onAssign: Set[TopicPartition] => Unit, onRevoke: Set[TopicPartition] => Unit): ManualSubscription
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -9,8 +9,6 @@ import java.util.Optional
 import java.util.concurrent.TimeUnit
 
 import akka.actor.{ActorRef, ActorSystem}
-import akka.japi.function.Procedure
-import akka.kafka.KafkaConsumerActor.ListenerCallbacks
 import akka.kafka.internal.ConfigSettings
 import com.typesafe.config.Config
 import org.apache.kafka.clients.consumer.{Consumer, ConsumerConfig, KafkaConsumer}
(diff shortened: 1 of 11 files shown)


ktoso:
Added docs and javadsl.

Please review; we could cut 0.20 even tomorrow if this looks good :L)

ktoso:
consumer_ _akka_streams_kafka

I could use some help on docs there, any better wording to explain the use case?

ktoso:
Rebased

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -20,21 +22,50 @@ import scala.collection.JavaConverters._
...
 
-sealed trait Subscription
-sealed trait ManualSubscription extends Subscription
-sealed trait AutoSubscription extends Subscription
+sealed trait Subscription {
+  def rebalanceListenerCallbacks: Option[ListenerCallbacks]
+  /** Sets callbacks ti be invoked when the consumer related to this subscription gets rebalancing events */
+  def withRebalanceListenerCallbacks(onAssign: Set[TopicPartition] => Unit, onRevoke: Set[TopicPartition] => Unit): Subscription
+  /** Java API: Sets callbacks ti be invoked when the consumer related to this subscription gets rebalancing events */
+  final def withRebalanceListenerCallbacksJavadsl(onAssign: Procedure[java.util.Set[TopicPartition]], onRevoke: Procedure[java.util.Set[TopicPartition]]): Subscription =
```
patriknw: the api naming is a mess here and there, we should take another round of cleaning up the API. Created https://github.com/akka/reactive-kafka/issues/424

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -100,6 +105,11 @@ private[kafka] abstract class SingleSourceLogic[K, V, Msg](
 
   }
 
+  private def invokeUserOnAssign(set: Set[TopicPartition]): Unit =
+    subscription.rebalanceListenerCallbacks.foreach(_.onAssign(set))
```
patriknw: do we need to protect ourselves more for exceptions thrown by user code?

ktoso:
Should we rather do an API based on an actor ref that should receive those messages? That's simple to do as well if we want

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
ktoso: We do:

```
try invokeUserOnAssign(newTps)
      finally requestMessages()
```

core/src/main/scala/akka/kafka/ConsumerSettings.scala
ktoso: Yeah, cleanup would be nice... I have not found a nice way without breaking API though

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
patriknw: but it would still result in restart of the KafkaConsumerActor if exception thrown

patriknw:
If they are ok with an async message based API I think that would be more safe.

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
ktoso: Hmm right; I can add a catch + log or move to the actor ref style, any preference?

ktoso:
Yeah I think it should be fine with messages. I'll change the API then

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -20,21 +22,51 @@ import scala.collection.JavaConverters._
 import scala.compat.java8.OptionConverters._
 import scala.concurrent.duration._
 
-sealed trait Subscription
-sealed trait ManualSubscription extends Subscription
-sealed trait AutoSubscription extends Subscription
+sealed trait Subscription {
+  def rebalanceListener: Option[ActorRef]
+  /** Configure this actor ref to receive [[akka.kafka.ConsumerRebalanceEvent]] signals */
```
patriknw: add this doc at more places

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -20,21 +22,51 @@ import scala.collection.JavaConverters._
...
+final case class TopicPartitionsRevoked(sub: Subscription, topicPartitions: Set[TopicPartition]) extends ConsumerRebalanceEvent
 
 object Subscriptions {
-  private[kafka] final case class TopicSubscription(tps: Set[String]) extends AutoSubscription
-  private[kafka] final case class TopicSubscriptionPattern(pattern: String) extends AutoSubscription
-  private[kafka] final case class Assignment(tps: Set[TopicPartition]) extends ManualSubscription
-  private[kafka] final case class AssignmentWithOffset(tps: Map[TopicPartition, Long]) extends ManualSubscription
-  private[kafka] final case class AssignmentOffsetsForTimes(timestampsToSearch: Map[TopicPartition, Long]) extends ManualSubscription
+  /** INTERNAL API */
+  private[kafka] final case class TopicSubscription(tps: Set[String])(val rebalanceListener: Option[ActorRef]) extends AutoSubscription {
```
patriknw: why second parameter list?

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -70,17 +70,22 @@ private[kafka] abstract class SingleSourceLogic[K, V, Msg](
...
+      notifyUserOnAssign(newTps)
       requestMessages()
     }
 
     val partitionRevokedCB = getAsyncCallback[Set[TopicPartition]] { newTps =>
       tps --= newTps
       log.log(partitionLogLevel, "Revoked partitions: {}. All partitions: {}", newTps, tps)
-      requestMessages()
+      notifyUserOnRevoke(newTps)
+      finally requestMessages()
```
patriknw: finally not needed any more (and where is the `try`?

docs/src/main/paradox/consumer.md
```diff
@@ -186,3 +186,15 @@ Scala
 
 Java
 : @@ snip [consumerMetrics](../../test/java/sample/javadsl/ConsumerExample.java) { #consumerMetrics }
+
+
+## Listening for rebalance events
+
+You may set up an rebalance event listener actor that will be notified when your consumer will be assigned or revoked
```
patriknw: two callbacks?
mention the message types instead

ktoso: yeap was fixing; fixed

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
ktoso: yep

core/src/main/scala/akka/kafka/ConsumerSettings.scala
ktoso: wanted to avoid including it in the pattern match but maybe does not matter since internal APIs anyway

core/src/main/scala/akka/kafka/ConsumerSettings.scala
ktoso: ok

ktoso:
Thanks for review, we arrived at a nice API I think

patriknw:
👍 thanks for changing

PR #425, opened 2018-04-05
show we support 1.0

ktoso:


johanandren:
(With the precondition that the statement is true, because I'm not sure ;) )

ktoso:
Yeah we bumped and it is compatible: https://cwiki.apache.org/confluence/display/KAFKA/Compatibility+Matrix

Issue #422, opened 2018-04-03
Expose metadata retrieving functions of Consumer in KafkaConsumerActor


dsitte-h:
List of functions to expose:
- `listTopics()`
- `partitionsFor()`
- `beginningOffsets()`
- `endOffsets()`
- `offsetsForTimes()`

Example usage:
```scala
//Consumer is represented by actor
val consumer: ActorRef = system.actorOf(KafkaConsumerActor.props(consumerSettings))

// Retrieve metadata
val endOffsets = for {
  partitions <- consumer ? GetPartitonsFor("topic").mapTo[List[PartitionInfo]]
  endOffsets <- consumer ? endOffsets(partitions).mapTo[Map[TopicPartition, Long]]
} yield {
  endOffsets
}

// wait for endOffsets Future

//Manually assign topic partition to it
Consumer
  .plainExternalSource[Array[Byte], String](consumer, Subscriptions....)
  .via(business)
  .runWith(Sink.ignore)
```
The point is being able to use the same underlying KafkaConsumer for retrieving metadata and consuming.  

A side-effect is that Kafka connection is established upon metadata retrieval, possible connection errors can be (and need to be!) handled before the consumer graph is even created.

Willing to contribute PR.

schrepfler:
It would be nice if we would be able to extract headers as well.

dsitte-h:
Headers are per message, you can access them on `ConsumerRecord`:

https://github.com/apache/kafka/blob/88da81b9454183955f34b822588f0f1ee7e9dcae/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecord.java#L155

(Headers were introduced in Kafka v0.11)

2m:
Fixed in https://github.com/akka/reactive-kafka/pull/491

PR #430, opened 2018-04-16
Expose metadata retrieving functions of Consumer in KafkaConsumerActor #422

dsitte-h:
See #422 for details, added tests for example usage.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import scalariform.formatter.preferences._
 
 name := "akka-stream-kafka"
 
-val akkaVersion = "2.5.12"
+val akkaVersion = "2.5.13"
 val kafkaVersion = "1.0.1"
 val kafkaVersionForDocs = "10"
 


core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -14,23 +14,42 @@ import java.util.regex.Pattern
...
+import scala.util.Try
 
 object KafkaConsumerActor {
   case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
 
   def props[K, V](settings: ConsumerSettings[K, V]): Props =
     Props(new KafkaConsumerActor(settings)).withDispatcher(settings.dispatcher)
 
+  //metadata fetching
+  //NOTE: These block the actor loop!
```
ennru: The `KafkaConsumerActor` is configured to run on its own dispatcher, so just as the other remote calls to Kafka the blocking happens within a designated thread pool.
Please add some ScalaDoc with a pointer to relevant Kafka documentation on Metadata.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -200,6 +219,46 @@ private[kafka] class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V])
...
+
+    case Metadata.GetBeginningOffsets(partitions) =>
+      sender() ! Metadata.BeginningOffsets(Try {
+        consumer.beginningOffsets(partitions.asJava).asScala.map {
+          case (k, v) => k -> (v: Long)
+        }.toMap
+      })
+
+    case Metadata.GetEndOffsets(partitions) =>
+      sender ! Metadata.EndOffsets(Try {
```
ennru: Please use the `sender()` notation on all places.

core/src/main/scala/akka/kafka/KafkaConsumerActor.scala
```diff
@@ -14,23 +14,42 @@ import java.util.regex.Pattern
...
   case class StoppingException() extends RuntimeException("Kafka consumer is stopping")
 
   def props[K, V](settings: ConsumerSettings[K, V]): Props =
     Props(new KafkaConsumerActor(settings)).withDispatcher(settings.dispatcher)
 
+  //metadata fetching
+  //NOTE: These block the actor loop!
+  object Metadata {
+    //requests
+    case object ListTopics extends NoSerializationVerificationNeeded
```
ennru: It would be good to introduce some super trait or class for requests and responses respectively. With that in place, the Metadata handling could easily be kept in a separate receive method.

dsitte-h:
Reorganized Metadata in response to inline comments.

What do you mean by Java API?
Request/Response actor messages that are Java compatible?
Should these be changed to POJOs instead of case classes?
What would be the preferred Try equivalent in Java?

(Will rebase to latest master once the details are nailed down.)

ennru:
The Java API would eg. offer the returned data as `java.util.Map`. Case classes are Ok, but create methods in Metadata taking Java structures would be good eg. `def createGetBeginningOffsets(partitions: java.util.Set[TopicPartition])`
For case objects I'd like to have `val createListTopics = ListTopics`.

For Try you could add accessors directly in the case class like `def getResponse(): java.util.Map[TopicPartiton, OffsetAndTimestamp] = response.get.asJava' accompanied with 'isSuccess' and 'isFailure'.

ennru:
And some documentation, probably a new page would be great. :-)

2m:
@dsitte-h, we continue the work on metadata in #491 where Java API is being added.

Issue #432, opened 2018-04-17
Document controlled shutdown of a Consumer stream


svroonland:
A nice addition to the Consumer docs would be how to correctly stop a stream with `Consumer.Control`.

svroonland:
I can submit a PR for this.

patriknw:
Would be great, thanks

ennru:
Fixed with #433

PR #433, opened 2018-04-19
Docs for consumer shutdown

svroonland:
See #432 

This is an addition to the Consumer docs on how to shutdown a consumer stream. This is especially interesting when using batched commits. Code examples are included.

My scala is better than my java, so I'm more than happy to accept improvements/corrections there of course.

### DIFF
--- a/docs/src/test/java/sample/javadsl/ConsumerExample.java
+++ b/docs/src/test/java/sample/javadsl/ConsumerExample.java
@@ -423,7 +423,7 @@ class ShutdownBatchedExample extends ConsumerExample {
         // #streamShutdownBatched
         final DB db = new DB();
 
-        Pair<Pair<Consumer.Control, UniqueKillSwitch>, CompletionStage<Done>> r =
+        Pair<Consumer.Control, CompletionStage<Done>> r =
                 Consumer.committableSource(consumerSettings, Subscriptions.topics("topic1"))
                         .mapAsync(1, msg ->
                                 db.update(msg.record().value()).thenApply(done -> msg.committableOffset()))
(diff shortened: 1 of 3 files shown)


rtimush:
Is `KillSwitch` really necessary here? In my opinion the correct sequence should be
```scala
for {
  _ <- Control.stop()     // stop consuming messages and complete the source
  _ <- done               // wait till the stream is completed
  _ <- Control.shutdown() // stop the consumer
} yield Done
```
This works for me with non-partitioned consumers. For partitioned consumers it doesn't work because the inner sources are not completed when the outer `Control.stop()` is called, but I'd say this should rather be fixed in the library.
Am I missing something here?

svroonland:
@rtimush You're right, after a closer look it turns out to be an error in an integration test that required the `KillSwitch`. I will adjust the examples.

svroonland:
Ready for another review. 

While debugging the KillSwitch related issue in my own application, I gained a better understanding of the shutdown mechanism, which is relfected now in the updated documentation. See also the commit messages.

docs/src/test/java/sample/javadsl/ConsumerExample.java
```diff
@@ -388,4 +386,60 @@ public void demo() {
...
+                db.update(msg.record().value()).thenApply(done -> msg.committableOffset()))
+            .batch(20,
+                first -> ConsumerMessage.emptyCommittableOffsetBatch().updated(first),
+                (batch, elem) -> batch.updated(elem))
+            .mapAsync(3, c -> c.commitJavadsl())
+            .toMat(Sink.ignore(), Keep.both())
+            .run(materializer);
+
+    Consumer.Control control = r.first();
+    CompletionStage<Done> done = r.second();
```
ennru: I wonder if there is a better variable name than `done`, which would make the `thenCompose` read more natural.

svroonland: `streamComplete` would be better. Or do you have something else in mind?

ennru: Yes, I like that.

Issue #435, opened 2018-04-27
Partitioned source is not stopped by ConsumerControl


rtimush:
When `ConsumerControl.stop` is called on the partitioned source, the source itself is completed, but substreams it created are not. This makes it difficult to correctly shutdown a stream.

PR #436, opened 2018-04-27
Add integration tests for stopping consumer

rtimush:
Tests for `ConsumerControl.stop` behavior for simple and partitioned sources. The first test works, the second fails.

Supporting test cases for #435.


ennru:
Thank you for providing this bug report and test case.

PR #453, opened 2018-05-16
Always increments the ProducerStage counter before decrement and check

patrick-premont:
There was a possibility that it would be decremented and checked before the corresponding increment, leading to a failure to check for completion.


patriknw:
Thanks for catching and fixing @patrick-premont

PR #460, opened 2018-05-18
Restructure integration test

ennru:
* Create SpecBase
* Split out Retention Period depending test
* Split out Transactions test
* Parallel execution needs to be false as EmbeddedKafka wants to be a Singleton (even though it uses different ports)

### DIFF
--- a/core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
+++ b/core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
@@ -5,80 +5,76 @@
 
 package akka.kafka.scaladsl
 
-import java.util.UUID
+import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.{ConcurrentLinkedQueue, TimeUnit}
 
 import akka.actor.ActorSystem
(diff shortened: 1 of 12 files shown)


core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -5,113 +5,32 @@
...
-
-  def createConsumerSettings(group: String): ConsumerSettings[Array[Byte], String] = {
-    ConsumerSettings(system, new ByteArrayDeserializer, new StringDeserializer)
-      .withBootstrapServers("localhost:9092")
-      .withGroupId(group)
-      .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
-      .withWakeupTimeout(10 seconds)
-      .withMaxWakeups(10)
-  }
+class IntegrationSpec extends SpecBase(kafkaPort = 9092) {
```
2m: It would be nice to factor port numbers to one single object so it is easier to track what ports are occupied and what are free. Will become espacially handly when writing new specs.

PR #463, opened 2018-05-21
Multi-consumer integration tests

ennru:


core/src/test/scala/akka/kafka/scaladsl/MultiConsumerSpec.scala
```diff
@@ -0,0 +1,142 @@
...
+
+      // produce 10 batches of 10 elements to all topics
+      val batches = 10
+      val batchSize = 10
+      Await.result(produceBatches(topics, batches, batchSize), remainingOrDefault)
+
+      val consumerSettings = consumerDefaults.withGroupId(group1)
+
+      val (control1, probe1) = createProbe(consumerSettings, topics: _*)
+      val (control2, probe2) = createProbe(consumerSettings, topics: _*)
```
2m: Shouldn't this use `group2`?

2m: Ahh, this test is for multiple consumers in a single group. So it is fine here. But line `36` then should be removed?

Issue #451, opened 2018-05-16
wakeup-debug should default to false


agemooij:
We have been noticing a lot of huge stack dumps in our code that uses reactive-kafka, especially when deployed on Kubernetes. It appears these stack dumps are created based on the `wakeup-debug` setting, which defaults to `true`. 

I would suggest that the `wakeup-debug` setting, which can produce enormous amounts of logging volume, should default to false.

agemooij:
To be precise, the setting is `akka.kafka.consumer.wakeup-debug`, with the following definition in reference.conf:

```
  # If enabled log stack traces before waking up the KafkaConsumer to give
  # some indication why the KafkaConsumer is not honouring the poll-timeout
  wakeup-debug = true
```

Shown in the docs here:

https://doc.akka.io/docs/akka-stream-kafka/current/consumer.html#settings

patriknw:
I agree that the right default is `fals`e, but since we got so many issue reports just saying "it doesn't work" we wanted to gather information about what the reasons for the blocking would be.

agemooij:
OK, so you set it to true on purpose to collect data? When are you planning o set it back to false?

We have explicitly set the setting to `false` for now, so big rush here. Just curious.

2m:
Instead of printing thread-dump on every wakeup, it will now do it when number of wakeups is more than max-wakeups/2. So we can keep wakeup-debug set to `true` by default to still have the option to collect the data and not being that verbose in normal operations.

mr-git:
what will happen, if `max-wakeups` is set to `0`?

PR #464, opened 2018-05-21
Poll wakeups: don't log stack traces until max-wakeups/2 tries

ennru:
Following the discussion in #235 where the stack trace for debugging the issue is putting people off, I implemented @patriknw 's suggestion to not print the stack trace for the first half of `max-wakeups'.


2m:
Fixes #451

PR #465, opened 2018-05-22
Reconnect integration tests

ennru:
Integration tests to try reconnect functionality in producers and consumers


core/src/test/scala/akka/kafka/scaladsl/ReconnectSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+      givenInitializedTopic(topic1)
+
+      val producerSettings = producerDefaults.withBootstrapServers(s"localhost:$proxyPort")
+
+      // start a TCP proxy forwarding to Kafka and start producing messages
+      val (proxyBinding, proxyKillSwtich) = createProxy()
+      Await.ready(proxyBinding, remainingOrDefault)
+
+      val messagesProduced = 100
+      produce(topic1, 1 to messagesProduced, producerSettings)
```
2m: It looks like all of the messages will be sent already before the producer proxy connection is killed. Or am I missing something?

ennru: Yes, you are right. The approach was too naïve.

PR #466, opened 2018-05-23
Mitigate errors in rebalances with partitions in SubSourceLogic

dvallejo:
This fix mitigates the problem about close partitions with pending requests. It is related with this issue #382 

To mitigate this problem I´ve wait a configured time before close the partition. This time can be configured with the property "wait-close-partition".
Also, I´ve added the possibility to re-use partitions. If we are going rebalance from 1 partition to 2, it´s not necessary close the original partition, just open a new one is necessary.

I had to ignore one IT because now a revoke is not required if we can re-use a partition.

I don´t have much experience with akka-streams or reactive kafka so any feedback is welcome.


lightbend-cla-validator:
Hi @dvallejo,

dvallejo:
CLA signed

core/src/main/resources/reference.conf
```diff
@@ -84,6 +84,9 @@ akka.kafka.consumer {
     # Disable auto-commit by default
     enable.auto.commit = false
   }
+
+  # Time to wait for pending requests when a partition is closed
+  wait-close-partition = 5s
```
ennru: 5 seconds as default sounds like a very long time.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +78,45 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
   def partitionAssignedCB(tps: Set[TopicPartition]) = {
     implicit val ec = materializer.executionContext
-    getOffsetsOnAssign.fold(pumpCB.invoke(tps)) { getOffsets =>
-      getOffsets(tps)
+
+    val topicsToBeAssigned = tps -- partitionsToRevoke
+    partitionsToRevoke = partitionsToRevoke -- tps
+
+    getOffsetsOnAssign.fold(pumpCB.invoke(topicsToBeAssigned)) { getOffsets =>
+      getOffsets(topicsToBeAssigned)
```
ennru: Good idea to restrict to the "new" ones.
Wouldn't `topicPartitions` or `partitions` be a better name?

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +78,45 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
+      pendingPartitions --= partitionsToRevoke
+      partitionsInStartup --= partitionsToRevoke
+      partitionsToRevoke.flatMap(subSources.get).foreach(_.shutdown())
+      subSources --= partitionsToRevoke
+    }
+    revokePendingCall = Option(
+      materializer.scheduleOnce(
+        settings.waitClosePartition,
+        new Runnable {
+          override def run(): Unit = cb.invoke(())
```
ennru: `() => cb.invoke()` should be enough.

core/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -332,7 +332,7 @@ class IntegrationSpec extends SpecBase(kafkaPort = KafkaPorts.IntegrationSpec) {
       }
     }
 
-    "call the onRevoked hook" in {
+    "call the onRevoked hook" ignore {
```
ennru: Why is this ignored?

dvallejo: This test was ignored because of the new policy about re-use partitions. I've re-activated the test again. Now, we have to wait for the default time to ensure that the old partition is closed and a new one will be created.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
dvallejo: Totally agree. Changed to `partitions`

core/src/main/resources/reference.conf
dvallejo: Totally agree. Changed to 500 ms

ennru:
Sorry, I misguided you - Scala 2.11 requires the Runnable.

dvallejo:
No problem. I´ve reverted to Runnable

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -291,6 +292,7 @@ class ConsumerSettings[K, V](
     val closeTimeout: FiniteDuration,
     val commitTimeout: FiniteDuration,
     val wakeupTimeout: FiniteDuration,
+    val waitClosePartition: FiniteDuration,
```
ennru: Please move the new parameter last.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +78,45 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
+              .map(_ => pumpCB.invoke(partitions))
               .recover {
-                case _: AskTimeoutException => stageFailCB.invoke(new ConsumerFailed(s"Consumer failed during seek for partitions: $tps."))
+                case _: AskTimeoutException => stageFailCB.invoke(new ConsumerFailed(s"Consumer failed during seek for partitions: $partitions."))
               }
         }
     }
   }
 
+  var partitionsToRevoke: Set[TopicPartition] = Set.empty
```
ennru: Move these vars up to the other mutable fields, please.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +78,45 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
-      getOffsets(tps)
+
+    val partitions = tps -- partitionsToRevoke
+    partitionsToRevoke = partitionsToRevoke -- tps
+
+    getOffsetsOnAssign.fold(pumpCB.invoke(partitions)) { getOffsets =>
+      getOffsets(partitions)
         .onComplete {
-          case Failure(ex) => stageFailCB.invoke(new ConsumerFailed(s"Failed to fetch offset for partitions: $tps.", ex))
+          case Failure(ex) => stageFailCB.invoke(new ConsumerFailed(s"Failed to fetch offset for partitions: $partitions.", ex))
```
ennru: Use `${partitions.mkString(", ")}`

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +78,45 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
           case Success(offsets) =>
             consumer.ask(KafkaConsumerActor.Internal.Seek(offsets))
-              .map(_ => pumpCB.invoke(tps))
+              .map(_ => pumpCB.invoke(partitions))
               .recover {
-                case _: AskTimeoutException => stageFailCB.invoke(new ConsumerFailed(s"Consumer failed during seek for partitions: $tps."))
+                case _: AskTimeoutException => stageFailCB.invoke(new ConsumerFailed(s"Consumer failed during seek for partitions: $partitions."))
```
ennru: As above.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +78,45 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
               }
         }
     }
   }
 
+  var partitionsToRevoke: Set[TopicPartition] = Set.empty
+  var revokePendingCall: Option[Cancellable] = None
```
ennru: It is more like `pendingRevokeCall`.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +78,45 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
+      pendingPartitions --= partitionsToRevoke
+      partitionsInStartup --= partitionsToRevoke
+      partitionsToRevoke.flatMap(subSources.get).foreach(_.shutdown())
+      subSources --= partitionsToRevoke
+    }
+    revokePendingCall = Option(
+      materializer.scheduleOnce(
+        settings.waitClosePartition,
+        new Runnable {
+          override def run(): Unit = cb.invoke(())
```
ennru: You could change `revokePendingCall` to `None` after invoking.

dvallejo:
Totally agree. I've added some log messages and your requests. If you think there is not enough log messages, feel free to request more. Thanks for the review!

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +80,52 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
   private implicit val askTimeout = Timeout(10000, TimeUnit.MILLISECONDS)
   def partitionAssignedCB(tps: Set[TopicPartition]) = {
     implicit val ec = materializer.executionContext
-    getOffsetsOnAssign.fold(pumpCB.invoke(tps)) { getOffsets =>
-      getOffsets(tps)
+
+    val partitions = tps -- partitionsToRevoke
+
+    if (partitions.nonEmpty) {
+      log.debug(s"Assigning new partitions: ${partitions.mkString(", ")}")
```
ennru: For logging you should use the `log.debug("message {}", value)` notation as it defers string creation.
In the particular case you might just add `log.isDebugEnabled` to the if clause.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -78,28 +80,52 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
+    partitionsToRevoke ++= tps
+    val cb = getAsyncCallback[Unit] { _ =>
+      onRevoke(partitionsToRevoke)
+      pendingPartitions --= partitionsToRevoke
+      partitionsInStartup --= partitionsToRevoke
+      partitionsToRevoke.flatMap(subSources.get).foreach(_.shutdown())
+      subSources --= partitionsToRevoke
+    }
+
+    log.debug(s"Waiting ${settings.waitClosePartition.toMillis} ms for pending requests before close partitions")
```
ennru: You might just use `settings.waitClosePartition` which will print its unit.

ennru:
Thank you for your contribution. Keep them coming!

edrevo:
yay! many thanks @dvallejo for this PR! @ennru, do you know when 0.21 will be released? 😄

PR #478, opened 2018-05-28
Introduce consumer control combining control and stream complete

### LINKED ISSUE #476: Create controlled stop and shutdown helper
As described in the new "controlled shutdown" section in the Consumer documentation (not online yet) the recommended way shut down a Consumer-Producer-stream with committing is to use Control.stop, wait for the stream to finish and than call Control.shutdown.
A helper class to be used with mapMaterializedValue could wrap this logic.

ennru:
Fixes #476

### DIFF
--- a/core/src/main/scala/akka/kafka/javadsl/Consumer.scala
+++ b/core/src/main/scala/akka/kafka/javadsl/Consumer.scala
@@ -19,8 +19,9 @@ import org.apache.kafka.clients.consumer.ConsumerRecord
 import org.apache.kafka.common.{Metric, MetricName, TopicPartition}
 
 import scala.collection.JavaConverters._
-import scala.compat.java8.FutureConverters
+import scala.compat.java8.FutureConverters._
 import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success}
 
 /**
  * Akka Stream connector for subscribing to Kafka topics.
@@ -52,10 +53,18 @@ object Consumer {
      * reach the end of the stream.
      * Failures in stream completion will be propagated, the source will be shut down anyway.
      */
-    def drainAndShutdown[T](streamComplete: CompletionStage[T]): CompletionStage[T] =
-      stop()
-        .thenCompose(_ => streamComplete)
-        .thenCompose(result => shutdown().thenApply(_ => result))
+    def drainAndShutdown[T](streamComplete: CompletionStage[T]): CompletionStage[T] = {
+      // TODO this should be possible to implement with CompletionStage instead
+      implicit val ec = ExecutionContexts.sameThreadExecutionContext
+      stop().toScala
+        .flatMap(_ => streamComplete.toScala)
+        .transformWith {
+          case Success(result) =>
+            shutdown().toScala.map(_ => result)
+          case Failure(e) =>
+            shutdown().toScala.transformWith(_ => streamComplete.toScala)
+        }.toJava
+    }
 
     /**
      * Shutdown status. The `CompletionStage` will be completed when the stage has been shut down
@@ -182,7 +191,7 @@ object Consumer {
       .plainPartitionedManualOffsetSource(
         settings,
         subscription,
-        (tps: Set[TopicPartition]) => FutureConverters.toScala(getOffsetsOnAssign(tps.asJava)).map(_.asScala.toMap)(ExecutionContexts.sameThreadExecutionContext),
+        (tps: Set[TopicPartition]) => getOffsetsOnAssign(tps.asJava).toScala.map(_.asScala.toMap)(ExecutionContexts.sameThreadExecutionContext),
         _ => ()
       )
       .map {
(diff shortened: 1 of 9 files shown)


core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -42,6 +42,18 @@ object Consumer {
...
 
+    /**
+     * Stop producing messages from the `Source`, wait for stream completion
+     * and shut down the consumer `Source` so that all consumed messages
+     * reach the end of the stream.
+     */
+    def drainAndShutdown[S](streamCompletion: Future[S])(implicit ec: ExecutionContext): Future[S] =
+      for {
+        _ <- stop()
+        result <- streamCompletion
```
ennru: This might need `recover` handling to shut down properly in case of failure.

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -55,6 +67,43 @@ object Consumer {
...
+    override def metrics: Future[Map[MetricName, Metric]] = control.metrics
+  }
+
+  object DrainingControl {
+    /**
+     * Combine control and a stream completion signal materialized values into
+     * one, so that the stream can be stopped in a controlled way without losing
+     * commits.
+     */
+    def create[T](tuple: (Control, Future[T])) = new DrainingControl[T](tuple)
```
2m: `apply` is nicer in scaladsl.

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -55,6 +67,43 @@ object Consumer {
     def metrics: Future[Map[MetricName, Metric]]
   }
 
+  /**
+   * Combine control and a stream completion signal materialized values into
+   * one, so that the stream can be stopped in a controlled way without losing
+   * commits.
+   */
+  final class DrainingControl[T] private (tuple: (Control, Future[T])) extends Control {
```
2m: We could have the private constructor accept two parameters. And the deconstruction from a tuple to two parameters then can be moved to the factory methods.

docs/src/test/java/sample/javadsl/ConsumerExample.java
```diff
@@ -456,14 +456,10 @@ public void demo() {
                 (batch, elem) -> batch.updated(elem))
             .mapAsync(3, c -> c.commitJavadsl())
             .toMat(Sink.ignore(), Keep.both())
+                .mapMaterializedValue(Consumer.DrainingControl::new)
```
2m: There is `create` factory method in the companion object.

svroonland:
Nice!

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -42,6 +44,19 @@ object Consumer {
...
+     * and shut down the consumer `Source` so that all consumed messages
+     * reach the end of the stream.
+     * Failures in stream completion will be propagated, the source will be shut down anyway.
+     */
+    def drainAndShutdown[S](streamCompletion: Future[S])(implicit ec: ExecutionContext): Future[S] =
+      stop()
+        .flatMap(_ => streamCompletion)
+        .transformWith { _ =>
+          shutdown().flatMap(_ => streamCompletion)
+        }
```
2m: If the `stop()` future fails, then that error is going to be dropped. What about:

```
.transformWith { t =>
  shutdown().flatMap(_ => Future.fromTry(t))
}
```

core/src/main/scala/akka/kafka/javadsl/Consumer.scala
```diff
@@ -46,6 +47,42 @@ object Consumer {
...
 
+    /**
+     * Stop producing messages from the `Source`, wait for stream completion
+     * and shut down the consumer `Source` so that all consumed messages
+     * reach the end of the stream.
+     * Failures in stream completion will be propagated, the source will be shut down anyway.
+     */
+    def drainAndShutdown[T](streamCompletion: CompletionStage[T]): CompletionStage[T] = {
+      // TODO this should be possible to implement with CompletionStage instead
+      // this might become easier when Scala 2.11 is not a requirement anymore
```
2m: Copy this comment to the scala implementation as well.

2m:
I played around with the code and tried to have only the scala implementation as the the only class that implements javadsl.Control is the `WrappedConsumerControl` which delegates everything to the scala implementation.

It almost worked, but the requirement of ExecutionContext then leaks to the javadsl and is not particularly nice.

Anyways, even if the javadsl impl is a mouthfull, I think it is the best we can do right now.

2m:
Instead of exposing Scala's ExecutionContext, I gave another try where javadsl.Control exposes `java.util.concurrent.Executor` which then is wrapped to ExecutionContext and then the implementation of drainAndShutdown can be reused.

Take a look here https://github.com/2m/reactive-kafka/commit/9c663e649a22a2782d9ed94a39a2b96a58f45cec

Issue #474, opened 2018-05-25
Failed: ReconnectSpec


2m:
https://travis-ci.org/akka/reactive-kafka/builds/383630942

2m:
```
[info] - must continue to work when there is another Kafka port available *** FAILED *** (10 seconds, 197 milliseconds)
[info]   java.lang.AssertionError: Expected OnNext(_), yet no element signaled during 10 seconds
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:325)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:316)
[info]   at akka.stream.testkit.TestSubscriber$Probe.requestNext(StreamTestKit.scala:753)
[info]   at akka.kafka.scaladsl.ReconnectSpec.$anonfun$new$14(ReconnectSpec.scala:168)
[info]   at akka.kafka.test.Utils$.assertAllStagesStopped(Utils.scala:31)
```

https://travis-ci.org/akka/reactive-kafka/jobs/384851107

Failed while validating #481

2m:
```
[info] - must continue to work when there is another Kafka port available *** FAILED *** (10 seconds, 621 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (10 seconds) during expectMsg while waiting for OnNext(1)
[info]   at scala.Predef$.assert(Predef.scala:170)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:402)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:379)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:896)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe$$anonfun$expectNextN$1.apply(StreamTestKit.scala:387)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe$$anonfun$expectNextN$1.apply(StreamTestKit.scala:387)
[info]   at scala.collection.Iterator$class.foreach(Iterator.scala:891)
```

https://travis-ci.org/akka/reactive-kafka/jobs/385186868

2m:
https://travis-ci.org/akka/reactive-kafka/jobs/385182545
https://travis-ci.org/akka/reactive-kafka/jobs/385182544

ennru:
We haven't seen any problems with this since improving the tests, closing.

PR #482, opened 2018-05-29
Use delayed log message construction

ennru:


2m:
I wonder if it would be possible to create a scalafix linter that would warn us against such usage. I humbly try to summon @olafurpg to see if he can give us some advice.

2m:
Validation failure was #474

PR #481, opened 2018-05-28
Transactions benchmarks

seglo:
Benchmarks for reactive-kafka transactional support.  I also included a `docker-compose.yml` file with up-to-date community Apache Kafka docker images that can be used with the tests.

PR #483, opened 2018-05-29
Minor update to benchmarks.  Comments and fixture param re-order.

seglo:
Just a little house cleaning from #481


2m:
Validation failure was #474

PR #491, opened 2018-05-31
Expose metadata retrieving functions of Consumer in KafkaConsumerActor

ennru:
Rebased #430 as we moved the KafkaConsumerActor implementation into internal.


ennru:
cc @dsitte-h

dsitte-h:
Thank you for taking care of this. (Sorry, I was swamped.)

docs/src/main/paradox/consumer-metadata.md
```diff
@@ -0,0 +1,30 @@
+# Consumer Metadata
+
+To access the Kafka consumer metadata you need to create the `KafkaConsumerActor` as described in the @ref[Consumer documentation](consumer.md#sharing-the-kafkaconsumer-instance) and send messages from `KafkaConsumerActor.Metadata` to it.
```
2m: The path to the metadata classes is not correct anymore.

docs/src/main/paradox/consumer-metadata.md
```diff
@@ -0,0 +1,30 @@
...
+
+The metadata the Kafka Consumer provides is documented in the @javadoc[Kafka Consumer API](org.apache.kafka.clients.consumer.KafkaConsumer).
+
+The supported metadata are
+
+| Request | Reply | 
+|---------|-------|
+| ListTopics | Topics |
```
2m: Maybe change to `GetTopics` as all of the other messages start with `Get`

docs/src/test/java/sample/javadsl/FetchMetadata.java
```diff
@@ -0,0 +1,43 @@
...
+// #metadata
+        import akka.actor.ActorRef;
+        import akka.kafka.KafkaConsumerActor;
+        import akka.kafka.Metadata;
+        import akka.pattern.PatternsCS;
+        import akka.util.Timeout;
+        import org.apache.kafka.common.TopicPartition;
+
+        import java.util.concurrent.CompletionStage;
+        import java.util.concurrent.TimeUnit;
```
2m: Ooh, what happened here to the formatting?

core/src/main/scala/akka/kafka/Metadata.scala
```diff
@@ -0,0 +1,149 @@
...
+ */
+object Metadata {
+
+  sealed trait Request
+  sealed trait Response
+
+  /**
+   * [[org.apache.kafka.clients.consumer.KafkaConsumer#listTopics()]]
+   */
+  case object ListTopics extends Request with NoSerializationVerificationNeeded
```
2m: This will require a getter for the Java API: `getInstance()`

core/src/main/scala/akka/kafka/Metadata.scala
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Java API:
+   * [[org.apache.kafka.clients.consumer.KafkaConsumer#listTopics()]]
+   */
+  val listTopics: ListTopics.type = ListTopics
+
+  /**
+   * [[org.apache.kafka.clients.consumer.KafkaConsumer#partitionsFor()]]
+   */
+  final case class GetPartitionsFor(topic: String) extends Request with NoSerializationVerificationNeeded
```
2m: Add `create(...)` methods for the Java API for this request message and others below.

docs/src/main/paradox/consumer-metadata.md
dsitte-h: Note that the names were based on the KafkaConsumer methods that are actually exposed here:
https://github.com/apache/kafka/blob/837f31dd1850b179918f83338b4b4487486b2c58/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1780

(In fact the `Get` prefix was prepended to the original methods that did not have a verb in the name to make more sense as names for actor messages.)

2m: I see. Then current names are fine.

docs/src/test/java/sample/javadsl/FetchMetadata.java
ennru: This indentation makes the code snippet look nicer.

core/src/main/scala/akka/kafka/Metadata.scala
ennru: I'll call it createListTopics in line with the case class constructing things.

Issue #415, opened 2018-03-06
introduce PassThrough-only producer messages for conditional processing purposes


svezfaz:
An approach to solve the [conditional message processing](https://doc.akka.io/docs/akka-stream-kafka/current/atleastonce.html#conditional-message-processing) issue could be to introduce the concept of "PassThrough-only" message. Basically the `ProducerMessage` could be available in 2 flavours, one that carry a `ProducerRecord`, and one that doesn't.

Something like

```
sealed trait Message

final case class RecordMessage[K, V, +PassThrough](
    record: ProducerRecord[K, V],
    passThrough: PassThrough
) extends Message

final case class PassThroughMessage[+PassThrough](
    passThrough: PassThrough
) extends Message
```

`ProducerStage` could actually call `producer.send` in case of `RecordMessage`, and only pass them through in case of, well, `PassThroughMessage`.

This would allow conditional processing logic like the following

```
source
  .filter(predicate)
  .map(toProducerMessage)
  .via(Producer.flow)
  .to(commitSink)
```

to be changed to 

```
source
  .map {
    case (k, v, o) if predicate(x) => RecordMessage(k, v, o)
    case (_, _, o)                 => PassThroughMessage(o)
  }
  .via(Producer.flow)
  .to(commitSink)
```

where the latter doesn't skip committing filtered out elements.

DWvanGeest:
Yes! I would like this as well.

My use case is that the source is actually another Kafka topic. Some messages can't be processed and re-published, and so are filtered out of the stream so as not to enter my `Producer.flow`. Since I commit _after_ the `Producer.flow`, those filtered messages are never committed. Instead, nothing is committed until a processable/produceable message comes through the stream.

ennru:
Yes, I like this idea.
@svezfaz Would you like to propose a PR?

schrepfler:
Might be good to have also message headers as well.

svezfaz:
@ennru I'd be happy to work on a PR. How acceptable is it to break the scaladsl/javadsl flow contracts? Not sure this could be pulled off otherwise.

ennru:
Not breaking APIs is better, of course. Give it a stab and let's see how it can be done.
For the new transactional support it does not matter, it hasn't been released, yet.

ennru:
I started to work on this idea. We'll release a new version soon and I try to get this into the new transaction supporting flows. This is a chance to get it in without breaking any APIs - as it is new.

svezfaz:
[...]

PR #495, opened 2018-06-01
Add flexible Message types - flexiFlow

ennru:
This PR improves scenarios where 
* a consumed message should not produce a message to another topic, but still be committed (see [excluding messages](https://doc.akka.io/docs/akka-stream-kafka/current/atleastonce.html#excluding-messages))
* a consumed message should [produce multiple messages](https://doc.akka.io/docs/akka-stream-kafka/current/atleastonce.html#multiple-effects-per-commit)

This solution uses a super trait `Messages` and three implementations `Message`, `MultiMessage` and `PassThroughMessage` (as partly suggested in #415).

It moves the factory methods for transaction support into one object.


ennru:
The documentation needs to be updated with these new message types.
A few more tests with the new message types would be great.

ennru:
@seglo Here is a chunk of work to look at...

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -121,32 +121,79 @@ private[kafka] object ProducerStage {
...
+              }
+
+              if (awaitingConfirmation.decrementAndGet() == 0 && inIsClosed)
+                checkForCompletionCB.invoke(())
             }
+          })
+          val future = r.future.asInstanceOf[Future[OUT]]
+          push(stage.out, future)
+
+        case multiMsg: MultiMessage[K, V, P] =>
```
seglo: It seems like there's an opportunity here to generalize these two cases to DRY things up.  A `Messages` that is not a `PassThroughMessage` represents 1..m messages to produce.  Maybe `Messages` could have an internal method to get a collection of messages which can be mapped to a produce method.  It could also contain a factory method to generate a `Results`.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
```diff
@@ -46,67 +48,97 @@ object Producer {
...
+   *
+   * - [[akka.kafka.ProducerMessage.PassThroughMessage PassThroughMessage]] does not publish anything, and continues in the stream as [[akka.kafka.ProducerMessage.PassThroughResult PassThroughResult]]
+   *
+   * The messages support the possibility to pass through arbitrary data, which can for example be a [[ConsumerMessage.CommittableOffset CommittableOffset]]
+   * or [[ConsumerMessage.CommittableOffsetBatch CommittableOffsetBatch]] that can
    * be committed later in the flow.
    */
-  def flow[K, V, PassThrough](settings: ProducerSettings[K, V]): Flow[Message[K, V, PassThrough], Result[K, V, PassThrough], NotUsed] =
-    scaladsl.Producer.flow(settings)
+  def flow2[K, V, PassThrough](settings: ProducerSettings[K, V]): Flow[Messages[K, V, PassThrough], Results[K, V, PassThrough], NotUsed] =
```
seglo: Would it make sense to change `Message` into a class with a secondary constructor (one with a `ProducerRecord` param and one without) to keep the API code compatible and not require a second flow implementation to the new `Messages` types?

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -0,0 +1,70 @@
...
+import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
+import akka.{Done, NotUsed}
+import org.apache.kafka.clients.producer.ProducerConfig
+
+import scala.concurrent.Future
+
+/**
+ * Akka Stream connector to support transactions between Kafka topics.
+ */
+object Transactional {
```
seglo: :+1:

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
ennru: Yes, but I think the different mechanics around firing off many produce calls an gathering the results are so much more complicated than the single message ones, that it is OK to have it kind of duplicated.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
ennru: The incompatibility is worse on the outgoing messages, as access to the pass through was available via the message field, which does not exist in the `PassThroughMessage`.
A better name than `flow2` has to be suggested, though.

2m: We could name this `plainFlow` which would fit the sink and source factories in this object that have the same prefix. We should also deprecate the original `flow`.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
2m: I agree that DRYing it up completely would bring in additional not needed complexity. However the `Callback` to the `producer.send` is almost completely (1 line difference) identical. That could be factored out.

core/src/main/scala/akka/kafka/ProducerMessage.scala
```diff
@@ -27,18 +33,62 @@ object ProducerMessage {
...
       record: ProducerRecord[K, V],
       passThrough: PassThrough
-  )
+  ) extends Messages[K, V, PassThrough] {
+
+    /**
+     * Returns this Message if applying the predicate to
+     * this Message returns true. Otherwise, returns a [[PassThroughMessage]].
+     */
+    def filter(p: Message[K, V, PassThrough] => Boolean): Messages[K, V, PassThrough] =
```
2m: This exposes scala function to the Java API as well. However it might be not bad as it is a SAM and can be used from java code like:
```java
.map(msg -> msg.filter(m -> m.record().key().equals("key")))
```

core/src/main/scala/akka/kafka/javadsl/Producer.scala
ennru: The `plain` refers to unwrapped Kafka types, I believe.
And yes, I think the existing `flow` should become deprecated.

core/src/main/scala/akka/kafka/internal/ProducerStage.scala
seglo: Yes, I think `producer.send` is the best opportunity too.  I'm definitely in favour in keeping things simple and readable.

core/src/main/scala/akka/kafka/ProducerMessage.scala
```diff
@@ -6,12 +6,18 @@
 package akka.kafka
 
 import org.apache.kafka.clients.producer.{ProducerRecord, RecordMetadata}
+import scala.collection.immutable.Seq
```
patriknw: I think it's better to `import scala.collection.immutable` and then use `immutable.Seq`, to avoid accidental loss of import which can easily happen when merging conflicting changes and such

core/src/main/scala/akka/kafka/javadsl/Producer.scala
seglo: To keep the convention of "_something_ Source|Sink|Flow" perhaps the new base type name could be used for `Producer.messagesFlow`.

Another idea, pass through already has meaning in this project, but what about just `through`.  `Producer.through`?

core/src/main/scala/akka/kafka/ProducerMessage.scala
seglo: Question: We still cross build for Scala 2.11.  Even though the 2.11 artifact is built with the 1.8 JVM can it still be used by Java 1.7 users?

ennru: No, the class files are in JDK 8 format.

patriknw: I think you can say that Java users should use `_2.12` artifacts. For Akka we tend to still support `_2.11` for Java, but there shouldn't be any strong reasons why one would use that.

ennru:
I've now re-worked the documentation, Scaladoc and the example snippets.
A few things are left:
* Rename `Messages` as it is too close too `Message`?
* Better name than `flow2` that reads nicely in `.via(Producer.flow2(...))`, `messagesFlow` is my favourite right now (depending on the choice above)
* Add `@deprecated` to `flow` methods
* Make the producer callback more DRY?

ennru:
Make `filter` more Java friendly while being Scala 2.11 friendly:
```
[error] /home/travis/build/akka/reactive-kafka/docs/src/test/java/sample/javadsl/ProducerExample.java:165:1: incompatible types: scala.Function1 is not a functional interface
[error]     multiple non-overriding abstract methods found in interface scala.Function1
[error]                 message.filter(m -> m.record().value().toString().length() > 1);
```

seglo:
This is a great UX improvement for the library.

PR #496, opened 2018-06-05
Rebalance test in a kafka cluster

2m:
This required a little modification to the kafka docker images so it is possible to create kafka clusters with the correct configuration for inside and outside listeners.


core/src/it/scala/akka/kafka/RebalanceSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+  implicit val pc = PatienceConfig(10.seconds, 100.millis)
+  implicit val mat = ActorMaterializer()
+
+  override def afterAll() = {
+    TestKit.shutdownActorSystem(system)
+  }
+
+  "alpakka kafka" should {
+
+    "not loose any mesasges during rebalance" in {
```
ennru: s/loose/lose/
s/mesasges/messages/

PR #511, opened 2018-06-18
Add scaladoc generation to travis build

2m:


2m:
It does not. Only the `docs/Local/paradox` does.

PR #513, opened 2018-06-19
Kafka 1.1.0

sullis:


ennru:
Can you explain the need for upgrading? 
How does it affect the Alpakka Kafka connector?
You may always override the dependency in your project if you see the need for later versions.

ennru:
Now that Kafka 2.0 is out, I'll close this.
See #544

PR #517, opened 2018-06-21
fix partition reassignment on partitioned sources

edrevo:
This PR fixes two subtle bugs that break partition re-assignment on the partitioned sources.

The bug shows itself like this:

1. A consumer is assigned partitions 1,2,3 by Kafka. Everything works fine
2. Kafka rebalances and the consumer is now assigned only partition 1. Everything still works fine.
3. Kafka rebalances again and the consumer is now assigned partitions 1,2,3 again. However, partitions 2 and 3 will remain paused indefinitely and will start accumulating lag.

This happens because the `partitionsToRevoke` set isn't empied when the revoke happens, which means that partitions that had been revoked will still be marked for revoke.

Another fix introduced in this PR is the fact that the `pendingRevokeCall` was set to `None` before the revoke code was run (since it is an async callback). This could lead to cases where the `waitClosePartition` wasn't honored when specific timing conditions happened.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
@@ -123,6 +123,8 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
       partitionsInStartup --= partitionsToRevoke
       partitionsToRevoke.flatMap(subSources.get).foreach(_.shutdown())
       subSources --= partitionsToRevoke
+      partitionsToRevoke = Set.empty
+      pendingRevokeCall = None
     }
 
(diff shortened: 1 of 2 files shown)


edrevo:
I have added a test case that fails without the fix and passes after the fix. The test is quite slow since there are several `Thread.sleep` statements. I am happy to change those sleep statements by something more advanced if there's an alternative way of achieving the same test.

edrevo:
@enru, this is related to https://github.com/akka/alpakka-kafka/pull/466, which is a PR that you reviewed. Just FYI.

ennru:
Thank you, great to see this fixed. Too bad I didn't notice these bugs during review.

PR #520, opened 2018-06-24
add 'oraclejdk10' to the Travis build matrix

sullis:


ennru:
Thank you for this suggestion @sullis.
For now, I do not think it adds a lot of value though. We could possibly change to use JDK10 instead of JDK9.

sullis:
We might consider removing JDK 9 from the build matrix.  

Oracle EOL document: http://www.oracle.com/technetwork/java/javase/eol-135779.html

ennru:
We'll work on supporting Scala 2.13 and later JDK versions during this fall.

PR #524, opened 2018-06-26
Improve wakeup logging

ennru:
* Delay logging to `maxWakeups / 2`
* Don't log stack trace (but be clear about it happening during poll)
* Log on error level if initial poll fails more than `maxWakeups / 2`, printing bootstrap servers
* Avoid allocation of `FiniteDuration` after committing

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -5,7 +5,6 @@
 
 package akka.kafka.internal
 
-import java.io.{PrintWriter, StringWriter}
 import java.util
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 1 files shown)


core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -321,8 +321,8 @@ class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V]) extends Actor w
   def poll(): Unit = {
     val wakeupTask = context.system.scheduler.scheduleOnce(settings.wakeupTimeout) {
       log.warning(
-        "KafkaConsumer poll has exceeded wake up timeout ({}ms). Waking up consumer to avoid thread starvation.",
-        settings.wakeupTimeout.toMillis
+        "KafkaConsumer poll has exceeded wake up timeout ({}). Waking up consumer to avoid thread starvation.",
+        settings.wakeupTimeout
```
2m: You could call `toCoarsest` on `wakeupTimeout` here and below so it is printed as seconds or minutes instead of milliseconds.

PR #525, opened 2018-06-26
Added passThroughFlow documentation example

davideicardi:
I have tried to add a documentation section on how to implement pass through as suggested by @ennru some day ago on gitter.

The general idea is to use a `commitableSource` then using a custom graph split the `msg` from the `commitableOffset`. Process the `msg` and merge again it with the offset. Then commit the offset.

The solution seems to work in my scenario, but of course a deep review from some expert is required (and appreciated :smile:  ).

You can try the graph using scala fiddle: https://scalafiddle.io/sf/pneGum2/1

Another point where I have some concerns is `mapAsync(3)(_.commitScaladsl())`. Doing commits in parallel is correct? My concern is that an old offset will overwrite a newer one.

thanks

### DIFF
--- a/docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
+++ b/docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
@@ -4,23 +4,17 @@
  */
 package sample.scaladsl
 
-import akka.kafka.ConsumerMessage.{CommittableMessage, CommittableOffsetBatch}
+import akka.kafka.ConsumerMessage.{CommittableMessage, CommittableOffset, CommittableOffsetBatch}
 import akka.kafka._
 import akka.actor.{Actor, ActorLogging, ActorRef, ActorSystem, Props}
 import akka.kafka.scaladsl.{Consumer, Producer}
-import akka.stream.scaladsl.{Flow, Keep, RestartSource, Sink, Source}
-import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
-import akka.stream.ActorMaterializer
+import akka.stream.scaladsl._
+import akka.stream.{ActorMaterializer, FlowShape, Graph}
 import akka.{Done, NotUsed}
 import org.apache.kafka.clients.consumer.{ConsumerConfig, ConsumerRecord}
 import org.apache.kafka.clients.producer.ProducerRecord
 import org.apache.kafka.common.{Metric, MetricName, TopicPartition}
-import org.apache.kafka.common.serialization.{
-  ByteArrayDeserializer,
-  ByteArraySerializer,
-  StringDeserializer,
-  StringSerializer
-}
+import org.apache.kafka.common.serialization.{ByteArrayDeserializer, ByteArraySerializer, StringDeserializer, StringSerializer}
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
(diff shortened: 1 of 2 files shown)


docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
```diff
@@ -191,6 +185,57 @@ object AtLeastOnceWithBatchCommitExample extends ConsumerExample {
...
+    // #passThroughFlow
+    var writeFlow = Flow[ConsumerRecord[String, Array[Byte]]].map(_ => ???)
+
+    val control =
+      Consumer
+        .committableSource(consumerSettings, Subscriptions.topics("topic1"))
+        .map(msg => (msg.record, msg.committableOffset))
+        .via(PassThroughFlow(
+          writeFlow, // Here you can replace a module
+          Flow[CommittableOffset].map(x => x) // just pass through the offset
```
ennru: You could just short-circuit the identity flow, or do you anticipate a good use-case for it?

docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
```diff
@@ -191,6 +185,57 @@ object AtLeastOnceWithBatchCommitExample extends ConsumerExample {
...
+object PassThroughFlowCommitExample extends ConsumerExample {
+  def main(args: Array[String]): Unit = {
+
+    // #passThroughFlow
+    var writeFlow = Flow[ConsumerRecord[String, Array[Byte]]].map(_ => ???)
+
+    val control =
+      Consumer
+        .committableSource(consumerSettings, Subscriptions.topics("topic1"))
+        .map(msg => (msg.record, msg.committableOffset))
```
ennru: Alternatively, the user could pass a selector function which extracts the part used for the inner flow from the generic in-data type.

docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
davideicardi: Do you mean changing
`unzip.out1 ~> passThroughB ~> zip.in1` to `unzip.out1 ~> zip.in1`? In this case yes, I think that this is the most common scenario and makes the Graph more "semantic". (just tried [here](https://scalafiddle.io/sf/pneGum2/2)).

docs/src/test/scala/sample/scaladsl/ConsumerExample.scala
davideicardi: Sorry, here I don't understand what do you mean ... In any case feel free to just modify it as you think it more correct, or move in another part of the doc.

ennru: You've seen the function in the HBase connector, converting the incoming data to mutations. Here the user could provide a function to select the data to be passed to the flow from the incoming message. 
The Graph DSL would `broadcast`, apply the selector function, and pass the result to the flow.
On the outgoing side, you would keep the `zip`.

ennru:
@davideicardi I'd really like you to open this PR to the Alpakka main project instead. Maybe as part of the Patterns section there https://developer.lightbend.com/docs/alpakka/current/patterns.html

ennru:
Thank you for moving it to the main Alpakka repo.

PR #527, opened 2018-06-28
WIP: Make example code run as tests

ennru:
This changes the code used in documentation snippets to be run in test cases to validate it does actually work. 
For now, it uses a copy of `SpecBase` which is used in the `core` module, but the samples should possibly move to that module instead.


docs/src/main/paradox/producer.md
```diff
@@ -32,7 +32,7 @@ See @javadoc[KafkaProducer API](org.apache.kafka.clients.producer.KafkaProducer)
 ## Producer as a Sink
 
 `Producer.plainSink` 
-(@scala[@scaladoc[Producer API](akka.kafka.scaladsl.Producer$)]@java[@scaladoc[Producer API](akka.kafka.javadsl.Producer$)]) 
+(@scala[@scaladoc[Producer API](akka.kafka.scaladsl.Producer$)]@java[@scaladoc[Producer API](akka.kafka.docs.javadsl.Producer$)])
```
2m: The addition of `docs` pacakge here is probably accidental.

PR #529, opened 2018-06-29
Improve integration rebalance spec

2m:
* use all kafka hosts for boostrapping
* create topic with replication factor
* wait until the cluster is fully formed

### DIFF
--- a/core/src/it/resources/reference.conf
+++ b/core/src/it/resources/reference.conf
@@ -1,0 +1,18 @@
+akka {
+  loglevel = "DEBUG"
+
+  actor {
+    serialize-messages = on
+    debug {
+      lifecycle = off
+      receive = off
+    }
+  }
+
+  test {
+    single-expect-default = 10s
+  }
+  kafka.consumer {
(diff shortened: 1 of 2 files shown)


core/src/it/resources/reference.conf
```diff
@@ -0,0 +1,18 @@
+akka {
+  loglevel = "DEBUG"
+
+  actor {
+    serialize-messages = on
```
ennru: Why?

core/src/it/resources/reference.conf
```diff
@@ -0,0 +1,18 @@
...
+      lifecycle = off
+      receive = off
+    }
+  }
+
+  test {
+    single-expect-default = 10s
+  }
+  kafka.consumer {
+    max-wakeups = 10
```
ennru: That's the default...

core/src/it/resources/reference.conf
2m: Copy-pasta from the reference.conf from tests.

PR #531, opened 2018-07-03
Wrap MetricsControl access to consumer in future

### LINKED ISSUE #528: Race condition when getting metrics from consumer
Hi everybody! I tried to use metrics from the consumer like this (simplified):
```scala
val control = runnableGraph.run()
control.metrics.onComplete { 
  case Success(metrics) =>
    println(metrics)  
  case Failure(e) =>
    println(e)
}
```
And I got an error:
```
java.lang.IllegalArgumentException: Unsupported recipient ActorRef type, question not sent to [null]. Sender[null] sent the message of type "akka.kafka.internal.KafkaConsumerActor$Internal$RequestMetrics$"
```

I tried to debug it and I found, that this error is from this code inside `SingleSourceLogic`:
```scala
override def metrics: Future[Map[MetricName, Metric]] = {
  import akka.pattern.ask
  import scala.concurrent.duration._
  consumer
    .?(RequestMetrics)(Timeout(1.minute))
    .mapTo[ConsumerMetrics]
    .map(_.metrics)(ExecutionContexts.sameThreadExecutionContext)
}
```
When I called `metrics` method, `consumer` variable is still uninitialized. That's because it initialized inside `preStart` block, which could be not finished at the moment when I call `metrics` method. Typical race condition.

ennru:
Fixes #528

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
@@ -91,7 +91,7 @@ private[kafka] object ConsumerStage {
     new KafkaSourceStage[K, V, CommittableMessage[K, V]] {
       override protected def logic(shape: SourceShape[CommittableMessage[K, V]]) =
         new ExternalSingleSourceLogic[K, V, CommittableMessage[K, V]](shape, consumer, subscription)
-        with CommittableMessageBuilder[K, V] with MetricsControl {
+        with CommittableMessageBuilder[K, V] {
           override def groupId: String = _groupId
           lazy val committer: Committer = {
             val ec = materializer.executionContext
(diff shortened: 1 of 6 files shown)


2m:
Needs a rebase.

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -341,15 +341,18 @@ private[kafka] trait PromiseControl extends GraphStageLogic with Control {
...
-      .?(RequestMetrics)(Timeout(1.minute))
-      .mapTo[ConsumerMetrics]
-      .map(_.metrics)(ExecutionContexts.sameThreadExecutionContext)
+    consumerFuture
+      .flatMap { consumer =>
+        consumer
+          .ask(RequestMetrics)(Timeout(1.minute))
+          .mapTo[ConsumerMetrics]
+          .map(_.metrics)(ExecutionContexts.sameThreadExecutionContext)
+      }(executionContext)
```
s4nk: @ennru I'm calling `metrics` after materializing a stream and I'm getting an exception like this:
```
java.lang.IllegalStateException: not yet initialized: only setHandler is allowed in GraphStageLogic constructor. To access materializer use Source/Flow/Sink.setup factory
	at akka.stream.stage.GraphStageLogic.interpreter(GraphStage.scala:393)
	at akka.stream.stage.GraphStageLogic.materializer(GraphStage.scala:403)
	at akka.kafka.internal.BaseSingleSourceLogic.executionContext(BaseSingleSourceLogic.scala:34)
	at akka.kafka.internal.MetricsControl.metrics(ControlImplementations.scala:88)
	at akka.kafka.internal.MetricsControl.metrics$(ControlImplementations.scala:78)
	at akka.kafka.internal.BaseSingleSourceLogic.metrics(BaseSingleSourceLogic.scala:26)
```
The flow takes messages from a `KafkaConsumerActor` and sends them to another actor, the whole thing looks like this:
```
      val sink: Sink[Output, NotUsed] =
        ActorSink.actorRefWithAck[Output, SinkActorCommandType, Ack](...)

      val control: Consumer.Control =
        Consumer.plainExternalSource[String, Array[Byte]](kafkaConsumerActor, subscription)
          .map { consumerRecord =>
            processRecord(consumerRecord)
          }
          .to(sink)
          .run()

      control.metrics
```
Did the race condition just move to a different place now?

s4nk: OK, looks like somebody already raised the issue, #847

PR #533, opened 2018-07-04
Fix missing implicit

2m:
The default embedded kafka config was picked up because of a missing implicit config in the `EmbeddedKafkaLike`.


ennru:
Just in time for me.

PR #535, opened 2018-07-05
Configure logging for IntegrationTest

ennru:
and better logging pattern


2m:
This will introduce conflicts to some PRs, but not a big problem to fix those.

PR #536, opened 2018-07-05
Refactor benchmarks to separate specs

2m:


2m:
Oops. Some println's sneaked through. Will fix those.

ennru:
Ah, I believed they were helpful in understanding the output.

Issue #549, opened 2018-08-21
Consumer fetches duplicate records when lagging, resulting in high network use, but low throughput


adkafka:
We noticed that when running our Kafka Consuming application with lag, there was a large discrepancy between the bytes the Kafka Brokers were sending us, and the number of messages we were processing. 

Here is an example, where we forced lag onto our pod, which resulted in a large spike in Network out on our Kafka Brokers:
![canonicalbytesoutbrokers](https://user-images.githubusercontent.com/398153/44424328-a125d780-a53d-11e8-8faf-c9028dbd27e3.png)

Most of the traffic was still making it to our application (a kubernetes pod):
![podnetworkusage](https://user-images.githubusercontent.com/398153/44424317-9bc88d00-a53d-11e8-9183-c28b00d99aa5.png)

But yet significantly fewer messages were making it into our application:
![messagesconsumedrate](https://user-images.githubusercontent.com/398153/44424342-b438a780-a53d-11e8-9151-0209b6168770.png)

We calculated an average message size for our application to be between 225 and 300 bytes / message. Cumulative bytes out on brokers was about 130MB/s, which comes out to between 430,000 and 570,000 messages per second. Our application was only processing about 7,000 messages per second, or about 60-80x less messages than expected!

To dig in further, we turned on "DEBUG" logging for our alpakka-kafka client, and sampled our logs at two different states: steady-state (no-lag), and lagging.

We noticed that when our application is lagging, it will produce **repetitive fetch requests** for the **same topic-partition**, that result in **responses of size `max.partition.fetch.bytes`**. 

```
akafka@SC-C02TC0JMHF1T:Desktop [eng:default] $ cat lagging.log | grep "canonical_L0_prd-0 at offset"
2018-08-21 16:35:11,676 DEBUG Fetcher - [Consumer clientId=consumer-1, groupId=canonical_L0_prd.to.influx_energy_eng_v6] Added READ_UNCOMMITTED fetch request for partition canonical_L0_prd-0 at offset 93793935 to node kaf01-stream-eng.sn.tesla.services:9093 (id: 1 rack: null)
2018-08-21 16:35:12,631 DEBUG Fetcher - [Consumer clientId=consumer-1, groupId=canonical_L0_prd.to.influx_energy_eng_v6] Added READ_UNCOMMITTED fetch request for partition canonical_L0_prd-0 at offset 93793935 to node kaf01-stream-eng.sn.tesla.services:9093 (id: 1 rack: null)
[...]

PR #551, opened 2018-08-23
Test case illustrating dropping of fetched data

ennru:
I used this test case to investigate #549 it does not assert anything useful but shows the reported misbehaviour.


ennru:
Hopefully, https://github.com/apache/kafka/pull/5844 will resolve this misbehaviour in Kafka and we can ignore this going forward.

ennru:
A new attempt to improve the Kafka client's behaviour is being made by @seglo in https://github.com/apache/kafka/pull/6988.

ennru:
Replaced by #837

Issue #561, opened 2018-09-14
Commit With Metadata


johnclara:
I would like to add metadata based on the ConsumerRecord with my committed offset.

My goal is to consume a replicated topic in a DR DC starting each consumer from around the time where it left off. My goal is to add the timestamp of the message in ConsumerOffsets and replicate this topic to a DR DC. From this timestamp, I can get an offset close to that of the oldest last read message per GroupTopic in the DR DC during a failover. A ConsumerGroup would read each replicated topic starting from these offsets.

johnclara:
PR is here: https://github.com/akka/alpakka-kafka/pull/562

ennru:
Fixed with #563

PR #563, opened 2018-09-17
Able to add metadata to the committed offset

johnclara:
I would like to add metadata based on the ConsumerRecord along with each commit. The issue for this is here: #561

This PR will allow users to pass in a function to the source constructor which will add metadata to each commit based on the consumer record. Each committed offset is guaranteed to contain metadata constructed from the message as long as the commitRefreshInterval is infinite.

If the commitRefreshInterval is not infinite, then when the Consumer is assigned a partition it will also receive an offset without a message. If it has not processed a new message before the end of the interval, it will recommit this offset for this partition without adding metadata.
In order to solve this, we can read the last committed message each time a partition is assigned in order to reconstruct the metadata. This could negatively impact performance so it is not implemented in this PR.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
@@ -20,8 +20,8 @@ import akka.stream.scaladsl.Source
 import akka.stream.stage.{GraphStageLogic, GraphStageWithMaterializedValue}
 import akka.util.Timeout
 import akka.{Done, NotUsed}
-import org.apache.kafka.clients.consumer.{ConsumerConfig, ConsumerRecord}
-import org.apache.kafka.common.requests.IsolationLevel
+import org.apache.kafka.clients.consumer.{ConsumerConfig, ConsumerRecord, OffsetAndMetadata}
+import org.apache.kafka.common.requests.{IsolationLevel, OffsetFetchResponse}
 import org.apache.kafka.common.{Metric, MetricName, TopicPartition}
 
 import scala.collection.JavaConverters._
@@ -49,7 +49,6 @@ private[kafka] object ConsumerStage {
       override protected def logic(shape: SourceShape[(TopicPartition, Source[CommittableMessage[K, V], NotUsed])]) =
         new SubSourceLogic[K, V, CommittableMessage[K, V]](shape, settings, subscription)
         with CommittableMessageBuilder[K, V] with MetricsControl {
-
           override def groupId: String = settings.properties(ConsumerConfig.GROUP_ID_CONFIG)
           lazy val committer: Committer = {
             val ec = materializer.executionContext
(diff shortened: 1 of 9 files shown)


core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -161,6 +161,18 @@ object Consumer {
                               subscription: Subscription): Source[CommittableMessage[K, V], Control] =
     Source.fromGraph(ConsumerStage.committableSource[K, V](settings, subscription))
 
+  /**
+   * The `committableSourceWithMetadata` makes it possible to add additional metadata (in the form of a string)
+   * when an offset is committed based on the record. This can be useful (for example) to store information about which
+   * node made the commit, what time the commit was made, the timestamp of the record etc.
+   */
+  def committableSourceWithMetadata[K, V](
```
ennru: Would there be a better name for this construct?
eg. `commitWithMetadataSource`?

core/src/main/scala/akka/kafka/internal/ConsumerStage.scala
```diff
@@ -84,6 +83,22 @@ private[kafka] object ConsumerStage {
...
     }
 
+  def committableSourceWithMetadata[K, V](settings: ConsumerSettings[K, V],
+                                          subscription: Subscription,
+                                          _metadataFromRecord: ConsumerRecord[K, V] => String) =
+    new KafkaSourceStage[K, V, CommittableMessage[K, V]] {
+      override protected def logic(shape: SourceShape[CommittableMessage[K, V]]) =
+        new SingleSourceLogic[K, V, CommittableMessage[K, V]](shape, settings, subscription)
+        with CommittableMessageWithMetadataBuilder[K, V] {
+          override def metadataFromRecord(record: ConsumerRecord[K, V]): String = _metadataFromRecord(record)
```
ennru: I believe this implementation may replace the existing `committableSource` version, if that one returns the empty String from `metadataFromRecord`.
Wouldn't that lead to fewer moving parts?

core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -161,6 +161,18 @@ object Consumer {
                               subscription: Subscription): Source[CommittableMessage[K, V], Control] =
     Source.fromGraph(ConsumerStage.committableSource[K, V](settings, subscription))
 
+  /**
+   * The `commitWithMetadataSource` makes it possible to add additional metadata (in the form of a string)
+   * when an offset is committed based on the record. This can be useful (for example) to store information about which
+   * node made the commit, what time the commit was made, the timestamp of the record etc.
+   */
+  def commitWithMetadataSource[K, V](
```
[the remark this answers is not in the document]
ennru: That would break backwards-compatibility, wouldn't it?

johnclara:
Great! I'll start on updating the documentation

johnclara:
Okay! I've updated the documentation and have the unit tests passing. I also added a javadsl

ennru:
Thank you for this great addition, please help us to improve it even more.

PR #571, opened 2018-09-25
Use named classes for consumer stages and set a name attribute

### LINKED ISSUE #569: Create classes instead of anonymous instances
I believe it would help to understand the Alpakka Kafka connector if it would be changed to use named classes instead of anonymous. Log output and Akka Stream tree visualisations would benefit from more exact names when debugging things.

ennru:
This makes the consumer stage implementations non-anonymous classes so that logging and debugging information becomes more useful.
Furthermore, do all consumer stages get the `Name` attribute set which even contains a subscription description. For that description to become more compact, the subscription implementations got specific `toString` methods.

Another useful refactoring could be to make the classes top-level instead of within the `ConsumerStage` object, but I refrained from doing altogether in one PR.

Fixes #569

### DIFF
--- a/core/src/main/scala/akka/kafka/Subscriptions.scala
+++ b/core/src/main/scala/akka/kafka/Subscriptions.scala
@@ -18,6 +18,14 @@ sealed trait Subscription {
 
   /** Configure this actor ref to receive [[akka.kafka.ConsumerRebalanceEvent]] signals */
   def withRebalanceListener(ref: ActorRef): Subscription
+
+  def renderStageAttribute: String
+
+  protected def renderListener: String =
+    rebalanceListener match {
+      case Some(ref) => s" rebalanceListener $ref"
+      case None => ""
+    }
 }
 sealed trait ManualSubscription extends Subscription {
   def withRebalanceListener(ref: ActorRef): ManualSubscription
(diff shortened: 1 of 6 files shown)


core/src/main/scala/akka/kafka/Subscriptions.scala
```diff
@@ -56,6 +72,15 @@ object Subscriptions {
       extends ManualSubscription {
     def withRebalanceListener(ref: ActorRef): Assignment =
       Assignment(tps, Some(ref))
+
+    override def toString: String = {
+      val listener = rebalanceListener match {
+        case Some(ref) => s",rebalanceListener=$ref"
+        case None => ""
+      }
+      s"Assignment(${tps.mkString(",")}$listener)"
```
2m: Parenthesis here and in other sunbscription definitions will be converted to the URL-encoded-format when used in the `name` attribute. Should we use something more URL friendly, like `/`?

tests/src/test/scala/akka/kafka/internal/SubscriptionsSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+    "be readable for assignments with offset" in {
+      encode(Subscriptions.assignmentWithOffset(Map(new TopicPartition("topic1", 1) -> 123L))) should be(
+        "topic1-1+offset123"
+      )
+    }
+
+    "be readable for assignments with timestamp" in {
+      encode(Subscriptions.assignmentOffsetsForTimes(Map(new TopicPartition("topic1", 1) -> 12345L))) should be(
+        "topic1-1+timestamp12345"
+      )
```
2m: A testcase that shows multiple partitions/topics would be also good to see.

2m:
Looks good. Now only MiMa needs some filters.

Issue #555, opened 2018-08-27
Integration Tests fail on connecting to Kafka in Docker


ennru:
The new integration test making use of sbt-docker-compose fail as they don't connect to Kafka within the Docker containers when running on Travis.
https://travis-ci.org/akka/alpakka-kafka/jobs/421072038

I've tried to understand the problem, in #554

ennru:
Closed with #573

PR #573, opened 2018-09-26
Fix integration tests

2m:
I have accidentally reverted to an older Kafka docker image in #536. We need at least `1.1.0` because we rely on subsitutions in your `docker-compose.yml` file:

https://github.com/akka/alpakka-kafka/blob/f7f227121ecb88595f2ec5f6a024f9ff276b5cd7/docker-compose.yml#L15-L17

This PR changes Kafka docker image version back to `1.1.0`.

What is more, #554 added support for `TimeoutException` while waiting for a desired cluster state. However that exception was still not handled in `waitUntilConsumerGroup`. This PR fixes that by unifying periodical check logic under `periodicalCheck`.


2m:
Fixes #555

2m:
For some reason Travis CI did not kick in. Closing and reopening.

PR #578, opened 2018-09-28
Resolve hostname from gateway ip

2m:
Might help workaround unresolvable "linuxkit" hostname on MacOS.

@ennru this one is for you to test by running `tests/dockerComposeTest it:test` from sbt console.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     ports:
       - "9094"
     environment:
-      HOSTNAME_COMMAND: "docker info | grep ^Name: | cut -d' ' -f 2"
+      HOSTNAME_COMMAND: "route -n | awk '/UG[ \t]/{print $$2}'"
       PORT_COMMAND: "docker port `hostname` 9094 | cut -d: -f 2"
       KAFKA_ADVERTISED_LISTENERS: INSIDE://:9092,OUTSIDE://_{HOSTNAME_COMMAND}:_{PORT_COMMAND}
       KAFKA_LISTENERS: INSIDE://:9092,OUTSIDE://:9094


ennru:
The whole thing does not work for me, but the hostname retrieval output looks ok.

```
[pr_578] ~/dev/alpakka-kafka> docker exec 4b038fd86302 route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.48.1    0.0.0.0         UG    0      0        0 eth0
192.168.48.0    0.0.0.0         255.255.240.0   U     0      0        0 eth0
[pr_578] ~/dev/alpakka-kafka> docker exec 4b038fd86302 route -n | awk '/UG[ \t]/{print $2}'
192.168.48.1
```

2m:
Alright. Can you do the following then:

Having this branch checked out run:

    docker-compose up --scale kafka=3

Then locate the line

    advertised.listeners = INSIDE://:9092,OUTSIDE://<ip>:<port>

On any of the kafka hosts and run `kafkacat` to get the metainfo of the topic:

    kafkacat -b <ip>:<port> -L -t test-topic

After first run the topic will be auto-created so it should report that no leader is available. But on the second run you should be seeing something like this:

```
~ ─╼ kafkacat -b 172.19.0.1:32858 -L -t test-topic                                               0
Metadata for test-topic (from broker 1001: 172.19.0.1:32858/1001):
 3 brokers:
  broker 1001 at 172.19.0.1:32858
  broker 1003 at 172.19.0.1:32860
  broker 1002 at 172.19.0.1:32859
 1 topics:
  topic "test-topic" with 1 partitions:
    partition 0, leader 1002, replicas: 1002, isrs: 1002
```

2m:
This is not relevant anymore. We fixed the current setup on MacOS by adding `linuxkit0000000...` hostname to the `/etc/hosts` and resolving it to `localhost`.

ennru:
It looks like OS X changed the hostname to `docker-desktop` instead of `linuxkit000...`.

PR #579, opened 2018-09-28
Be able to commit metadata for partitioned sources

johnclara:
We would like to be able to commit metadata with the offset from a partitioned source


ennru:
Thank you for continuing your work on this!

PR #583, opened 2018-10-02
Run docker integration tests sequentially

2m:
Something is still not quite right here. First test starts alone and finishes. But the the second seems to start too soon.


2m:
Made it work using commands. There is some peculiarity when it comes to `concurrentRestrictions` and nested tasks which I could not quite get right.

Issue #424, opened 2018-04-05
API naming improvements


patriknw:
Some naming in the API is not nice, such as `commitJavadsl` and `commitScaladsl` and now I noticed new things added with similar naming convention.

ktoso:
Yeah; I followed existing style in the new https://github.com/akka/reactive-kafka/pull/423
Not quite sure what we should do about it.

On one hand, it would be best to have complete javadsl/scaladsl split as we have in most Akka APIs, but that would be breaking APIs.

If naming change, then the same should be applied to commitJavadsl I etc perhaps?
Open for ideas, was pondering it a bit and did not figure out a nice solution without breaking people's code so far

ktoso:
Thankfully we were able to avoid adding another method with "javadsl" in the name.
Ticket remains valid

ennru:
A non-API-breaking way to hide the `commitScaladsl`/`commitJavadsl` methods is to introduce a flow eg. `Committer` that does it behind the scenes.
It could offer pre-packaged `groupedWithin` and other useful patterns.

patriknw:
I like that idea, including the added convenience.

ennru:
With where we are today, users should not interact with the `commitJavadsl` and `commitScaladsl` methods. We could even deprecate them, I believe.

seglo:
> We could even deprecate them, I believe.

Yes, they only ever offered the illusion of control anyway.  Let's deprecate them. :+1:

ennru:
With the deprecation of the old methods in #959 we can close this.

chetankm-cs:
@seglo 

> Yes, they only ever offered the illusion of control anyway. Let's deprecate them. 👍

Can you explain this a little bit, at my work we were relying on these apis to handle commit failures as restarting the committer flow can cause commits to be missed.

seglo:
Hi @chetankm-cs 

> Can you explain this a little bit, at my work we were relying on these apis to handle commit failures 

That's interesting. How do you handle commit failures when they're observed?

> as restarting the committer flow can cause commits to be missed.

Were you using [`RestartFlow.onFailuresWithBackoff`](https://doc.akka.io/docs/akka/current/stream/operators/RestartFlow/onFailuresWithBackoff.html)? I don't think I've seen this use before with the committer flow, but it makes sense that it could cause commits, from the docs:

> The restart process is inherently lossy

[...]

PR #584, opened 2018-10-02
Add support for merging commit batches

rtimush:
I often find it useful to do some grouping on the consumed messages, that is not related to the commit process. However, in order to be able to build an offsets batch at the end of the flow I currently have to keep all `CommittableOffset`s.

 It would be useful if I could already aggregate them into a `CommittableOffsetBatch`, and then aggregate even further to keep commit rate low.

An alternative implementation would be to change `updated` signature to accept `Commitable`, but the comment says that it might be generalized and moved to `akka.stream`, and I'm not sure if merging `CommittableOffsetBatch` with an arbitrary `Commitable` should be supported.

### DIFF
--- a/core/src/main/mima-filters/1.0-M1.backwards.excludes
+++ b/core/src/main/mima-filters/1.0-M1.backwards.excludes
@@ -1,2 +1,3 @@
 # new method in sealed trait
 ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.Subscription.renderStageAttribute")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.updated")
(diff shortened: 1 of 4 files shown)


ennru:
MiMa reports correctly that this is not fully backwards-compatible.
```
abstract method updated(akka.kafka.ConsumerMessage#CommittableOffsetBatch)akka.kafka.ConsumerMessage#CommittableOffsetBatch in interface akka.kafka.ConsumerMessage#CommittableOffsetBatch is present only in current version
```
Please add `ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.updated")` to `1.0-M1.backwards.excludes`.

ennru:
Would you like to do a stab on [my idea regarding the commit API](https://github.com/akka/alpakka-kafka/issues/424#issuecomment-424328601), as well?

rtimush:
Thanks, I have added the MiMa exclusion. As for the "committer" flow — I already have it in all projects that use alpakka-kafka, I will try to extract it and create a PR.

ennru:
Thank you for this! And we look forward to your battle-proved committer flows...

PR #587, opened 2018-10-03
More tests for partitioned sources and a few tracing improvements

ennru:


core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -115,17 +116,23 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
...
         }
     }
   }
 
   val partitionRevokedCB = getAsyncCallback[Set[TopicPartition]] { tps =>
+    // TODO this called in startup with empty tps, some existing tests reply on the callback
     pendingRevokeCall.map(_.cancel())
     partitionsToRevoke ++= tps
     val cb = getAsyncCallback[Unit] { _ =>
+      if (log.isDebugEnabled) {
```
2m: I thought that `log.debug` takes arguments after format by name, thus not calling `toString` on them if `isDebugEnabled` is false. But I checked the source code and that is not the case. :/

2m:
Rebased on top of master and made it wait until the desired consumer state.

PR #589, opened 2018-10-03
Seek to the last unemitted offset after subsource is cancelled

2m:
This is an attempt to address the situation reported in https://github.com/akka/alpakka-kafka/issues/382#issuecomment-412092233 where a number of messages are skipped when the partition SubSource is closed.

This PR adds last unemitted element propagation from a SubSourceStage back to SubSourceLogic which then seeks the partition to the last unemitted offset.


ennru:
This makes sure any elements already in the stream do not get ignored, as the position in Kafka has been moved forward. 
This does work if the subsource is **completed** downstream, even though the log text and the test case call it "cancelled".

PR #592, opened 2018-10-04
A test case for partition subsource that fails

2m:
An aditional test case for #589

This was originally reported here: https://github.com/akka/alpakka-kafka/issues/382#issuecomment-412092233


ennru:
This is just a test case illustrating that #589 needs to cover even stream failure, right?

2m:
#589 covered stream failure as well. Not sure why the test failed on Travis.

Anyways, now improved consumer stream completion by using `takeWhile` instead of `keepAlive`.

PR #595, opened 2018-10-08
Rebalance listener 

### LINKED ISSUE #539: Rebalance event listeners don't respect ConsumerRebalanceListener contract
Contract:
https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html#onPartitionsAssigned-java.util.Collection-

Quote:
>It is guaranteed that all the processes in a consumer group will execute their `onPartitionsRevoked(Collection)` callback before any instance executes its `onPartitionsAssigned(Collection)` callback.

This guarantees that revoked partitions' records get the change to be taken care of: e.g. we can commit them etc etc.

However, here:
https://github.com/akka/alpakka-kafka/blob/master/core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala#L72
user callbacks are invoked asynchronously via the `notifyUserOnAssign` method which does not wait for the completion of user's handler, which opens a possibility for a "race condition" where consuming from newly assigned partitions will start before callbacks (e.g. assuring offsets commits) are finished.

An idea of a fix would be something along these lines:

```diff
-subscription.rebalanceListener.foreach(ref ⇒ ref ! TopicPartitionsAssigned(subscription, set))
+subscription.rebalanceListener
+  .map(ref ⇒ ref ? TopicPartitionsAssigned(subscription, set)
+  .map(_ => Done))
+  .getOrElse(Future.successful(()))
```

However, I'm not familiar with custom graph stages enough to make this into a functional implementation (since this becomes a `Future`, how do I incorporate that into an async callback?), so if maintainers see this as a valid concern (and a fix approach), I could use some help.

### LINKED ISSUE #596: Partitioned sources do not send rebalance events to listener
Working to fix #539 I discovered that partitioned sources do not send rebalance events at all.

ennru:
* split `ConsumerTest` into `ConsumerMock`, `ConsumerTest`, and `CommittingSpec` 
* tests to ensure rebalance events are sent in order
* send rebalance events before calling async callback
* send rebalance events from partitioned sources (fixes #596)

There is a tiny change in the behaviour of the `onRevoke` call-back in `plainPartitionedManualOffsetSource`: Before this change, it was called with an empty set before first partition assignment to a consumer -- now it is called only when partitions are revoked (the set is non-empty).


core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -32,22 +32,33 @@ private[kafka] abstract class SingleSourceLogic[K, V, Msg](
...
-      }
     }
 
     def rebalanceListener: KafkaConsumerActor.ListenerCallbacks =
-      KafkaConsumerActor.ListenerCallbacks(partitionAssignedCB.invoke, partitionRevokedCB.invoke)
+      KafkaConsumerActor.ListenerCallbacks(
+        assignedTps => {
+          subscription.rebalanceListener.foreach {
+            _.tell(TopicPartitionsAssigned(subscription, assignedTps), sourceActor.ref)
+          }
```
2m: I am not sure if moving `TopicPartitionsAssigned` messages sends from the callback to here will make big of a difference.

`partitionAssignedCB.invoke` is also an actor message send to the stages ActorRef. Therefore messages for both callback calls will come in order and eventually `TopicPartitionsAssigned` and `TopicPartitionsRevoked` will also be sent in order.

Am I missing something?

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -143,9 +146,12 @@ private[kafka] abstract class SubSourceLogic[K, V, Msg](
       }
   }
 
-  val partitionRevokedCB = getAsyncCallback[Set[TopicPartition]] { tps =>
+  val partitionRevokedCB = getAsyncCallback[Set[TopicPartition]] { revoked =>
+    subscription.rebalanceListener.foreach {
+      _.tell(TopicPartitionsRevoked(subscription, revoked), sourceActor.ref)
+    }
```
2m: I think we should keep these consistent: either send those messages from the callback, or from the client thread.

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
ennru: The `KafkaConsumerActor.WrappedAutoPausedListener` is mounted directly in the Kafka consumer and called on a Kafka consumer thread. It in turn calls `ListenerCallbacks.onAssign` and `ListenerCallbacks.onRevoke` directly without any async involved.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
ennru: You're so right, what was I thinking...

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
2m: Yes, but the implementation of `onAssign` and `onRevoke` sends actor messages. And before the change the actor messages were sent to the graph stage (async callback invocation) and after the change the message send is directly to the rebalance listener.

That is one less hop, but I do not see how this changes the semantics.

ennru: Ok, what would be needed to solve #539 are synchronous callbacks.
That would require a new API.

PR #598, opened 2018-10-09
Split producer stage file

ennru:
* Unwrap some classes from being inner classes of the `ProducerStage` object
* Properly mark classes as internal
* Rename `TransactionProducerStage` to `TransactionalProducerStage` to align with `TransactionalSource`


2m:
Failure was https://github.com/akka/alpakka-kafka/issues/601

PR #600, opened 2018-10-09
Run integration tests with oldest Kafka possible

2m:
wurstmeister now [publishes Kafka docker images for many different Kafka versions](https://hub.docker.com/r/wurstmeister/kafka/tags/). All of the images contain the latest wurstmeister scripts. So substitutions now work for every image. More info on their image versioning scheme [here](https://github.com/wurstmeister/kafka-docker#tags-and-releases).

Integration tests timed-out when running with `2.11-0.11.0.3`. They passed with `2.11-1.0.1` so this is what we should be running integration tests against.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - "2181:2181"
   kafka:
     scale: ${KAFKA_SCALE:-1}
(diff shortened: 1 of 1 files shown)


ennru:
We're using the 1.1.0 image right now and will continue with it for now.

Issue #608, opened 2018-10-11
Adapt to the new Kafka 2.0 client behaviour


ennru:
As the Kafka client is now upgraded to 2.0.0, the Alpakka Kafka connector should make use of the new methods that support a timeout instead of the risk of blocking indefinitely.
See [Kafka KIP-266](https://cwiki.apache.org/confluence/display/KAFKA/KIP-266%3A+Fix+consumer+indefinite+blocking+behavior)

Especially the new behaviour in `poll(Duration)` needs to be considered carfully:

> One notable usage is prohibited by the new poll: previously, you could call poll(0) to block for metadata updates, for example to initialize the client, supposedly without fetching records. Note, though, that this behavior is not according to any contract, and there is no guarantee that poll(0) won't return records the first time it's called. Therefore, it has always been unsafe to ignore the response.

zaharidichev:
Yes, so pretty much the new semantics  guarantee that the timeout will be honoured and that there will be no infinite blocking. Seems to me a few things in the actor consumer may be changed to adopt to that behaviour (and simplify things a bit): 

- pollInterval can be close or equal to `max.poll.interval`
- We introduce a configurable timeout for the usual demand driven poll call 
- We get rid of the poll(0) calls and use a normal poll there with a timeout that we can control.
- We get rid of the waking up logic as it seems to me that it has been introduced to prevent from what  might happen in case things go wrong while using the previous version of poll. 

I am sure there are other intricacies, but I have not spend too much time looking at the code. I will have another things. In any way, transitioning to the new poll + incorporating the solution proposed in #559 should simplify the logic as well as solve the problems of throwing away fetched data..  Please, do correct me if I am wrong..

zaharidichev:
@patriknw 
I guess as long as we can control the maximum amount of blocking per call it is a judgement call whether we want a short blocking poll once or do poll(Duration.Zero) the same way poll(0) is done now. Its just that we need to bear in mind that even with poll(0), we are not guaranteed to not be handed results back. So, in case we are, we have to deal with that. 

But the wake up stuff can certainly go away by the looks of it. I can submit a PR with these changes. If you guys have any more recommendations, they are all welcome. What is of even more interest to me is how we deal with the dropped fetches illustrated in #559

zaharidichev:
Should I base myself of of #610 and try to tackle the poll(Duration....), wake up part of this or somebody is already working on it ?

PR #610, opened 2018-10-11
Use some 2.0 features (timeouts)

ennru:
* use a configurable timeout on all Metadata requests
* deprecate `ConsumerSettings` constructors
* use a *for now* hardcoded timeout for `position`
* use a *for now* hardcoded timeout for `offsetForTimes`

This PR does the easy part of #608.


ennru:
Should the two now hard-coded timeouts be made part of `ConsumerSettings`?

2m:
Lets do that. It'll help with finetuning things. And also it does not pollute user API much, as those are set by corresponding builder methods.

PR #614, opened 2018-10-14
New polling and removal of wake up functionality

zaharidichev:
As already discussed in #608, this PR adopts the new Kafka poll introduced in [KIP-266](https://cwiki.apache.org/confluence/display/KAFKA/KIP-266%3A+Fix+consumer+indefinite+blocking+behavior). This enables us to get rid of the manual wake up functionality that was present in the consumer actor


tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -189,7 +189,7 @@ class IntegrationSpec extends SpecBase(kafkaPort = KafkaPorts.IntegrationSpec) w
       }
 
       control2 should not be null
-      sleep(4.seconds)
+      sleep(10.seconds)
```
2m: This test-case is timing dependent. Try increasing this even more to see if the test-case starts to pass when running all of the test-cases in the spec.

Ideally we should refactor the test to not be timing dependent.

zaharidichev: Is there a reason for the initial value of four or it was just  derived by trial and error.

2m: Yup, trial and error. 

The ideal solution would be to not wait for a defined amount of time, but instead look at the arriving elements and complete the consumer stream when some condition is met. Take a look here for inspiration:

https://github.com/akka/alpakka-kafka/blob/9a08f599bd3476ffcd54b574933a433b8058c2c9/tests/src/test/scala/akka/kafka/scaladsl/PartitionedSourcesSpec.scala#L361

But we can leave this improvement for another PR.

zaharidichev: We can indeed but still these random failures are quite strange. I slashed the number of events by 4 and increased the sleep duration and am still getting failures. I cannot quite explain it to myself why my change has triggered this strange behaviour...

2m: Aha. If decreasing the number of messages and increasing sleep times does not solve this, then maybe indeed we are loosing messages now. I'll dig deeper as well.

zaharidichev: Its strange indeed. I took another look and only changing the poll call to the deprecated once makes all the tests pass. Looking at the semantics of the new poll call, the main difference is the fact that it might block for infinity on fetching metadata... Will have to think a bit more on that one.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -340,72 +339,15 @@ class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V]) extends Actor w
...
-          stacks
-        )
-      }
-      consumer.wakeup()
-    }(context.system.dispatcher)
+
     val currentAssignmentsJava = consumer.assignment()
 
-    def tryPoll(timeout: Long): ConsumerRecords[K, V] =
+    def tryPoll(timeout: FiniteDuration): ConsumerRecords[K, V] =
```
patriknw: use `java.time.Duration` here to avoid conversions all the time

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -431,7 +373,7 @@ class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V]) extends Actor w
         var i = 10
         while (i > 0 && commitsInProgress > 0) {
           LockSupport.parkNanos(10 * 1000)
-          val pollTimeout = if (i == 1) 1L else 0L
+          val pollTimeout = if (i == 1) 1.millis else Duration.Zero
```
patriknw: don't create new duration instance each time

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -340,72 +339,15 @@ class KafkaConsumerActor[K, V](settings: ConsumerSettings[K, V]) extends Actor w
...
-            // have prevented the re-balance callbacks to be called leaving the Source in an inconsistent state w.r.t
-            // assigned TopicPartitions. In order to reconcile the state we manually call callbacks for all added/remove
-            // TopicPartition assignments aligning the Source's state the consumer's.
-            // We are safe to perform the operation here since the poll() thread has been aborted by the wakeup call
-            // and there are no other threads are using the consumer.
-            // Note: in case of manual partition assignment this is not needed since rebalance doesn't take place.
-            if (subscriptions.nonEmpty) {
-              val newAssignments = consumer.assignment().asScala
-              reconcileAssignments(currentAssignmentsJava.asScala.toSet, newAssignments.toSet)
-            }
```
patriknw: wonder how this is handled now, if `poll` returns without any records?

perhaps it's handled by the metadata calls that are part of the poll?

zaharidichev: Looking at the code, that's what the case seems to be

zaharidichev:
@patriknw Albeit looking good in theory, it fails some of the tests quite frequently. So either we have flaky tests that this change has made more obvious or the change itself is having some undesired effect. In any case I will sit down and investigate this over the weekend  and run the benchmarks you suggested as well.

zaharidichev:
retest this please

zaharidichev:
After quite a bit of looking into that over the weekend, I have changed a few things especially around the way how we call poll when we do not have any commits. I have tweaked some of the test to be a bit less time - dependent. However some other tests still seem to fail from time to time and is hard fo me to tell why.  Looking at the build history however, this does not seem like something new. If you have any further suggestions, they are welcome

zaharidichev:
Is there a chance someone takes a looks over that to see whether I am doing something particularily wrong which causes tests to fail... ?

2m:
Sorry about being slow on this. Most of the Lightbend is currently in the Reactive Summit. I'll give more attention to this change.

zaharidichev:
Ah no worries,  its just bugging me what the problem is. I will take another looks as well. But I have the feeling tests failures occur every now and then on master as well , am I right ?

2m:
There were a couple of flaky tests, but we have been working on those recently and situation has improved. On this PR, I see that the benchmark job eventually stalls. Taking a look why that is happening.

2m:
The failing benchmark job can be run from sbt by running `benchmarks/dockerComposeTest it:testOnly *.AlpakkaKafkaAtMostOnceConsumer` This benchmark jobs consumes messages really slowly in this PR (around 100 messages every š seconds). The full benchmark expects 50k messages and it times out.

zaharidichev:
I will look into it in more detail. Thanks for the tip

zaharidichev:
@2m I am getting some exceptions while running benchmarks... Namely `TimeoutException: Timed out waiting for a node assignment.`  Does that look familiar ? Is it something maybe with my docker setup ? 

```

**Comment by 2m** [comment]
Seems like a docker problem. Are you on MacOS? If so, you need to add an
entry to your hosts file that resolves the hostname used in Kafka brokers
to localhost. You can find that hostname in the output of Kafka nodes when
starting up. Look for a line containing  'advertised.listeners'.

**Comment by zaharidichev** [comment]
Hmmm, yes I see the line but that does not look very correct to me, does it ? 

```
kafka_1      | 	advertised.host.name = null
kafka_1      | 	advertised.listeners = INSIDE://:9092,OUTSIDE://linuxkit-025000000001:32768
kafka_1      | 	advertised.port = null
```

**Comment by 2m** [comment]
Discussion continued here: https://gitter.im/akka/alpakka-kafka?at=5bd0bc330386d0635fedf0bf

**Comment by zaharidichev** [comment]
Hmm, interestingly enough the benchmark test hangs for me (after consuming the messages), during the clean-up phase. Just hangs at: 

```
kafka_1      | [2018-10-26 18:26:37,223] INFO [GroupMetadataManager brokerId=1001] Removed 0 expired offsets in 6 milliseconds. (kafka.coordinator.group.GroupMetadataManager)
```

Interesting whether there is something wrong with my environment

**Comment by 2m** [comment]
I think it is still consuming messages, albeit quote slowly. Try printing every incoming message.

**Comment by zaharidichev** [comment]
Yes, that was going very vey slowly event on master. What is a sane rate of msg/s here ?

**Comment by 2m** [comment]
For me master starts out around 20msgs/s but then it speeds up after 1000 messages or so. Eventually completing the test in around a minute.

**Comment by zaharidichev** [comment]
After some more digging, I am quite puzzled. Tests are still failing sporadically. As a matter of fact the build has currently been running for more than 2 hours and is stuck at running the benchmarks. Reading through the KIP that describes the changes introduced with the non blocking poll, it is hard for me to imagine what has gone wrong. I think the kafka implementation might be implicitly relying on the fact that the consumer blocks indefinitely on metadata requests... The moment I swap the consumer.poll call in the KafkaConsumerActor for the old, deprecated poll, I seize to see sporadic failures. Why might that be happening, is very unclear to me. I mean, after all the only difference that there is in the Kafka Consumer code is: 

```
if (includeMetadataInTimeout) {
    final long metadataStart = time.milliseconds();
    if (!updateAssignmentMetadataIfNeeded(remainingTimeAtLeastZero(timeoutMs, elapsedTime))) {
        return ConsumerRecords.empty();
    }
    metadataEnd = time.milliseconds();
    elapsedTime += metadataEnd - metadataStart;
} else {
    while (!updateAssignmentMetadataIfNeeded(Long.MAX_VALUE)) {
        log.warn("Still waiting for metadata");
    }
    metadataEnd = time.milliseconds();
}
```

Before we were hitting the else branch and now we call the `updateAssignmentMetadataIfNeeded` with the poll timeout. How is that affecting things to such a degree, is unclear to me...  Could it be something to do with the commit callbacks and whether they are called in timely manner (since their execution shall be included in the timeout as well from what I take it..). Also it might be useful to get a threaddump of this stuck build and see where things are. Are messages still getting consumed, are we blocked waiting for some state change that will never happen.. Maybe a bit of help here from someone more familiar with the current constraints of the KafkaConsumerActor will be useful.

**Comment by ennru** [comment]
Thank you so much for your efforts with this!
I guess we'll cut a 1.0-M1 release with the old solution and dig even more into why this change behaves so surprisingly.

**Comment by ennru** [comment]
I rebased this on current master. Some tests will fail as failing to reach the broker looks different now (see #636).

File: core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala [hunk 5]
```diff
@@ -360,95 +359,27 @@ import scala.util.control.{NoStackTrace, NonFatal}
...
         // poll(0). Using poll(1) will always block for 1 ms, since there are no messages.
         // Therefore we do 10 poll(0) with short 10 μs delay followed by 1 poll(1).
         // If it's still not completed it will be tried again after the scheduled Poll.
         var i = 10
         while (i > 0 && commitsInProgress > 0) {
           LockSupport.parkNanos(10 * 1000)
-          val pollTimeout = if (i == 1) 1L else 0L
-          checkNoResult(tryPoll(pollTimeout))
+          val pollTimeout = if (i == 1) oneMilli else java.time.Duration.ZERO
+          checkNoResult(consumer.poll(pollTimeout))
```
ennru: We should take measures and try to find out if any `consumer.poll(ZERO)` does anything useful. Maybe `parkNanos(100)` followed by a `consumer.poll(oneMilli)` is just as good.

zaharidichev:
Did anyone get the chance to take a look at why these tests are failing in such a misterious manner ?

ennru:
Yes, @2m has made some progress - specifically that removing the loop with `poll(0)`s made committing significantly slower. (Don't ask me why :-) )

zaharidichev:
@2m  Can I help somehow? Do you have any idea why the looping made committing slower? I had the feeling our __new poll__ times out before the commit callback was invoked and we need multiple calls to poll before tha commit happens (with the new non blocking poll) as opposed to aways completing the commit callback with the old blocking call. Was that the case ?

2m:
From what I see the change to a different poll changed timings a bit and so some of the tests are now failing a bit more. So I am refactoring some tests to be less racy like replacing `Thread.sleep` with future composition or similar.

As to the committing optimization, the loop allows us to not do the poll with a full pollInterval when there are no requests in progress and there are some committing to be done.

From the empiric tests I noticed that on average after 5 such polls commits usually succeed in the benchmark code.

zaharidichev:
I see, this makes sense. These benchmarks (refering to the ones showing that commits succeed after 5 polls on average) have been run using the new non blocking poll ?

ennru:
There are issues in the test cases left around rebalance event notifications.

ennru:
The `PartitionedSourcesSpec` `must rebalance when a new consumer comes up` fails sometimes locally for me. 
I captured this bit of logging which shows the two partitions that get revoked are believed to have been consumed until offset 49, but the stream counts only 10 elements.

```
:24,658 [a.k.internal.PlainSubSource$$anon$3 ]  #5 Assigning new partitions: topic-1-7-3, topic-1-7-2
:24,658 [a.kafka.internal.KafkaConsumerActor ]  received handled message PartitionAssigned(Map(topic-1-7-3 -> 49, topic-1-7-2 -> 49)) from Actor[akka://Spec/deadLetters]
:24,659 [a.k.scaladsl.PartitionedSourcesSpec ]  Inner Sub-source for topic-1-7-2
:24,660 [a.k.internal.SubSourceStage$$anon$1 ]  #5 Starting SubSource for partition topic-1-7-2
:24,660 [a.k.scaladsl.PartitionedSourcesSpec ]  Inner Sub-source for topic-1-7-3
:24,660 [a.kafka.internal.KafkaConsumerActor ]  received handled message RequestMessages(0,Set(topic-1-7-2)) from Actor[akka://Spec/system/StreamSupervisor-0/$$p#1395606315]
:24,661 [a.k.internal.SubSourceStage$$anon$1 ]  #5 Starting SubSource for partition topic-1-7-3
:24,672 [a.kafka.internal.KafkaConsumerActor ]  received handled message PartitionRevoked(Set(topic-1-7-3, topic-1-7-0, topic-1-7-1, topic-1-7-2)) from Actor[akka://Spec/deadLetters]
:24,672 [a.kafka.internal.KafkaConsumerActor ]  received handled message PartitionAssigned(Map(topic-1-7-0 -> 49, topic-1-7-1 -> 49)) from Actor[akka://Spec/deadLetters]
:24,733 [a.kafka.internal.KafkaConsumerActor ]  received handled message Poll(akka.kafka.internal.KafkaConsumerActor@3abc63c,true) from Actor[akka://Spec/deadLetters]
:24,763 [a.kafka.internal.KafkaConsumerActor ]  received handled message RequestMessages(0,Set(topic-1-7-3)) from Actor[akka://Spec/system/StreamSupervisor-0/$$q#-2040830947]
:24,764 [a.kafka.internal.KafkaConsumerActor ]  received handled message Poll(akka.kafka.internal.KafkaConsumerActor@64414b1,true) from Actor[akka://Spec/deadLetters]
:24,864 [a.kafka.internal.KafkaConsumerActor ]  received handled message Poll(akka.kafka.internal.KafkaConsumerActor@64414b1,false) from Actor[akka://Spec/system/kafka-consumer-5#1677781759]
:24,904 [a.kafka.internal.KafkaConsumerActor ]  received handled message Poll(akka.kafka.internal.KafkaConsumerActor@3abc63c,true) from Actor[akka://Spec/deadLetters]
:24,967 [a.kafka.internal.KafkaConsumerActor ]  received handled message Poll(akka.kafka.internal.KafkaConsumerActor@64414b1,true) from Actor[akka://Spec/deadLetters]
:25,004 [a.k.internal.PlainSubSource$$anon$3 ]  #4 Closing SubSources for revoked partitions: topic-1-7-3, topic-1-7-2
:25,005 [a.k.internal.SubSourceStage$$anon$1 ]  #4 Completing SubSource for partition topic-1-7-2
:25,005 [a.k.scaladsl.PartitionedSourcesSpec ]  Sub-source for topic-1-7-3 completed: Received [10] messages in total.
:25,005 [a.k.scaladsl.PartitionedSourcesSpec ]  Sub-source for topic-1-7-2 completed: Received [10] messages in total.
:25,005 [a.k.internal.SubSourceStage$$anon$1 ]  #4 Completing SubSource for partition topic-1-7-3

```

ennru:
My last commit introduced asking and logging of partition offsets from Kafka - that test works now...
And the other fails on JDK 11, only. Kafka 2.1 brings JDK 11 support...

core/src/main/resources/reference.conf
```diff
@@ -61,10 +61,12 @@ akka.kafka.consumer {
   # The KafkaConsumerActor will throw
   # `org.apache.kafka.common.errors.WakeupException` which will be ignored
   # until `max-wakeups` limit gets exceeded.
+  # DEPRECATED: No op as waking up consumer is not used anymore
```
ennru: Shouldn't these settings be removed altogether? Or left there commented out to let people understand that they may remove them from their code?
And add `deprecated` to the `ConsumerSettings` `withXyz` methods.

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -32,7 +32,7 @@ import scala.concurrent.{Future, Promise}
   final val actorNumber = KafkaConsumerActor.Internal.nextNumber()
   final def consumerFuture: Future[ActorRef] = consumerPromise.future
 
-  private val partitionLogLevel = if (settings.wakeupDebug) Logging.InfoLevel else Logging.DebugLevel
+  private val partitionLogLevel = Logging.DebugLevel
```
ennru: Remove this?

core/src/main/scala/akka/kafka/javadsl/Consumer.scala
```diff
@@ -73,7 +73,7 @@ object Consumer {
    * one, so that the stream can be stopped in a controlled way without losing
    * commits.
    */
-  final class DrainingControl[T] private[javadsl] (control: Control, streamCompletion: CompletionStage[T])
+  final class DrainingControl[T] private[javadsl] (control: Control, val streamCompletion: CompletionStage[T])
```
ennru: Tiny feature slip. I introduced this to get a handle to the stream completion without calling `drainAndShutdown`.

tests/src/test/scala/akka/kafka/internal/PartitionedSourceSpec.scala
```diff
@@ -566,7 +566,10 @@ object PartitionedSourceSpec {
     override def assign(partitions: java.util.Collection[TopicPartition]): Unit =
       tps.set(partitions.asScala.map(_ -> Assigned).toMap)
 
-    override def poll(timeout: Long): ConsumerRecords[K, V] = {
+//    override def poll(timeout: Long): ConsumerRecords[K, V] = {
+//      tps = partitions.asScala.map(_ -> Assigned).toMap
```
ennru: Remove this.

2m:
The last CI failure was https://github.com/akka/alpakka-kafka/issues/683

Issue #542, opened 2018-08-05
Support/documentation for actor initialization problem handling


PAStheLoD:
Hello!

Using 0.22, and after browsing the issue tracker I think this is a request for enhancement. Currently it's either not possible (or at least "very" undocumented to figure out how) to catch initialization exceptions.

I'm thinking of these:
```
akka.actor.ActorInitializationException: akka://application/system/kafka-consumer-1: exception during creation
...
Caused by: org.apache.kafka.common.KafkaException: Failed to construct kafka consumer
...
Caused by: org.apache.kafka.common.config.ConfigException: No resolvable bootstrap urls given in bootstrap.servers
```

Even if I manually initialize the actor, since the exception happens in Akka, there's no way to handle it.

I see that the documentation [changed](https://github.com/akka/alpakka-kafka/commit/281a91eed81836975e9288c656d7d890b48f8cc1), and it used to recommend using a supervisor actor. The current recommendation of using watchTermination doesn't seem to work for termination due to initialization errors.

2m:
When creating a consumer stream with the wrong settings for the consumer actor:

```scala
val result: Future[Done] = Consumer
   .plainSource(..., ...)
   .runForeach(println)
```

The `result` Future will be failed with a `akka.kafka.ConsumerFailed` exception which you can handle and recover.

Granted, not every failure sets the cause of the `ConsumerFailed` exception. We should improve that.

ennru:
Fixed with #617

PR #617, opened 2018-10-18
Expose consumer creation failures

### LINKED ISSUE #612: Improve handling of misconfigured Kafka client
The `KafkaConsumerActor` can do better in situations where the `ConsumerSettings` are not complete:
1. don't have bootstrap servers (fails in preStart)
1. don't have a group id (fails during poll)

ennru:
Catch exceptions in the actor and send them as `Status.Failure` to known stage actors so that the stream can fail with the appropriate error.

* Pass creating stage's ActorRef to consumer actor (owner)
* Catch Kafka Consumer creation exceptions and pass to owner
* Catch exceptions in poll's outer catch clause

Fixes #612

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
@@ -5,8 +5,7 @@
 
 package akka.kafka.internal
 
-import akka.actor.Status.Failure
-import akka.actor.{ActorRef, Terminated}
+import akka.actor.{ActorRef, Status, Terminated}
 import akka.kafka.Subscriptions.{Assignment, AssignmentOffsetsForTimes, AssignmentWithOffset}
 import akka.kafka.{ConsumerFailed, ManualSubscription}
 import akka.stream.SourceShape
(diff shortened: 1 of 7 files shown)


ennru:
Fixes #542

core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
```diff
@@ -52,11 +49,12 @@ private[kafka] abstract class BaseSingleSourceLogic[K, V, Msg](
...
         }
         pump()
-      case (_, Failure(e)) =>
+      case (_, Status.Failure(e)) =>
         failStage(e)
       case (_, Terminated(ref)) if ref == consumerActor =>
         failStage(new ConsumerFailed())
     }
+    consumerActor = createConsumerActor()
     sourceActor.watch(consumerActor)
```
2m: `watch` returns the same `ActorRef` that it gets. So it is possible to rewrite this with:

```scala
consumerActor = sourceActor.watch(createConsumerActor())
```

ennru: No, it is `Unit` typed.

2m: Ahh, `sourceActor` is a stage sctor. Well that is unfortunate that the API differs.

PR #618, opened 2018-10-19
Add factory methods for ProducerMessages

ennru:
* Add `single`, `multi` and `passThrough` methods for constructing `Envelope` sub-classes
* The factory methods allow for better type-inference
* A few cases still need explicit types

### DIFF
--- a/tests/src/test/java/docs/javadsl/ConsumerExample.java
+++ b/tests/src/test/java/docs/javadsl/ConsumerExample.java
@@ -279,7 +279,7 @@ class ConsumerToProducerSinkExample extends ConsumerExample {
         Consumer.committableSource(consumerSettings, Subscriptions.topics("topic1", "topic2"))
             .map(
                 msg ->
-                    new ProducerMessage.Message<String, byte[], ConsumerMessage.Committable>(
+                    ProducerMessage.<String, byte[], ConsumerMessage.Committable>single(
                         new ProducerRecord<>(
                             "targetTopic", msg.record().key(), msg.record().value()),
                         msg.committableOffset()))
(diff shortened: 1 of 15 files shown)


tests/src/test/java/docs/javadsl/ConsumerExample.java
```diff
@@ -273,14 +273,11 @@ public void demo() {
...
-                          new ProducerRecord<>("topic2", msg.record().value()),
-                          msg.committableOffset() // the passThrough
-                          );
-                  return prodMsg;
-                })
+                msg ->
+                    ProducerMessage.single(
+                        new ProducerRecord<>("topic2", msg.record().key(), msg.record().value()),
+                        msg.committableOffset() // the passThrough
+                        ))
```
2m: Very nice simplification!

tests/src/test/scala/docs/scaladsl/SerializationSpec.scala
```diff
@@ -221,6 +221,53 @@ class SerializationSpec
...
+        .run()
+
+    sleep(30.seconds, "to see if it fails")
+
+    control1.isShutdown.futureValue should be(Done)
+    control2.isShutdown.futureValue should be(Done)
+    result1.failed.futureValue shouldBe a[org.apache.kafka.common.errors.SerializationException]
+    result2.failed.futureValue shouldBe a[org.apache.kafka.common.errors.SerializationException]
+  }
+   */
```
2m: We should remove this or make it work.

ennru: Removed.

Issue #619, opened 2018-10-27
Introduce committing flow


ennru:
A non-API-breaking way to hide the commitScaladsl/commitJavadsl methods is to introduce a flow eg. Committer that does it behind the scenes.
It could offer pre-packaged groupedWithin and other useful patterns.

ennru:
@rtimush Did you look into contributing your solution for this already?

rtimush:
@ennru Yes, I started working on it, but then had a couple of busy weeks and didn't have time to finish. I will try to push something this week.

PR #622, opened 2018-10-30
Committer sink

rtimush:
Design considerations:
 * committer is a `Sink` as in my experience it is usually the last step of the flow
 * committer settings `max-batch` and `max-interval` define the amount of work we can afford to lose, in terms of messages or time

Implements #619

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -103,6 +103,21 @@ akka.kafka.consumer {
 }
 # // #consumer-settings
 
+# // #committer-settings
+# Properties for akka.kafka.CommitterSettings can be
+# defined in this section or a configuration section with
+# the same layout.
+akka.kafka.committer {
+
+  # Maximum number of messages in a single commit batch
+  max-batch = 1000
+
+  # Maximum interval between commits
+  max-interval = 1m
+
+}
+# // #committer-settings
+
 # The dispatcher that will be used by default by consumer and
 # producer stages.
 akka.kafka.default-dispatcher {
--- a/core/src/main/scala/akka/kafka/CommitterSettings.scala
+++ b/core/src/main/scala/akka/kafka/CommitterSettings.scala
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 - 2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.kafka
+import java.util.concurrent.TimeUnit
+
+import akka.actor.ActorSystem
+import akka.util.JavaDurationConverters._
+import com.typesafe.config.Config
+
+import scala.concurrent.duration._
+
+object CommitterSettings {
+
+  /**
+   * Create settings from the default configuration
+   * `akka.kafka.committer`.
+   */
+  def apply(actorSystem: ActorSystem): CommitterSettings =
+    apply(actorSystem.settings.config.getConfig("akka.kafka.committer"))
+
+  /**
+   * Create settings from a configuration with the same layout as
+   * the default configuration `akka.kafka.committer`.
+   */
+  def apply(config: Config): CommitterSettings = {
+    val maxBatch = config.getLong("max-batch")
+    val maxInterval = config.getDuration("max-interval", TimeUnit.MILLISECONDS).millis
+    new CommitterSettings(maxBatch, maxInterval)
+  }
+
+  /**
+   * Java API: Create settings from the default configuration
+   * `akka.kafka.committer`.
+   */
+  def create(actorSystem: ActorSystem): CommitterSettings =
+    apply(actorSystem)
+
+  /**
+   * Java API: Create settings from a configuration with the same layout as
+   * the default configuration `akka.kafka.committer`.
+   */
+  def create(config: Config): CommitterSettings =
+    apply(config)
+
+}
+
+/**
+ * Settings for committer. See `akka.kafka.committer` section in
+ * reference.conf. Note that the [[CommitterSettings companion]] object provides
+ * `apply` and `create` functions for convenient construction of the settings, together with
+ * the `with` methods.
+ */
+class CommitterSettings private (
+    val maxBatch: Long,
+    val maxInterval: FiniteDuration
+) {
+
+  def withMaxBatch(maxBatch: Long): CommitterSettings =
+    copy(maxBatch = maxBatch)
+
+  def withMaxInterval(maxInterval: FiniteDuration): CommitterSettings =
+    copy(maxInterval = maxInterval)
+
+  def withMaxInterval(maxInterval: java.time.Duration): CommitterSettings =
+    copy(maxInterval = maxInterval.asScala)
+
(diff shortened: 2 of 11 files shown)


core/src/main/scala/akka/kafka/CommitterSettings.scala
```diff
@@ -0,0 +1,46 @@
...
+    apply(actorSystem.settings.config.getConfig("akka.kafka.committer"))
+
+  def apply(config: Config): CommitterSettings = {
+    val maxBatch = config.getLong("max-batch")
+    val interval = config.getDuration("interval", TimeUnit.MILLISECONDS).millis
+    new CommitterSettings(maxBatch, interval)
+  }
+}
+
+class CommitterSettings(
```
ennru: Mark the constructor as `private`, please.
```suggestion
class CommitterSettings private (
```

core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -153,6 +148,11 @@ object ConsumerMessage {
      * Java API: Get current offset positions
      */
     def getOffsets(): JMap[GroupTopicPartition, Long]
+
+    /**
+     * Get number of processed messages this batch is built of
+     */
+    def batchSize(): Long
```
ennru:
```suggestion
    def batchSize: Long
```

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scala.concurrent.Future
+
+object Committer {
+
+  def sink(settings: CommitterSettings): Sink[Committable, Future[Done]] =
+    Flow[Committable]
+    // Not very efficient, ideally we should merge offsets instead of grouping them
+      .groupedWeightedWithin(settings.maxBatch, settings.interval)(commitableSize)
+      .map(_.foldLeft(CommittableOffsetBatch.empty)(_ updated _))
```
ennru: We should push that logic down to `CommittableOffsetBatch`, there is an `apply` method for `Seq[CommittableOffset]`. I guess it can be changed to accept `Committable` instead.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
```diff
@@ -0,0 +1,30 @@
...
+      .mapAsync(1)(_.commitScaladsl())
+      .toMat(Sink.ignore)(Keep.right)
+
+  private def commitableSize(committable: Committable): Long =
+    committable match {
+      case batch: CommittableOffsetBatch => batch.batchSize()
+      case single: CommittableOffset => 1
```
ennru: You might as well push `batchSize` up to `Committable`.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
rtimush: Thanks, I haven't noticed that there is such method.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
rtimush: Moved to `Committable`. Didn't do this initially because it felt weird to add batch-related methods to a potentially non-batched `Committable`.

rtimush:
Thanks for the review! I'm not sure about `Committer.oneByOne` — while technically it's trivial, it's also a subcase of the generic sink with `max-batch = 1`. If we add `oneByOne` it might make an impression that this case is somehow special, or that it gives some additional guarantees. Also, in the real environment committing one by one will very likely become a bottleneck, so one will have to switch to the batching alternative anyway, but it will be a code change rather than a configuration change.

core/src/main/scala/akka/kafka/CommitterSettings.scala
```diff
@@ -0,0 +1,75 @@
...
+ */
+class CommitterSettings private (
+    val maxBatch: Long,
+    val maxInterval: FiniteDuration
+) {
+
+  def withMaxBatch(maxBatch: Long): CommitterSettings =
+    copy(maxBatch = maxBatch)
+
+  def withMaxInterval(maxInterval: FiniteDuration): CommitterSettings =
```
ennru: Complement with `withMaxInterval(value: java.time.Duration)`, please.

core/src/main/scala/akka/kafka/CommitterSettings.scala
```diff
@@ -0,0 +1,75 @@
...
+    copy(maxInterval = maxInterval)
+
+  private def copy(maxBatch: Long = maxBatch, maxInterval: FiniteDuration = maxInterval): CommitterSettings =
+    new CommitterSettings(maxBatch, maxInterval)
+
+  override def toString: String =
+    "akka.kafka.CommitterSettings(" +
+    s"maxBatch=$maxBatch," +
+    s"maxInterval=$maxInterval}" +
```
ennru:
```suggestion
    s"maxInterval=${maxInterval.toCoarsest}" +
```

core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -82,6 +82,7 @@ private final case class CommittableOffsetImpl(override val partitionOffset: Con
   override def commitScaladsl(): Future[Done] =
     committer.commit(immutable.Seq(partitionOffset.withMetadata(metadata)))
   override def commitJavadsl(): CompletionStage[Done] = commitScaladsl().toJava
+  override def batchSize: Long = 1
```
ennru:
```suggestion
  override val batchSize: Long = 1
```

core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -96,11 +97,17 @@ private[kafka] trait Committer {
 @InternalApi
 private[kafka] final class CommittableOffsetBatchImpl(
     val offsetsAndMetadata: Map[GroupTopicPartition, OffsetAndMetadata],
-    val stages: Map[String, Committer]
+    val stages: Map[String, Committer],
+    size: Long
```
ennru:
```suggestion
    override val batchSize: Long
```

ennru:
Thank you for the super fast turn-around.

core/src/main/resources/reference.conf
```diff
@@ -103,6 +103,21 @@ akka.kafka.consumer {
...
+# Properties for akka.kafka.CommitterSettings can be
+# defined in this section or a configuration section with
+# the same layout.
+akka.kafka.committer {
+
+  # Maximum number of messages in a single commit batch
+  max-batch = 1000
+
+  # Maximum interval between commits
+  max-interval = 1m
```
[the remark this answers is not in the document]
rtimush: I really had a hard time defining the defaults here. For `kafka-streams` the default commit frequency is every 30s. In our production setup with alpakka-kafka we also couldn't get much below that number.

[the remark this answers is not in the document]
ennru: I changed it to 1000/10 seconds as the default. I think of the defaults to be most important for people starting to use the library -- when they master it, they will understand their needs better and change to suitable values.

PR #623, opened 2018-10-31
Tighten up visibility; sprinkle Scaladoc

ennru:
* Add some Scaladoc to internal classes 
* Make all stage logics are package private
* Add `@InternalApi` for all `private[kafka]` uses
* All Config support in `ConfigSettings` (opened https://github.com/akka/akka/pull/25845 to move some to Akka)


core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
```diff
@@ -17,7 +18,12 @@ import org.apache.kafka.common.TopicPartition
 import scala.annotation.tailrec
 import scala.concurrent.{ExecutionContext, Future}
 
-private[kafka] abstract class BaseSingleSourceLogic[K, V, Msg](
+/**
+ * Internal API.
+ *
+ * Shared GraphStageLogic for [[SingleSourceLogic]] and [[ExternalSingleSourceLogic]].
+ */
+@InternalApi private abstract class BaseSingleSourceLogic[K, V, Msg](
```
2m: The `Internal API` api doc is mostly relevant on code which show up in the generated scaladoc or javadoc. As `private[package]` members still show in in javadoc, this is where `Internal API` is mostly useful.

Here, and below, by making these `private[kafka]` classes `private` we can remove the comment as well.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -45,11 +51,11 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
...
 
-  lazy val decider: Decider =
+  private lazy val decider: Decider =
     inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
-  val awaitingConfirmation = new AtomicInteger(0)
-  @volatile var inIsClosed = false
-  var completionState: Option[Try[Unit]] = None
+  protected val awaitingConfirmation = new AtomicInteger(0)
+  private var inIsClosed = false
+  private var completionState: Option[Try[Done]] = None
```
2m: Reducing the visibility here is great!

ennru:
I thought about removing it, but thought it does not get in the way and might be good in case something needs to be opened up again.

PR #626, opened 2018-11-02
Include requested partition in CommittedOffset

giftig:
Provides context when receiving the response so that you can use the
tell pattern without losing important context about what information you
requested

### DIFF
--- a/core/src/main/scala/akka/kafka/Metadata.scala
+++ b/core/src/main/scala/akka/kafka/Metadata.scala
@@ -170,7 +170,7 @@ object Metadata {
    * [[org.apache.kafka.clients.consumer.KafkaConsumer#committed()]]
    */
   final case class GetCommittedOffset(partition: TopicPartition) extends Request with NoSerializationVerificationNeeded
-  final case class CommittedOffset(response: Try[OffsetAndMetadata])
+  final case class CommittedOffset(response: Try[OffsetAndMetadata], requestedPartition: TopicPartition)
       extends Response
       with NoSerializationVerificationNeeded {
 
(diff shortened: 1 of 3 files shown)


ennru:
This PR changes the API a tiny bit, we'll leave it for 1.0-M2, and make sure the other metadata requests don't suffer from something similar until than.

core/src/main/scala/akka/kafka/Metadata.scala
```diff
@@ -170,7 +170,7 @@ object Metadata {
    * [[org.apache.kafka.clients.consumer.KafkaConsumer#committed()]]
    */
   final case class GetCommittedOffset(partition: TopicPartition) extends Request with NoSerializationVerificationNeeded
-  final case class CommittedOffset(response: Try[OffsetAndMetadata])
+  final case class CommittedOffset(response: Try[OffsetAndMetadata], requestedPartition: TopicPartition)
```
ennru: I believe it's more natural to have 
```suggestion
  final case class CommittedOffset(partition: TopicPartition, response: Try[OffsetAndMetadata])
```

ennru:
The other Metadata requests do not suffer from the same problem.

core/src/main/scala/akka/kafka/Metadata.scala
giftig: Personally I'd expect the other way around, the response to my query as the first argument and a reminder of what my query was second (i.e. the thing I definitely want first and the thing I might also want second). Either way round is fine imo though.

ennru:
Thank you!

PR #627, opened 2018-11-05
Fetch full git history for version detection

2m:
There are two BC issues that are currently on master that have slipped in while the check was not happening.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ before_install:
   - curl -L https://github.com/docker/compose/releases/download/1.22.0/docker-compose-`uname -s`-`uname -m` > docker-compose
   - chmod +x docker-compose
   - sudo mv docker-compose /usr/local/bin
+  # fetch full history for correct current and previous version detection
+  - git fetch --unshallow
 
 script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -10,6 +10,8 @@ before_install:
   - curl -L https://github.com/docker/compose/releases/download/1.22.0/docker-compose-`uname -s`-`uname -m` > docker-compose
   - chmod +x docker-compose
   - sudo mv docker-compose /usr/local/bin
+  # fetch full history for correct current and previous version detection
+  - git fetch --unshallow
```
patriknw: is this something we need to change also in other projects?

2m: Yes, for the projects that use `previousStableVersion` to derive `mimaPreviousArtifacts`.

2m: According to the GitHub code search the only other project that used `previousStableVersion` is Alpakka. The fix is here: https://github.com/akka/alpakka/pull/1305

https://github.com/search?q=org%3Aakka+previousStableVersion&type=Code
https://github.com/search?q=org%3Alightbend+previousStableVersion&type=Code

PR #636, opened 2018-11-07
Fail with more specific errors on exceeded wakeup...

ennru:
... and pass that error to the stages

This is a follow-up on #617


tests/src/it/resources/logback-test.xml
```diff
@@ -23,7 +23,7 @@
     <logger name="org.apache.kafka.common.utils.AppInfoParser" level="ERROR"/>
     <logger name="org.apache.kafka.clients.NetworkClient" level="ERROR"/>
 
-    <root level="DEBUT">
```
2m: Yup. That is not the first gig for this logger.

2m:
Failure was https://github.com/akka/alpakka-kafka/issues/637

Issue #611, opened 2018-10-12
Failed: PartitionedSourceSpec partitioned source should resume topics with demand


2m:
https://travis-ci.org/akka/alpakka-kafka/jobs/440658900

```
[info] PartitionedSourceSpec:
[info] partitioned source
[info] - should resume topics with demand *** FAILED *** (430 milliseconds)
[info]   Set(topic-1) was not empty (PartitionedSourceSpec.scala:77)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:346)
[info]   at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6668)
[info]   at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6704)
[info]   at akka.kafka.internal.PartitionedSourceSpec$$anonfun$1$$anonfun$apply$1.apply(PartitionedSourceSpec.scala:77)
[info]   at akka.kafka.internal.PartitionedSourceSpec$$anonfun$1$$anonfun$apply$1.apply(PartitionedSourceSpec.scala:52)
```

It seems that #601 does not want to go away quietly.

ennru:
Experiencing it repeatedly today https://travis-ci.org/akka/alpakka-kafka/jobs/449473247

ennru:
See even #601 

It happened again in https://travis-ci.org/akka/alpakka-kafka/jobs/615474265#L439

ennru:
https://travis-ci.org/akka/alpakka-kafka/jobs/616606436#L539

ennru:
Recent failures are different than the original report:
```
[info] partitioned source
[info] - should resume topics with demand *** FAILED *** (10 seconds, 347 milliseconds)
[info]   java.lang.AssertionError: Expected OnNext(_), yet no element signaled during 10 seconds
```

seglo:
Now on GitHub Actions:

https://github.com/akka/alpakka-kafka/runs/1865861742?check_suite_focus=true#step:6:3181

johanandren:
https://github.com/akka/alpakka-kafka/actions/runs/4721371265/jobs/8374740353?pr=1622#step:6:3202

PR #647, opened 2018-11-18
Committer updates

ennru:
* Add `Committer.flow`
* Add setting for parallelism in `Committer`
* Use type-bound instead of strict type in Java API
* Prefer `Committer` over direct committing API
* Prefer `DrainingControl` in all snippets


schrepfler:
Is the `Commiter.flow` something which is Kafka specific or it can be used with other `Source`s which are comittable (example MQTT)?

2m:
Failure was https://github.com/akka/alpakka-kafka/issues/611

2m:
@schrepfler the implementation of the flow uses Kafka specific APIs to commit messages. However this pattern could be implemented in other Alpakka connectors as well.

schrepfler:
I think unifying it behind one interface/trait for all alpakka projects would be a big win.

2m:
I agree! That is the long-term goal: having an abstraction that could be used for committing and passthroughs in various connectors.

PR #649, opened 2018-11-20
Use val instead of def for rebalanceListener

muditsin:
There is no use of rebalanceListener being a def and a function call happening.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
@@ -46,7 +46,7 @@ import scala.concurrent.{Future, Promise}
       log.log(partitionLogLevel, "Revoked partitions: {}. All partitions: {}", revokedTps, tps)
     }
 
-    def rebalanceListener: KafkaConsumerActor.ListenerCallbacks =
+    val rebalanceListener: KafkaConsumerActor.ListenerCallbacks =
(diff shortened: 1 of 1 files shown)


muditsin:
run this build

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -46,7 +46,7 @@ import scala.concurrent.{Future, Promise}
       log.log(partitionLogLevel, "Revoked partitions: {}. All partitions: {}", revokedTps, tps)
     }
 
-    def rebalanceListener: KafkaConsumerActor.ListenerCallbacks =
```
muditsin: I see it was earlier `def` as rebalanceListener would not need to be evaluated in the case of a `ManualSubscription`. Even in this case, wouldn't `lazy val` be better suited?

helena: It seems like `rebalanceListener`, `partitionAssignedCB` and `partitionRevokedCB` may or may not be called and `def` on all would be preferred.

ennru:
Hi @muditsin, thank you for engaging in the Alpakka Kafka connector. As you already found out the listener is not always needed so let's only construct it if needed.
This bit is not run very often, so it is not really worth diving into micro-optimisation just here.
But please, keep going and improve things!

PR #655, opened 2018-11-21
Atomically update topic partitions in the Consumer Dummy

2m:
This is required as `tps` can be concurrently updated from the test case and from the thread that is running `KafkaConsumerActor`.


ennru:
Scala 2.11 doesn't like this.

2m:
Added little something for Scala 2.11.

PR #656, opened 2018-11-21
Close Admin Clients in KafkaSpec Cleanup / Junit testkit improvements

johnclara:
Admin Clients are currently not closed after the KafkaSpec cleanup.
In our tests I see a bunch of warnings:
```
23:35:41.206 [kafka-admin-client-thread | adminclient-1] WARN  o.apache.kafka.clients.NetworkClient - [AdminClient clientId=adminclient-1] Connection to node 0 could not be established. Broker may not be available.
23:35:42.344 [kafka-admin-client-thread | adminclient-1] WARN  o.apache.kafka.clients.NetworkClient - [AdminClient clientId=adminclient-1] Connection to node 0 could not be established. Broker may not be available.
```
Changes in this PR include:
Only one new and old admin client will be used per KafkaSpec.
A new and old admin client will be created even if they're not needed.
The new and old admin clients will be closed on cleanup.

### DIFF
--- a/testkit/src/main/scala/akka/kafka/testkit/scaladsl/KafkaSpec.scala
+++ b/testkit/src/main/scala/akka/kafka/testkit/scaladsl/KafkaSpec.scala
@@ -57,6 +57,10 @@ abstract class KafkaSpec(val kafkaPort: Int, val zooKeeperPort: Int, actorSystem
     extends TestKit(actorSystem)
     with KafkaTestKit {
 
+  // children should setup() and use admin or oldAdmin instead of constructors
+  private def adminClient(): AdminClient = super.adminClient()
+  private def oldAdminClient(): OldAdminClient = super.oldAdminClient()
+
   def this(kafkaPort: Int) = this(kafkaPort, kafkaPort + 1, ActorSystem("Spec"))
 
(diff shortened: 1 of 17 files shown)


testkit/src/main/scala/akka/kafka/testkit/scaladsl/KafkaSpec.scala
```diff
@@ -55,6 +56,10 @@ abstract class KafkaSpec(val kafkaPort: Int, val zooKeeperPort: Int, actorSystem
     extends TestKit(actorSystem)
     with KafkaTestKit {
 
+  // children should setup() and use admin or oldAdmin instead of constructors
+  private def adminClient(): AdminClient = super.adminClient()
```
johnclara: @2m Is there a better way of removing visibility to children in Scala?

ennru: Renaming the factory methods in `internal.KafkaTestKit` to `createXyz` is the best way, I guess as it indicates the caller has some responsibility.

johnclara:
LGTM! Sorry, haven't had a change to look at this recently

ennru:
Thank you @johnclara for initiating quite some improvements to the testkit!

PR #660, opened 2018-11-23
Kafka 2.1.0 client

ennru:
This upgrade enables the use of the zstandard compression (with Kafka 2.1 brokers).

[Kafka release notes](https://dist.apache.org/repos/dist/release/kafka/2.1.0/RELEASE_NOTES.html)

[Kafka upgrade instructions](http://kafka.apache.org/21/documentation.html#upgrade):
> Note that 2.1.x contains a change to the internal schema used to store consumer offsets. Once the upgrade is complete, it will not be possible to downgrade to previous versions. See the rolling upgrade notes below for more detail.


ennru:
Failure was #683 (which is fixed in master)

PR #661, opened 2018-11-26
Use def for subscription callbacks as they are not always used

muditsin:
Based on https://github.com/akka/alpakka-kafka/pull/649#discussion_r235253314 by @helena


muditsin:
Also note, `lazy val` could be used if preferred.

helena:
These two internal functions are only conditionally called once, in one potential case
```scala
def partitionAssignedCB = getAsyncCallback[Set[TopicPartition]] 
def partitionRevokedCB = getAsyncCallback[Set[TopicPartition]]
```

The internal `ListenerCallbacks` is also only conditionally called once, in two of the three potential cases. 
```scala
 def rebalanceListener: KafkaConsumerActor.ListenerCallbacks
```

If `subscription match { case s: ManualSubscription => configureManualSubscription(s) }`  none of the above are called AFACT, why `val` it? I must be missing something :)

muditsin:
@helena, I think I understand what @ennru is saying. Everytime a partition is assigned or revoked, the function passed in `rebalanceListener` would call `partitionAssignedCB` or `partitionRevokedCB`.
@ennru please correct me if I am wrong.
I will revert them as `val`s and move them into `rebalanceListener`

helena:
@muditsin @ennru ah, I am getting familiar with this project so thank you for your patience. I dug in further to see the `getAsyncCallback` doc states that the callback `can be used asynchronously to re-enter the current [[GraphStage]] with an asynchronous notification`. Thanks 👍

PR #662, opened 2018-11-26
Build paradox and api docs with sbt-site

2m:
Do not merge until we have where to deploy both of these.


2m:
Added preprocessing that fixes scala API docs. Also added publishing to gustav, which can be tested after merging while building the master branch.

PR #664, opened 2018-11-27
Rename Committer to InternalCommitter to reduce confusion

ennru:
...with the newish user API `Committer` (The internal class will show in completion for Java users as it is `private[kafka]`).


2m:
Mima is not happy. We can ignore everything internal with `akka.kafka.internal.*`

PR #669, opened 2018-11-30
Add Scaladoc flags

ennru:
* Add linking to sources from Scaladoc
* Set title and version in Scaladoc
* Exclude the mysterious `akka.pattern` package

TODO: select the correct tree for the Github link

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,4 @@
+import com.lightbend.paradox.markdown.GitHubResolver
 import sbt.Resolver
 enablePlugins(AutomateHeaderPlugin)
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -66,6 +66,8 @@ val kafkaScale = settingKey[Int]("Number of kafka docker containers")
 
 resolvers in ThisBuild ++= Seq(Resolver.bintrayRepo("manub", "maven"))
 
+// TODO read version or master
+val githubTree = "master"
```
2m: As Paradox does try to figure out the correct github tree itself, maybe we could reuse that value here as well? Or do you think it might be too hacky to rely on paradox property here?

ennru: Yes, that seems hacky to me. I looked into copying the logic from there, but it was not completely straightforward.
How easy would it be to access the Paradox property?

2m: paradoxProperties.value.get(GitHubResolver.baseUrl)

ennru: It gets more involved as that is available only in the `docs` project (which inherits settings from `CommonSettings`...
I went with duplicating the URL and some logic.

Issue #609, opened 2018-10-11
Log configuration in KafkaConsumerActor preStart


ennru:
Just as the Kafka client logs its configuration the actor could dump its configuration on debug level.

ennru:
And maybe adding `BuildInfo` and include the library version number into that log message would be a good idea.

muditsin:
I could try to take this up. What configurations are you looking to log? the `akka.kafka.consumer` from `context.system.settings`?

ennru:
Yes, when a Consumer stage or Producer stage is materialized it could log the `ConsumerSettings` or `ProducerSettings`. Just as Kafka consumers log their settings when started.

PR #676, opened 2018-12-06
WIP: Log configuration in KafkaConsumerActor preStart

muditsin:
Fixes: https://github.com/akka/alpakka-kafka/issues/609

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -315,6 +315,8 @@ import scala.util.control.{NoStackTrace, NonFatal}
   override def preStart(): Unit = {
     super.preStart()
     try {
+      if (log.isDebugEnabled)
+        log.debug(s"Creating Kafka consumer with ${settings.toString}")
       consumer = settings.createKafkaConsumer()
     } catch {
       case e: Exception =>


ennru:
Thanks for looking into this. We'd need to use `ConsumerSettings.toString` instead, as the settings can be tweaked in the code.
And it must be wrapped with `if (log.isDebugEnabled)`.

muditsin:
@ennru, my pleasure! :) Updated based on your comment. Is there a test to add for this? If so, how would I go about it?

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -315,6 +315,8 @@ import scala.util.control.{NoStackTrace, NonFatal}
   override def preStart(): Unit = {
     super.preStart()
     try {
+      if (log.isDebugEnabled)
+        log.debug(settings.toString)
```
ennru: Some note "creating Kafka consumer with" would be great.

muditsin: @ennru Updated

muditsin:
> And it must be wrapped with `if (log.isDebugEnabled)`.
Although I have added it, shouldn't the logger take care of something like the `isDebugEnabled`?

ennru:
Yes, it won't log the message, but the `toString` would still be called before it detects that it doesn't need the `String`.

ennru:
Thank you!

Issue #683, opened 2018-12-10
Failed: Partitioned source must handle exceptions in stream without commit failures


ennru:
This test seems to fail more likely on JDK 11.

```
[info] PartitionedSourcesSpec:
[info] Partitioned source
[info] - must handle exceptions in stream without commit failures *** FAILED *** (6 seconds, 871 milliseconds)
[info]   The future returned an exception of type: java.util.NoSuchElementException, with message: Future.failed not completed with a throwable.. (PartitionedSourcesSpec.scala:496)
```

PR #686, opened 2018-12-10
Remove the three wakeup-related settings

### LINKED ISSUE #684: Kafka 2 polling: Remove unused settings
The wakeup handling goes away with the new `poll(Duration)` method, thus
* wakeup-timeout
* max-wakeups
* wakeup-debug
have no effect anymore and should be removed.

ennru:
Fixes #684

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -54,29 +54,19 @@ akka.kafka.consumer {
 
   # If commits take longer than this time a warning is logged
   commit-time-warning = 1s
-  
-  # If for any reason `KafkaConsumer.poll` blocks for longer than the configured
-  # poll-timeout then it is forcefully woken up with `KafkaConsumer.wakeup`.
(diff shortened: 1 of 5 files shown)


ennru:
Failure is #683

core/src/main/resources/reference.conf
```diff
@@ -54,29 +54,19 @@ akka.kafka.consumer {
...
-  # DEPRECATED: No op as waking up consumer is not used anymore
-  wakeup-timeout = 3s
-
-  # After exceeding maximum wakeups the consumer will stop and the stage will fail.
-  # Setting it to 0 will let it ignore the wakeups and try to get the polling done forever.
-  # DEPRECATED: No op as waking up consumer is not used anymore
-  max-wakeups = 10
+
+  # Not used anymore (since 1.0-RC1)
+  # wakeup-timeout = 3s
```
2m: Should we just remove these altogether?

ennru: My thinking was to leave them (during 1.0.x) so people looking at their existing configs understand that they may remove them.

2m: Alright.

PR #691, opened 2018-12-17
Use JUnit5

ennru:
Benefits
* JUnit 5 supports test class lifecycle control
  * useful for simpler admin client use in tests
  * useful for simpler embedded Kafka support
* JUnit 5 makes use of Java 8 features (mainly lambdas)

This PR
* adds [sbt-jupiter-interface](https://github.com/maichler/sbt-jupiter-interface) plugin to run JUnit 5-annotated tests
* introduces `KafkaTest` and `EmbeddedKafkaTest` which rely on JUnit 5 per-class lifecyle
* changes `ProducerExampleTest` and `ConsumerExampleTest` to use JUnit 5 instead
* re-implements the JUnit base-classes for use of *EmbeddedKafka* in Java and makes use of constructor parameters instead of overrides (which look weird in Java)


build.sbt
```diff
@@ -113,11 +120,12 @@ val commonSettings = Seq(
     "akka.pattern" // for some reason Scaladoc creates this
   ),
   // show full stack traces and test case durations
-  testOptions += Tests.Argument("-oDF"),
+  testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oDF"),
   // -a Show stack traces and exception class name for AssertionErrors.
   // -v Log "test run started" / "test started" / "test run finished" events on log level "info" instead of "debug".
   // -q Suppress stdout for successful tests.
   testOptions += Tests.Argument(TestFrameworks.JUnit, "-a", "-v", "-q"),
+  testOptions += Tests.Argument(jupiterTestFramework, "-a", "-v", "-q"),
```
2m: It is interesting that with this change these options are still used when running JUnit 4 tests.

project/plugins.sbt
```diff
@@ -20,3 +20,6 @@ addSbtPlugin("com.github.ehsanyou" % "sbt-docker-compose" % "67284e73-envvars-2m
 // depend directly on the patched version see https://github.com/akka/alpakka/issues/1388
 addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2+10-148ba0ff")
 resolvers += Resolver.bintrayIvyRepo("2m", "sbt-plugins")
+
+addSbtPlugin("net.aichler" % "sbt-jupiter-interface" % "0.8.0")
+resolvers += Resolver.jcenterRepo
```
2m: When using `coursier` I need this resolver also in `build.sbt`

testkit/src/main/scala/akka/kafka/testkit/javadsl/KafkaTest.scala
```diff
@@ -0,0 +1,94 @@
...
+  @AfterAll def cleanUpAdmin() = cleanUpAdminClient()
+
+  @AfterEach def checkForStageLeaks() = StreamTestKit.assertAllStagesStopped(materializer)
+
+  /**
+   * Overwrite to set different default timeout for [[KafkaTest.resultOf]].
+   */
+  def resultOfTimeout: java.time.Duration = java.time.Duration.ofSeconds(5)
+
+  def produceString(topic: String, messageCount: Int, partition: Int): CompletionStage[Done] =
```
2m: Common code between JUnit4 and Junit5 tests should be moved to a common abstract class.

tests/src/test/java/docs/javadsl/ConsumerExampleTest.java
```diff
@@ -45,7 +46,8 @@
 
 import static org.junit.Assert.assertEquals;
 
-public class ConsumerExampleTest extends EmbeddedKafkaJunit4Test {
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
2m: Good to have per class test case instances!

build.sbt
```diff
@@ -171,6 +179,7 @@ lazy val `alpakka-kafka` =
 
 lazy val core = project
   .enablePlugins(AutomateHeaderPlugin)
+  // see https://github.com/maichler/sbt-jupiter-interface/issues/24
```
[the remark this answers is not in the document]
ennru: The comment was a left-over. The plugin changed behaviour (on my inquiry) to run tests on a project when it detects `jupiter` as library dependency and it fails the build if it finds JUnit tests and no library.

build.sbt
```diff
@@ -38,6 +41,10 @@ val testDependencies = Seq(
   "com.fasterxml.jackson.core" % "jackson-databind" % "2.9.7" % Test, // ApacheV2
   "com.novocode" % "junit-interface" % "0.11" % Test,
   "junit" % "junit" % junit4Version % Test,
+  // See http://hamcrest.org/JavaHamcrest/distributables#upgrading-from-hamcrest-1x
+  "org.hamcrest" % "hamcrest-library" % "2.1" % Test,
+  "org.hamcrest" % "hamcrest" % "2.1" % Test,
+  "net.aichler" % "jupiter-interface" % jupiterInterfaceVersion % Test,
```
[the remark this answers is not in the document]
ennru: The `.value` macro is not available to the dependency listings. Why don't we just move those into the project settings sections?

project/plugins.sbt
ennru: Sorry, forgot to add it there.

build.sbt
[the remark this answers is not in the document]
2m: Yup. We can move these over. I think some of these were being reused across projects some time ago. But not anymore.

build.sbt
ennru: The JUnit 4 tests are now run by Jupiter.

testkit/src/main/scala/akka/kafka/testkit/internal/KafkaTestKit.scala
```diff
@@ -109,6 +107,11 @@ trait KafkaTestKit {
   }
 }
 
-object KafkaTestKit {
+abstract class KafkaTestKitClass extends KafkaTestKit
```
ennru: I needed to add this class so that Java can properly inherit from it.

build.sbt
2m: That makes sense. Then we should be able to get rid of the `"com.novocode" % "junit-interface"` dependency.

ennru:
I switched to use the Juniper Vintage Engine to run the JUnit 4 tests.
It has an issue with formatting https://github.com/maichler/sbt-jupiter-interface/issues/33 and `AssignmentTest.mustConsumeFromTheSpecifiedTopicPattern` fails for me locally when using it?!

PR #698, opened 2019-01-04
Changes for running benchmarks on Jenkins

### LINKED ISSUE #688: Capture benchmark results to compare over time
The benchmarks run regularly on Travis, but the results are not captured.
It would be great to keep the results of the runs, and possibly run the benchmarks in an environment where the values are more comparable over time.

2m:
This PR changes Travis configuration to only compile benchmark source files.

Benchmarks are now being run nightly twice on the Akka Jenkins cluster: https://jenkins.akka.io:8498/job/alpakka-kafka-perf-master/plot/ Last 40 runs are stored and plotted.

Fixes #688

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,6 +17,7 @@ val coreDependencies = Seq(
 )
 
 val testkitDependencies = Seq(
+  "com.typesafe.akka" %% "akka-testkit" % akkaVersion,
   "com.typesafe.akka" %% "akka-stream-testkit" % akkaVersion,
   "net.manub" %% "scalatest-embedded-kafka" % "2.0.0" exclude ("log4j", "log4j"),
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -17,6 +17,7 @@ val coreDependencies = Seq(
 )
 
 val testkitDependencies = Seq(
+  "com.typesafe.akka" %% "akka-testkit" % akkaVersion,
```
ennru: Why did you add it to testkit and not only benchmarks?

2m: On the Jenkins node testkit did not compile with missing classes from the testkit dependency. Note that alpakka-kafka testkit already depended transitively on akka-testkit because of akka-stream-testkit. So this only adds transitive dependency explicitly.

PR #701, opened 2019-01-09
Keep to the commit refresh deadlines per partition

Philippus:
Applying this PR will:

- instead of one commit refresh deadline, keep a map of deadlines keyed by partition
- refresh only the commits of partitions with overdue deadlines

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -343,9 +343,8 @@ import scala.util.control.NonFatal
 
   private def receivePoll(p: Poll[_, _]): Unit =
     if (p.target == this) {
-      if (commitRefreshDeadlines.exists(_._2.isOverdue())) {
-        val overdueTps = commitRefreshDeadlines.filter(_._2.isOverdue()).keySet
-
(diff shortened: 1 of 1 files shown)


Philippus:
That's a good idea, I will work on it.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -342,13 +343,20 @@ import scala.util.control.NonFatal
 
   private def receivePoll(p: Poll[_, _]): Unit =
     if (p.target == this) {
-      if (commitRefreshDeadline.exists(_.isOverdue())) {
+      if (commitRefreshDeadlines.exists(_._2.isOverdue())) {
+        val overdueTps = commitRefreshDeadlines.filter(_._2.isOverdue()).keySet
```
ennru: `exists` and `filter` look like a repetition.

Philippus: Yes, look like, but the `if`-statement is guarding the whole section that should only run if there are overdue deadlines.

ennru: Yes, please.

ennru:
Thank you! Great fix.

PR #705, opened 2019-01-11
Add partitioned transactional source

charlibot:
By adding a partitioned transactional source, a new producer will be created per topic-partition when hooked up with the sink or flow. If the `transactional.id` is based on the topic-partition then can get around the restriction of distributing the workload across multiple instances without getting `ProducerFencedException`s.

This was based on reviewing the comments in this thread: https://github.com/akka/alpakka-kafka/pull/420#pullrequestreview-111036369

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
+++ b/core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
@@ -46,6 +46,8 @@ private[kafka] trait TransactionalMessageBuilderBase[K, V, Msg] extends MessageB
   def committedMarker: CommittedMarker
 
   def onMessage(consumerMessage: ConsumerRecord[K, V]): Unit
+
+  val fromPartitionedSource: Boolean
 }
 
 /** Internal API */
@@ -54,15 +56,17 @@ private[kafka] trait TransactionalMessageBuilder[K, V]
     extends TransactionalMessageBuilderBase[K, V, TransactionalMessage[K, V]] {
   override def createMessage(rec: ConsumerRecord[K, V]): TransactionalMessage[K, V] = {
     onMessage(rec)
-    val offset = PartitionOffsetCommittedMarker(
-      GroupTopicPartition(
-        groupId = groupId,
-        topic = rec.topic,
-        partition = rec.partition
-      ),
-      offset = rec.offset,
-      committedMarker
+    val groupTopicPartition = GroupTopicPartition(
+      groupId = groupId,
+      topic = rec.topic,
+      partition = rec.partition
     )
+    val offset = if (fromPartitionedSource) {
+      new PartitionOffsetCommittedMarker(groupTopicPartition, offset = rec.offset, committedMarker)
+      with FromPartitionedSource
+    } else {
+      PartitionOffsetCommittedMarker(groupTopicPartition, offset = rec.offset, committedMarker)
+    }
     ConsumerMessage.TransactionalMessage(rec, offset)
   }
 }
--- a/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
@@ -30,19 +30,24 @@ import scala.util.{Failure, Success}
 /**
  * Internal API.
  *
- * Anonymous sub-class instance is created in [[CommittableSubSource]].
+ * Anonymous sub-class instance is created in [[CommittableSubSource]], [[PlainSubSource]] and [[TransactionalSubSource]].
  */
 @InternalApi
-private abstract class SubSourceLogic[K, V, Msg](
+private class SubSourceLogic[K, V, Msg](
     val shape: SourceShape[(TopicPartition, Source[Msg, NotUsed])],
     settings: ConsumerSettings[K, V],
     subscription: AutoSubscription,
+    msgSourceLogicFactory: (SourceShape[Msg],
+                            TopicPartition,
+                            ActorRef,
+                            AsyncCallback[(TopicPartition, (Control, ActorRef))],
+                            AsyncCallback[(TopicPartition, Option[ConsumerRecord[K, V]])],
+                            Int) => MessageSubSourceLogic[K, V, Msg],
     getOffsetsOnAssign: Option[Set[TopicPartition] => Future[Map[TopicPartition, Long]]] = None,
     onRevoke: Set[TopicPartition] => Unit = _ => ()
 ) extends TimerGraphStageLogic(shape)
     with PromiseControl
     with MetricsControl
-    with MessageBuilder[K, V, Msg]
     with StageLogging {
   import SubSourceLogic._
 
@@ -58,7 +63,7 @@ private abstract class SubSourceLogic[K, V, Msg](
 
   /** We have created a source for these partitions, but it has not started up and is not in subSources yet. */
   var partitionsInStartup: immutable.Set[TopicPartition] = immutable.Set.empty
-  var subSources: Map[TopicPartition, Control] = immutable.Map.empty
+  var subSources: Map[TopicPartition, (Control, ActorRef)] = immutable.Map.empty
 
   /** Kafka has signalled these partitions are revoked, but some may be re-assigned just after revoking. */
   var partitionsToRevoke: Set[TopicPartition] = Set.empty
@@ -81,7 +86,11 @@ private abstract class SubSourceLogic[K, V, Msg](
     sourceActor.watch(consumerActor)
 
     def rebalanceListener =
-      KafkaConsumerActor.ListenerCallbacks(subscription, sourceActor.ref, partitionAssignedCB, partitionRevokedCB)
+      KafkaConsumerActor.ListenerCallbacks(subscription,
+                                           sourceActor.ref,
+                                           partitionAssignedCB,
+                                           partitionRevokedCB,
+                                           blockingRevokedHandler)
 
     subscription match {
       case TopicSubscription(topics, _) =>
@@ -152,6 +161,8 @@ private abstract class SubSourceLogic[K, V, Msg](
     scheduleOnce(CloseRevokedPartitions, settings.waitClosePartition)
   }
 
+  protected def blockingRevokedHandler(revokedTps: Set[TopicPartition]): Unit = {}
+
   override def onTimer(timerKey: Any): Unit = timerKey match {
     case CloseRevokedPartitions =>
       if (log.isDebugEnabled) {
@@ -160,7 +171,7 @@ private abstract class SubSourceLogic[K, V, Msg](
       onRevoke(partitionsToRevoke)
       pendingPartitions --= partitionsToRevoke
       partitionsInStartup --= partitionsToRevoke
-      partitionsToRevoke.flatMap(subSources.get).foreach(_.shutdown())
+      partitionsToRevoke.flatMap(subSources.get).map(_._1).foreach(_.shutdown())
       subSources --= partitionsToRevoke
       partitionsToRevoke = Set.empty
   }
@@ -181,17 +192,18 @@ private abstract class SubSourceLogic[K, V, Msg](
         }
     }
 
-  val subsourceStartedCB: AsyncCallback[(TopicPartition, Control)] = getAsyncCallback[(TopicPartition, Control)] {
-    case (tp, control) =>
-      if (!partitionsInStartup.contains(tp)) {
-        // Partition was revoked while
-        // starting up.  Kill!
-        control.shutdown()
-      } else {
-        subSources += (tp -> control)
-        partitionsInStartup -= tp
-      }
-  }
+  val subsourceStartedCB: AsyncCallback[(TopicPartition, (Control, ActorRef))] =
+    getAsyncCallback[(TopicPartition, (Control, ActorRef))] {
+      case (tp, value @ (control, actorRef)) =>
+        if (!partitionsInStartup.contains(tp)) {
+          // Partition was revoked while
+          // starting up.  Kill!
+          control.shutdown()
+        } else {
+          subSources += (tp -> value)
+          partitionsInStartup -= tp
+        }
+    }
 
   setHandler(shape.out, new OutHandler {
     override def onPull(): Unit =
@@ -212,14 +224,14 @@ private abstract class SubSourceLogic[K, V, Msg](
 
       pendingPartitions = pendingPartitions.tail
       partitionsInStartup += tp
-      val subSource = Source.fromGraph(
-        new SubSourceStage(tp,
-                           consumerActor,
-                           subsourceStartedCB,
-                           subsourceCancelledCB,
-                           messageBuilder = this,
-                           actorNumber)
-      )
+      val graphStage = new GraphStage[SourceShape[Msg]] {
+        val out = Outlet[Msg]("out")
+        val shape = new SourceShape(out)
+
+        override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+          msgSourceLogicFactory(shape, tp, consumerActor, subsourceStartedCB, subsourceCancelledCB, actorNumber)
+      }
+      val subSource = Source.fromGraph(graphStage)
       push(shape.out, (tp, subSource))
       emitSubSourcesForPendingPartitions()
     }
@@ -233,7 +245,7 @@ private abstract class SubSourceLogic[K, V, Msg](
   override def performStop(): Unit = {
     setKeepGoing(true)
     subSources.foreach {
-      case (_, control) => control.stop()
+      case (_, (control, _)) => control.stop()
     }
     complete(shape.out)
     onStop()
@@ -243,7 +255,7 @@ private abstract class SubSourceLogic[K, V, Msg](
     setKeepGoing(true)
     //todo we should wait for subsources to be shutdown and next shutdown main stage
     subSources.foreach {
-      case (_, control) => control.shutdown()
+      case (_, (control, _)) => control.shutdown()
     }
 
     if (!isClosed(shape.out)) {
@@ -268,89 +280,91 @@ private object SubSourceLogic {
 }
 
 /** Internal API */
-private final class SubSourceStage[K, V, Msg](
+@InternalApi
+private abstract class MessageSubSourceLogic[K, V, Msg](
+    val shape: SourceShape[Msg],
     tp: TopicPartition,
     consumerActor: ActorRef,
-    subSourceStartedCb: AsyncCallback[(TopicPartition, Control)],
+    subSourceStartedCb: AsyncCallback[(TopicPartition, (Control, ActorRef))],
     subSourceCancelledCb: AsyncCallback[(TopicPartition, Option[ConsumerRecord[K, V]])],
-    messageBuilder: MessageBuilder[K, V, Msg],
     actorNumber: Int
-) extends GraphStage[SourceShape[Msg]] { stage =>
-  val out = Outlet[Msg]("out")
-  val shape = new SourceShape(out)
+) extends GraphStageLogic(shape)
+    with MessageBuilder[K, V, Msg]
+    with PromiseControl
+    with MetricsControl
+    with StageLogging {
 
-  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
-    new GraphStageLogic(shape) with PromiseControl with MetricsControl with StageLogging {
-      override def executionContext: ExecutionContext = materializer.executionContext
-      override def consumerFuture: Future[ActorRef] = Future.successful(consumerActor)
-      val shape = stage.shape
-      val requestMessages = KafkaConsumerActor.Internal.RequestMessages(0, Set(tp))
-      var requested = false
-      var subSourceActor: StageActor = _
-      var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
+  override def executionContext: ExecutionContext = materializer.executionContext
+  override def consumerFuture: Future[ActorRef] = Future.successful(consumerActor)
+  val requestMessages = KafkaConsumerActor.Internal.RequestMessages(0, Set(tp))
+  var requested = false
+  var subSourceActor: StageActor = _
+  var buffer: Iterator[ConsumerRecord[K, V]] = Iterator.empty
 
-      override def preStart(): Unit = {
-        log.debug("#{} Starting SubSource for partition {}", actorNumber, tp)
-        super.preStart()
-        subSourceStartedCb.invoke(tp -> this.asInstanceOf[Control])
-        subSourceActor = getStageActor {
-          case (_, msg: KafkaConsumerActor.Internal.Messages[K, V]) =>
-            requested = false
-            // do not use simple ++ because of https://issues.scala-lang.org/browse/SI-9766
-            if (buffer.hasNext) {
-              buffer = buffer ++ msg.messages
-            } else {
-              buffer = msg.messages
-            }
-            pump()
-          case (_, Status.Failure(e)) =>
-            failStage(e)
-          case (_, Terminated(ref)) if ref == consumerActor =>
-            failStage(new ConsumerFailed)
-        }
-        subSourceActor.watch(consumerActor)
+  override def preStart(): Unit = {
+    log.debug("#{} Starting SubSource for partition {}", actorNumber, tp)
+    super.preStart()
+    subSourceActor = getStageActor(messageHandling)
+    subSourceActor.watch(consumerActor)
+    val controlAndActor = (this.asInstanceOf[Control], subSourceActor.ref)
+    subSourceStartedCb.invoke(tp -> controlAndActor)
+  }
+
+  protected def messageHandling: PartialFunction[(ActorRef, Any), Unit] = {
+    case (_, msg: KafkaConsumerActor.Internal.Messages[K, V]) =>
+      requested = false
+      // do not use simple ++ because of https://issues.scala-lang.org/browse/SI-9766
+      if (buffer.hasNext) {
+        buffer = buffer ++ msg.messages
+      } else {
+        buffer = msg.messages
       }
+      pump()
+    case (_, Status.Failure(e)) =>
+      failStage(e)
+    case (_, Terminated(ref)) if ref == consumerActor =>
+      failStage(new ConsumerFailed)
+  }
 
-      override def postStop(): Unit = {
-        onShutdown()
-        super.postStop()
-      }
+  override def postStop(): Unit = {
+    onShutdown()
+    super.postStop()
+  }
 
-      setHandler(
-        out,
-        new OutHandler {
-          override def onPull(): Unit =
-            pump()
+  setHandler(
+    shape.out,
+    new OutHandler {
+      override def onPull(): Unit =
+        pump()
 
-          override def onDownstreamFinish(): Unit = {
-            val firstUnconsumed = if (buffer.hasNext) {
-              Some(buffer.next())
-            } else {
-              None
-            }
-
-            subSourceCancelledCb.invoke(tp -> firstUnconsumed)
-            super.onDownstreamFinish()
-          }
+      override def onDownstreamFinish(): Unit = {
+        val firstUnconsumed = if (buffer.hasNext) {
+          Some(buffer.next())
+        } else {
+          None
         }
-      )
 
-      def performShutdown() = {
-        log.debug("#{} Completing SubSource for partition {}", actorNumber, tp)
-        completeStage()
+        subSourceCancelledCb.invoke(tp -> firstUnconsumed)
+        super.onDownstreamFinish()
       }
+    }
+  )
 
-      @tailrec
-      private def pump(): Unit =
-        if (isAvailable(out)) {
-          if (buffer.hasNext) {
-            val msg = buffer.next()
-            push(out, messageBuilder.createMessage(msg))
-            pump()
-          } else if (!requested) {
-            requested = true
-            consumerActor.tell(requestMessages, subSourceActor.ref)
-          }
-        }
+  def performShutdown() = {
+    log.debug("#{} Completing SubSource for partition {}", actorNumber, tp)
+    completeStage()
+  }
+
+  @tailrec
+  private def pump(): Unit =
+    if (isAvailable(shape.out)) {
+      if (buffer.hasNext) {
+        val msg = buffer.next()
+        push(shape.out, createMessage(msg))
+        pump()
+      } else if (!requested) {
+        requested = true
+        consumerActor.tell(requestMessages, subSourceActor.ref)
+      }
     }
 }
--- a/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
@@ -7,7 +7,7 @@ package akka.kafka.internal
 import akka.Done
 import akka.annotation.InternalApi
 import akka.kafka.ConsumerMessage
-import akka.kafka.ConsumerMessage.{GroupTopicPartition, PartitionOffsetCommittedMarker}
+import akka.kafka.ConsumerMessage.{FromPartitionedSource, GroupTopicPartition, PartitionOffsetCommittedMarker}
 import akka.kafka.ProducerMessage.{Envelope, Results}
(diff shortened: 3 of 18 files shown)


ennru:
Thank you for adding this. Let's ask @seglo to review it.

seglo:
@charlibot Thanks for this contribution.  I'm looking forward to having this merged!

> I just tried this with some load, starting one instance then another after a few seconds. I'm getting `ProducerFencedException` on the first instance because the producers in the second instance (with the same `transactional.id`) are being created before the first has finished producing from the revoked partitions. Will have to resolve this before it is ok to merge.

I think the path forward here would be to [listen for revoked partitions](https://doc.akka.io/docs/akka-stream-kafka/current/consumer.html#listening-for-rebalance-events) and somehow trigger a commit from the revoked partitions handler (`TopicPartitionsRevoked`).  Partition assignments as part of a consumer group rebalance won't occur until this handler completes, so that should prevent your second instance from running before the first has resolved the transaction in some way.

seglo:
Great.  I'll do a more thorough review on Monday.

> @seglo thanks for the pointer. I had to use the handler directly from the sub source logic because the rebalance listener approach with actors is using `tell` and so not waiting for the `TopicPartitionsRevoked` code block to complete.

That's interesting.  IMO this is a bug to _not block_ when using these handlers since their primary use case is to gracefully handle a partition being revoked from a group member during a rebalance.  We should either change the rebalance listern to use the `ask` pattern or add a new feature which uses `ask` and waits for an acknowledgement.  Your workaround sounds fine to me for this PR.

From the Kafka consumer javadocs for `ConsumerRebalanceListener`:

> It is guaranteed that all consumer processes will invoke onPartitionsRevoked prior to any process invoking onPartitionsAssigned. So if offsets or other state is saved in the onPartitionsRevoked call it is guaranteed to be saved by the time the process taking over that partition has their onPartitionsAssigned callback called to load the state.

https://kafka.apache.org/20/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html

ennru:
@seglo I agree that the non-blocking rebalance event notification is not that useful. I did not find a good way to do something about it yet, please see https://github.com/akka/alpakka-kafka/issues/539#issuecomment-436944290 for the forces on that...

seglo:
Thanks for the pointer @ennru .  IMO the greatest value of the Partitioned Transactional Source is to easily distribute transactional workloads across consumer groups with more than 1 member.  Therefore it's important to find a way to figure out a way we can synchronously handle the group rebalance.  

Since transactions are involved we could workaround this by using an Akka Streams `RestartSource` which will restart the stream until we stop getting `ProducerFencedExceptions`, but the clean way would be to commit or rollback the transaction, shutdown the subsource, and force the group member the partition was re-assigned to, to wait for its assignment.

I suggest we continue the discussion in #539 and come up with a new API that makes this possible and then leverage that for this contribution.

core/src/main/scala/akka/kafka/javadsl/Transactional.scala
```diff
@@ -46,6 +72,20 @@ object Transactional {
       .mapMaterializedValue(_.toJava)
       .asJava
 
+  /**
+   * Sink that is aware of the [[ConsumerMessage.TransactionalMessage.partitionOffset]] from a [[Transactional.source]].  It will
+   * initialize, begin, produce, and commit the consumer offset as part of a transaction.
+   */
+  def sink[K, V, IN <: Envelope[K, V, ConsumerMessage.PartitionOffset]](
```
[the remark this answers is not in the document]
charlibot: I'm not sure about the producer side of things. I know the `transactional.id` has to be unique for each set of topic-partitions consumed (in this case it will be one `transactional.id` per topic-partition) to keep EOS but didn't realise we need a similar pattern on the sinks side. The blog post does imply that this is the case though but when I first read that I thought it only applied to the consumer side.

My thinking is, if two producers produce to the same `sink-topic-partition` it is down to the consumer to check if the two transactions have been committed with `read_committed` mode. If one of those transactions failed due to a `ProducerFencedException`, those messages would be skipped until reprocessed. 

In kafka streams, it looks like the `TaskId`s (that are used to create the `transactional.id`) are based on the source topics only, implying there is no "producer group". See https://github.com/apache/kafka/blob/046b0087bd76637bbfd813ccef31693fa358ff2d/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java#L556.

core/src/main/scala/akka/kafka/javadsl/Transactional.scala
```diff
@@ -46,6 +72,20 @@ object Transactional {
...
+      settings: ProducerSettings[K, V],
+      transactionalId: String,
+      promise: Promise[Done]
```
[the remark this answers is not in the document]
charlibot: The original sink and flow methods are still present so will use the default defined in scaladsl.

tests/src/test/scala/docs/scaladsl/TransactionsExample.scala
```diff
@@ -101,4 +101,44 @@ class TransactionsExample extends DocsSpecBase(KafkaPorts.ScalaTransactionsExamp
...
+  "Partitioned transactional sink" should "work" in {
+    val consumerSettings = consumerDefaults.withGroupId(createGroupId())
+    val producerSettings = producerDefaults
+    val immutable.Seq(sourceTopic, sinkTopic) = createTopics(1, 2)
+    val transactionalId = createTransactionalId()
+    // #partitionedTransactionalSink
+    val control =
+      Transactional
+        .partitionedSource(consumerSettings, Subscriptions.topics(sourceTopic))
+        .mapAsyncUnordered(8) {
```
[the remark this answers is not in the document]
charlibot: I think I saw it on a `committablePartitionedSource` example. I guess because it is on the source of sources it shouldn't matter in what order the inner block completes

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -28,12 +29,13 @@ private[kafka] final class TransactionalProducerStage[K, V, P](
     val closeTimeout: FiniteDuration,
     val closeProducerOnStop: Boolean,
     val producerProvider: () => Producer[K, V],
-    commitInterval: FiniteDuration
+    commitInterval: FiniteDuration,
+    promise: Promise[Done]
```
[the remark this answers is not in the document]
charlibot: Is `streamCompletePromise` ok?

core/src/main/scala/akka/kafka/internal/TransactionalSubSourceLogic.scala
```diff
@@ -0,0 +1,236 @@
...
+            _.tell(TopicPartitionsRevoked(subscription, revokedTps), sourceActor.ref)
+          }
+          pendingPartitions --= revokedTps
+          partitionsInStartup --= revokedTps
+          implicit val ec: ExecutionContext = materializer.executionContext
+          log.debug(s"Shutting down subsources for: $revokedTps")
+          val futures = revokedTps
+            .flatMap(subSources.get)
+            .map { case (control, streamFuture) => control.shutdown().flatMap(_ => streamFuture) }
+          Await.result(Future.sequence(futures), settings.transactionalStreamStopTimeout)
```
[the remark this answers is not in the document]
charlibot: That's right. We could defer the shutdown until we know the surviving partitions but would involve some more work in `TransactionalProducerStageLogic` by making it lazily init transactions. The issue is, in the `onPartitionsAssigned` callback other instances would have their partitions assigned too and therefore they would have called created the subsources and the corresponding sinks which would `initTransactions`. They would not yet have started consuming until the callback is complete.

We could ask the user to use `Sink.lazyInitAsync` in this case instead of modifying `TransactionalProducerStageLogic`. What do you think?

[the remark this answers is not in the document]
charlibot: The other approach is to skip this promise based stuff and just make sure `ProducerFencedException`s don't kill the entire stream but only the relevant subsource after a rebalance has occurred.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
```diff
@@ -73,6 +73,54 @@ class TransactionsSpec extends SpecBase(kafkaPort = KafkaPorts.TransactionsSpec)
       }
     }
 
+    "complete with partitioned source" in {
+      assertAllStagesStopped {
+        val sourceTopic = createTopicName(1)
```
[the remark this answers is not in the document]
seglo: I see.  Got it. :+1:

core/src/main/scala/akka/kafka/javadsl/Transactional.scala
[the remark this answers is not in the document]
seglo: You're right.  The [Kafka Streams EoS design doc](https://docs.google.com/document/d/1pGZ8xtOOyGwDYgH5vA6h19zOMMaduFK1DAB8_gBYA2c/edit#heading=h.vkrkjfth3p8p) confirms that as well.  I'm not sure how Kafka Streams handles this situation.  Perhaps it's not as critical as I first thought.  I have a question outstanding on the confluent slack.  I'll follow up when I get an answer.

core/src/main/scala/akka/kafka/internal/TransactionalSubSourceLogic.scala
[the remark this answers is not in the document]
charlibot: @seglo @ennru what do you think regarding the above discussion?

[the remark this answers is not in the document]
seglo: I don't see a problem using `Sink.lazyInitAsync`, but would this only solve the issue of the initial rebalance?   On a subsequent rebalance would we have the same problem?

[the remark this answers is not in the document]
charlibot: On second look, the the lazy sink approach isn't going to work. I misunderstood the consumer rebalance listener and thought all `onPartitionsAssigned` will complete before any instance starts consuming but that's not the case (only applies to all `onPartitionsRevoked` executing before any single `onPartitionsAssigned` is executed). Since have to wait until `onPartitionsAssigned` is called before knowing which partitions survive, the revoked partitions would already be consuming from another instance and thus create the `ProducerFencedException` in the original instance.

For now I will have to stick with the original approach of stopping all revoked partitions and letting surviving ones restart. I'll clean things up in next few days and hopefully will be good to go.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
```diff
@@ -73,6 +73,63 @@ class TransactionsSpec extends SpecBase(kafkaPort = KafkaPorts.TransactionsSpec)
...
+            .partitionedSource(consumerSettings, TopicSubscription(Set(sourceTopic), None))
+            .mapAsyncUnordered(maxPartitions) {
+              case (tp, streamCompletePromise, source) =>
+                source
+                  .filterNot(_.record.value() == InitialMsg)
+                  .map { msg =>
+                    ProducerMessage.single(new ProducerRecord[String, String](sinkTopic, msg.record.value),
+                      msg.partitionOffset)
+                  }
+                  .runWith(Transactional.sink(producerDefaults, group + "-" + tp, streamCompletePromise))
```
seglo: Not super important, but it would be nice to find a way to make creating a unique transactional id a little smarter when using the partitioned source.  A convenience function that takes the group and tp?  Group and tp are also present in the offset info passed along to the sink & flow, but it's too late to use it then unless we lazily begin the transaction on the first message that arrives.  We would also need to enforce this behaviour only occurs when using the partitioned source.  WDYT?

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -331,6 +336,21 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
   def withStopTimeout(stopTimeout: java.time.Duration): ConsumerSettings[K, V] =
     copy(stopTimeout = stopTimeout.asScala)
 
+  /**
+   * The timeout waiting for the transactional partitioned streams that close
+   * on every rebalance when revoking the partitions.
+   */
+  def withTransactionalStreamStopTimeout(transactionalStreamStopTimeout: FiniteDuration): ConsumerSettings[K, V] =
```
seglo: I recommend renaming these builder methods replacing `transactional` with `eos` to remain consistent with `ProducerSettings`.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -61,6 +61,7 @@ object ConsumerSettings {
     val pollInterval = config.getDuration("poll-interval").asScala
     val pollTimeout = config.getDuration("poll-timeout").asScala
     val stopTimeout = config.getDuration("stop-timeout").asScala
+    val transactionalStreamStopTimeout = config.getDuration("eos-stream-stop-timeout").asScala
```
seglo: Recommend renaming to `eosStreamStopTimeout` for consistency.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -177,7 +187,7 @@ private abstract class SubSourceLogic[K, V, Msg](
       onRevoke(partitionsToRevoke)
       pendingPartitions --= partitionsToRevoke
       partitionsInStartup --= partitionsToRevoke
-      partitionsToRevoke.flatMap(subSources.get).foreach(_.shutdown())
+      partitionsToRevoke.flatMap(subSources.get).foreach(_._1.shutdown())
```
ennru:
```suggestion
      partitionsToRevoke.flatMap(subSources.get).foreach {
        case (control, ) => control.shutdown()
      }
```

core/src/main/resources/reference.conf
```diff
@@ -44,6 +44,12 @@ akka.kafka.consumer {
   # messages already in the stream (required for successful committing).
   # Prefer use of `DrainingControl` over a large stop-timeout.
   stop-timeout = 30s
+
+  # The timeout waiting for the transactional partitioned streams that close
+  # on every rebalance when revoking the partitions.
+  # This timeout is used in a blocking operation and will suspend consumption
+  # until either it times out, or the partitioned streams complete.
+  eos-stream-stop-timeout = 40s
```
ennru: How does this timeout relate to eg. `session.timeout.ms` and `heartbeat.interval.ms`? 
If the consumer is blocked for the whole timeout, at what values would the broker consider it dead?

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -95,6 +96,15 @@ private abstract class SubSourceLogic[K, V, Msg](
             _.tell(TopicPartitionsRevoked(subscription, revokedTps), sourceActor.ref)
           }
           if (revokedTps.nonEmpty) {
+            if (waitForStreamCompletion) {
+              implicit val ec: ExecutionContext = materializer.executionContext
+              log.debug(s"Shutting down subsources for: $revokedTps")
+              val futures = revokedTps
+                .flatMap(subSources.get)
+                .map { case (control, streamFuture) => control.shutdown().flatMap(_ => streamFuture) }
```
ennru:
```suggestion
                .map { case (control, subSourceCompletion) => control.shutdown().flatMap(_ => subSourceCompletion) }
```

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -95,6 +96,15 @@ private abstract class SubSourceLogic[K, V, Msg](
             _.tell(TopicPartitionsRevoked(subscription, revokedTps), sourceActor.ref)
           }
           if (revokedTps.nonEmpty) {
+            if (waitForStreamCompletion) {
+              implicit val ec: ExecutionContext = materializer.executionContext
+              log.debug(s"Shutting down subsources for: $revokedTps")
+              val futures = revokedTps
```
ennru:
```suggestion
              val completions = revokedTps
```

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
ennru: I wonder if it needs to handle sub-source failures here. (For a horrible example look at `akka.kafka.scaladsl.Consumer.Control.drainAndShutdown`.)

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
```diff
@@ -0,0 +1,82 @@
...
+  override protected def logic(shape: SourceShape[TransactionalMessage[K, V]]): GraphStageLogic with Control =
+    new SingleSourceLogic[K, V, TransactionalMessage[K, V]](shape, txConsumerSettings, subscription)
+    with TransactionalMessageBuilder[K, V] {
+      override def groupId: String = txConsumerSettings.properties(ConsumerConfig.GROUP_ID_CONFIG)
+    }
+}
+
+/** Internal API */
+@InternalApi
+private[kafka] final class TransactionalSubSource[K, V](consumerSettings: ConsumerSettings[K, V],
```
ennru: You may alternatively put `TransactionalSource` and `TransactionalSubSource` into two source files.

core/src/main/scala/akka/kafka/javadsl/Transactional.scala
```diff
@@ -33,6 +36,29 @@ object Transactional {
       .mapMaterializedValue(new ConsumerControlAsJava(_))
       .asJava
 
+  /**
+   * The `partitionedSource` is a way to track automatic partition assignment from kafka.
+   * When a topic-partition is assigned to a consumer, this source will emit tuples with the assigned topic-partition,
+   * a promise that should be passed to the [[Transactional.sink]] or [[Transactional.flow]] and a corresponding
+   * source of `TransactionalMessage`s. Each source is setup for for Exactly Only Once (EoS) kafka message semantics.
+   * To enable EoS it's necessary to use the [[Transactional.sink]] or [[Transactional.flow]] (for passthrough).
+   * When a topic-partition is revoked, the corresponding source completes.
```
ennru:
```suggestion
   * When Kafka rebalances partitions, all sources complete before the remaining sources are issued again.
```

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -30,15 +31,33 @@ object Transactional {
                    subscription: Subscription): Source[TransactionalMessage[K, V], Control] =
     Source.fromGraph(new TransactionalSource[K, V](settings, subscription))
 
+  /**
+   * The `partitionedSource` is a way to track automatic partition assignment from kafka.
+   * When a topic-partition is assigned to a consumer, this source will emit tuples with the assigned topic-partition,
+   * a promise that should be passed to the [[Transactional.sink]] or [[Transactional.flow]] and a corresponding
+   * source of `TransactionalMessage`s. Each source is setup for for Exactly Only Once (EoS) kafka message semantics.
+   * To enable EoS it's necessary to use the [[Transactional.sink]] or [[Transactional.flow]] (for passthrough).
+   * When a topic-partition is revoked, the corresponding source completes.
```
ennru:
```suggestion
   * When Kafka rebalances partitions, all sources complete before the remaining sources are issued again.
```
as in the Java-DSL

core/src/main/resources/reference.conf
charlibot: Looks like these settings do not apply when revoking: `Attempt to heartbeat failed since group is rebalancing`. I'm not sure any timeouts affect the consumer when it's revoking (I tried a thread sleep of 100 seconds to pause the stream expecting `default.api.timeout.ms` (default 60s) to throw a timeout error but that doesn't seem to apply since it rebalanced without any issues even in this case).

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
charlibot: I noticed the `PlainSources` and `CommittableSources` contain both the normal source and subsource so followed the same logic in `TransactionalSources`

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
charlibot: That's a good idea! How would you imagine the API to look like in that case? I had a stab at it earlier, trying to make the producer lazy and create the transactional id based on the first message to send like you said. See: https://github.com/charlibot/alpakka-kafka/compare/transactional-partitioned-source...charlibot:tx-part-src-from-partitioned?expand=1

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
charlibot: What would the ideal outcome be in the case of failure? Currently, the rebalancer logs an exception `failed on partition revocation` but the rebalancing continues and assignment still occurs.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
seglo: That looks pretty good.  I was thinking that a transactional id would still be provided, but it would be appended with GroupTopicPartition info.  One way to differentiate that messages are coming from a partitioned source instead of having `sinkFromPartitioned` and `flowFromPartitioned` would be to mix in a trait with `TransactionalMessage` (i.e. `FromPartitionedSource`) that you could match on in the `TransactionalProducerStageLogic`.

core/src/main/resources/reference.conf
seglo: I don't see anything obvious that aborts the call to the revoke handler.  It seems to be called synchronously as part of the poll, which makes sense since the `ConsumerRebalanceListener` javadocs says these callbacks are made as part of the user poll thread.  There's a timer with an infinite (`Long.MAX`) duration that is started before it's invoked.

seglo: It looks like the Consumer Group Coordinator process on the broker determines the timeout.  `group.max.session.timeout.ms` has a default of 300 seconds.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
seglo: Ah, yes of course.  That looks like a good solution.  Could use a trait on the PartitionOffset type too.  Either way.

charlibot: Do you think I should merge it into this PR or open a new one with those changes after this has been merged?

seglo: Either is fine with me.  @ennru WDYT?

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -95,6 +96,17 @@ private abstract class SubSourceLogic[K, V, Msg](
...
+              if (log.isDebugEnabled) {
+                log.debug(s"Shutting down subsources for: $revokedTps")
+              }
+              val completions = revokedTps
+                .flatMap(subSources.get)
+                .map { case (control, subSourceCompletion) => control.shutdown().flatMap(_ => subSourceCompletion) }
+              Await.result(Future.sequence(completions), settings.eosStreamStopTimeout)
+              log.debug(s"Finished shutting down subsources")
+            }
             partitionRevokedCB.invoke(revokedTps)
```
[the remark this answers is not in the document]
charlibot: I wasn't sure if it was still needed. I figured it doesn't hurt since the `onRevoke` is called and `pendingPartitions` etc has the `partitionsToRevoke` removed. 

I think that PR would be super helpful in tidying this class up and could help https://github.com/akka/alpakka-kafka/issues/539.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -95,6 +96,17 @@ private abstract class SubSourceLogic[K, V, Msg](
             _.tell(TopicPartitionsRevoked(subscription, revokedTps), sourceActor.ref)
           }
           if (revokedTps.nonEmpty) {
+            if (waitForStreamCompletion) {
+              implicit val ec: ExecutionContext = materializer.executionContext
+              if (log.isDebugEnabled) {
+                log.debug(s"Shutting down subsources for: $revokedTps")
+              }
+              val completions = revokedTps
```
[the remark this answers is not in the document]
charlibot: In `performShutdown` can already `Await` the `control.shutdown()` on all the subsources. Could also use the promise if `waitForStreamCompletion` is true. Something like:
```suggestion
    val completions = if (waitForStreamCompletion) {
      subSources.mapValues {
        case (control, streamCompleteFuture) => control.drainAndShutdown(streamCompleteFuture)
      }
    } else {
      subSources.mapValues {
        case (control, streamCompleteFuture) => control.shutdown()
      }
    }
    Await.result(Future.sequence(completions), 40 seconds)
```

ennru:
Thank you @charlibot and @seglo for your work with this.
We discussed it and will work more thoroughly with it, but I already now want to let you know that this won't be merged into Alpakka Kafka 1.0. It will have to wait until 1.1 as we are in the release-candidate phase already and this touches quite a few things.

ennru:
Sorry, I didn't have the time to look at your new approach, yet. Will do soon.

ennru:
@charlibot I've finally got around to look into your proposal again. We've quite some digging going on as some have discovered duplicates in transactional source/sink flows #756 and #758. 
Everything circles around rebalancing #755. So @szymonm and I have simplified the rebalance listener and the `ListenerCallbacks`.
That introduced some conflict to your suggestions, would you like to rebase those?

ennru:
Thank you.
In your `remove-async-callbacks` branch the (now non-async) callbacks touch state in the stage concurrently from the actor's thread which called poll. That is not allowed.
I'm making some progress on a new take for the rebalancing callbacks which will block, but that is still a long way to go.

ennru:
I've just opened #761 which explores synchronous callbacks for rebalancing.

ennru:
The work on the transactional flows hasn't come to an end, yet. We'll revisit your PR once that is done -- hopefully very soon.

ennru:
I see you merged `master` into this branch. I'm afraid that could give us trouble when merging it back to `master`. A rebase would be preferred.

brendan-mcadams-exa:
Any chance of getting this PR finished and approved? The functionality would be extremely useful.

seglo:
I think this PR can be revived now that there's an internal implementation to handle revoked partitions synchronously from #761

ennru:
@brendan-mcadams-exa Thank you for pinging this suggested feature. 
Could you explain your use-case a bit? I wonder if parts of what this kind of source offers would be easier solved with a proper rebalance handler which I have been investigating.

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -51,24 +67,29 @@ object Transactional {
   ): Flow[Envelope[K, V, ConsumerMessage.PartitionOffset], Results[K, V, ConsumerMessage.PartitionOffset], NotUsed] = {
     require(transactionalId != null && transactionalId.length > 0, "You must define a Transactional id.")
 
-    val txSettings = settings.withProperties(
-      ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG -> true.toString,
-      ProducerConfig.TRANSACTIONAL_ID_CONFIG -> transactionalId,
-      ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION -> 1.toString
-    )
+    val txSettingsToProducer = (txid: String) => {
```
raboof: So this creates a producer (with a unique `transactionalId`) for each partition that is assigned to this node. Isn't that expensive, or is the Kafka client smart about pooling the resources for those multiple producers?

seglo: @raboof The original idea was to create a producer per partition, but you're right that it's possible that that could have unintended performance problems for large partition subscriptions.  AFAIK the separate instances of `KafkaProducer` don't share a common resource pool, so they would be managed separately.  We don't appear to have a precedent in our docs for using producers per substream with other partitioned sources.

I looked at Kafka Streams impl and the original design doc (https://docs.google.com/document/d/1pGZ8xtOOyGwDYgH5vA6h19zOMMaduFK1DAB8_gBYA2c/edit#heading=h.mki3gltx1zw) and they create 1 `KafkaProducer` per `Task`.  Each `TaskId` (1:1 with a `Task`) appears to represent one assigned partition.  The Kafka Streams `Task` `transactional.id` for a producer is set with a pattern of the application id (defined by the user across all instances), and a `TaskId`.  So my initial understanding is that it will indeed create one producer per partition.  I think we need to study this implementation more closely to understand if we need to be smarter about the number of producers we spawn.

raboof:
I started on rebasing this PR - it now compiles but there are still tests failing (https://travis-ci.org/akka/alpakka-kafka/builds/593117356)

seglo:
Thanks @raboof .  I've started reviewing your branch.  I'll follow up in the next week.

core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -54,15 +56,17 @@ private[kafka] trait TransactionalMessageBuilder[K, V]
...
-      ),
-      offset = rec.offset,
-      committedMarker
+    val groupTopicPartition = GroupTopicPartition(
+      groupId = groupId,
+      topic = rec.topic,
+      partition = rec.partition
     )
+    val offset = if (fromPartitionedSource) {
+      new PartitionOffsetCommittedMarker(groupTopicPartition, offset = rec.offset, committedMarker)
```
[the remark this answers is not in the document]
ennru: No, it will be just one class created at compile time. Just a new instance of it per element. But it should possibly define a proper class.

seglo:
I've created a new PR #930 to continue this work with the goal of including it in Alpakka Kafka 2.0.0.

tests/src/test/scala/docs/scaladsl/TransactionsExample.scala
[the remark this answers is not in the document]
seglo: Yeah, this is fine.

core/src/main/resources/reference.conf
seglo: Config was removed.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
seglo: I'll create a test for this.

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
seglo: > I noticed the `PlainSources` and `CommittableSources` contain both the normal source and subsource so followed the same logic in `TransactionalSources`

This is still the case after rebase.  I think it should follow the same convention as `PlainSources` and `CommittableSources`, unless we want to re-factor those as well.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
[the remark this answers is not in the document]
seglo: N/A after the rebase AFAICT..

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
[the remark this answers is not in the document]
seglo: N/A.  Now using draining logic.

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
seglo: I took a look at this again and it does look like a producer is created per task.  I'm less clear on whether a task is 1:1 with a partition, but since you can only define one partition per task I don't see how it could be anything else.

This understanding conflicts with the original design doc, so I posed the question to the kafka-users mailing list.  Hopefully we'll get an authoritative reply.

https://lists.apache.org/thread.html/491b89840776498009fa773a737bf512a2c144558cce961b4ff8c768@%3Cusers.kafka.apache.org%3E

PR #710, opened 2019-01-29
Generalize API for ProduceSettings

bmaggi:
* the producerFactory required by ProducerSettings should require
to create a Producer and not the implementation
* rename method createKafkaProducer in createProducer

# Pull Request Checklist

* [ ] Have you added tests for any changed functionality?

## Purpose

Use an interface instead of an implementation to easily override the default KafkaProducer.

## Background Context

I was trying to override the default KafkaProducer, it's the best solution I found to inject a personal Producer (Is there an other recommended way?)

## References

Ennru suggested to open a PR  in this [discussion ](https://discuss.lightbend.com/t/inject-a-producer-using-producersettings/3322)

### DIFF
--- a/core/src/main/scala/akka/kafka/ProducerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ProducerSettings.scala
@@ -145,9 +145,10 @@ object ProducerSettings {
   /**
    * Create a [[org.apache.kafka.clients.producer.KafkaProducer KafkaProducer]] instance from the settings.
    */
-  def createKafkaProducer[K, V](settings: ProducerSettings[K, V]): KafkaProducer[K, V] = {
-    new KafkaProducer[K, V](settings.getProperties, settings.keySerializerOpt.orNull, settings.valueSerializerOpt.orNull)
-  }
+  def createKafkaProducer[K, V](settings: ProducerSettings[K, V]): KafkaProducer[K, V] =
+    new KafkaProducer[K, V](settings.getProperties,
+                            settings.keySerializerOpt.orNull,
+                            settings.valueSerializerOpt.orNull)
 
 }
 
@@ -275,8 +276,8 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
   ): ProducerSettings[K, V] = copy(producerFactory = factory)
 
   /**
-    * Get the Kafka producer settings as map.
-    */
+   * Get the Kafka producer settings as map.
+   */
   def getProperties: java.util.Map[String, AnyRef] = properties.asInstanceOf[Map[String, AnyRef]].asJava
 
   private def copy(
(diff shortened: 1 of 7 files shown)


bmaggi:
The failure is indicating an API break, which is true since I renamed createKafkaProducer in createProducer. As far as I understand this method is internal API and can be changed.

core/src/main/scala/akka/kafka/ProducerSettings.scala
```diff
@@ -273,7 +273,7 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
    * Replaces the default Kafka producer creation logic.
    */
   @InternalApi private[kafka] def withProducerFactory(
```
ennru: As @charlibot suggests, you may remove the Internal API flags to open up for other use cases.

core/src/main/scala/akka/kafka/ProducerSettings.scala
```diff
@@ -307,7 +307,7 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
     ")"
 
   /**
-   * Create a `KafkaProducer` instance from the settings.
+   * Create a `Producer` instance from the settings.
    */
-  def createKafkaProducer(): KafkaProducer[K, V] = producerFactory.apply(this)
+  def createProducer(): Producer[K, V] = producerFactory.apply(this)
```
ennru: We have many things called "Producer", the method name could be kept as `createKafkaProducer` and with `Producer` as return type.

bmaggi: I changed the name back.

core/src/main/scala/akka/kafka/ProducerSettings.scala
bmaggi: I removed the tag internal API

bmaggi:
I pushed a new commit with the suggested changes.

core/src/main/scala/akka/kafka/ProducerSettings.scala
```diff
@@ -269,11 +269,10 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
     copy(eosCommitInterval = eosCommitInterval.asScala)
 
   /**
-   * Internal API.
    * Replaces the default Kafka producer creation logic.
    */
-  @InternalApi private[kafka] def withProducerFactory(
-      factory: ProducerSettings[K, V] => KafkaProducer[K, V]
+  private[kafka] def withProducerFactory(
```
ennru: Please remove the `private[kafka]` as well.

bmaggi:
I implemented your suggestions. 

Please note that the same changes could/should probably be done to ConsumerSettings

ennru:
The code formatting check failed, please run `Test/compile` and push the changes.

MiMa might fail as well. Run `sbt mimaReportBinaryIssues` to check it and add a `1.0-RC1.backwards.excludes` file in `core/src/main/mima-filters` with the reported excludes.

bmaggi:
An integration test "CMD="dockerComposeTestAll"" is still failing but i don't know how this is related to this PR. 

Logs:
[info] PartitionedSourceFailoverSpec:
[info] partitioned source
[info] - should not lose any messages when a Kafka node dies *** FAILED *** (33 seconds, 430 milliseconds)
[info]   7500 was not equal to 10000 (PartitionedSourceFailoverSpec.scala:100)

PR #713, opened 2019-02-01
Add production considerations

ennru:
## Purpose

Complement the documentation with things to consider when preparing production use.

### DIFF
--- a/docs/src/main/paradox/production.md
+++ b/docs/src/main/paradox/production.md
@@ -1,0 +1,43 @@
+# Production considerations
+
+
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/production.md
```diff
@@ -0,0 +1,43 @@
+# Production considerations
+
+
+## Alpakka Kafka API
+
+1. Do not use `Consumer.atMostOnceSource` in production
```
2m: We should also add the rationale for these tips.

PR #727, opened 2019-02-21
Fix paradox generation command

2m:
## Fixes

Paradox generation needs Configuration as this is where all of the settings are set.


2m:
Superseded by https://github.com/akka/alpakka-kafka/pull/726

PR #737, opened 2019-03-04
Release notes for 1.0.1

ennru:


docs/src/main/paradox/release-notes/1.1.md
```diff
@@ -0,0 +1,23 @@
+# Alpakka Kafka 1.1
+
+## Alpakka Kafka 1.1 to upgrade the Apache Kafka client to 2.1.1
```
patriknw: Why isn't it a patch version? 1.0.1? Update of the Kafka client from 2.1.0 to 2.1.1 seems (from the version number) to be a compatible update.

ennru: Changed to 1.0.1 now.

docs/src/main/paradox/release-notes/1.0.md
```diff
@@ -1,8 +1,31 @@
 # Alpakka Kafka 1.0
```
2m: Should we call the page and the title `1.0.x` like it is is akka-http?

We could add a redirect from the current `1.0.md` to the new `1.0.x.md`.

PR #739, opened 2019-03-04
Testcase to illustrate duplicated transactional messages when rebalancing

szymonm:
In the test below I'm creating many transactional streams that process copy elements from input to output.

The test fails because some elements are copied multiple times.


ennru:
Thank you for trying to isolate this into test cases.

ennru:
I assume this is superceded by #742?

szymonm:
Correct, @ennru

PR #742, opened 2019-03-08
Fix transaction offsets for transactional producer

szymonm:
# Pull Request Checklist

* [x] Have you added tests for any changed functionality?

## Purpose

Callbacks from `producer.send` are guaranteed to execute in order only for
a single output partition. That means we can have a race condition where we
execute a callback for input record with offset 1 before executing a
callback for input record with offset 0. That causes
`NonEmptyTransactionBatch` to contain offset 0, when committing
transaction. That leads to data duplication.

## Background Context

Use maximal offset when committing transaction in
TransactionalProducerStage

This fix ensures that we only increase the offsets stored in
`TransactionBatch`. Since having all consecutive offsets wrote to Kafka is
guaranteed by `awaitingConfirmation == 0`, we can only keep the maximal
offset in the `TransactionBatch`.

## References

See #539 for some relevant context.

Are there any relevant issues / PRs / mailing lists discussions?

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
@@ -54,7 +54,12 @@ private object TransactionalProducerStage {
   final class NonemptyTransactionBatch(head: PartitionOffset,
                                        tail: Map[GroupTopicPartition, Long] = Map[GroupTopicPartition, Long]())
       extends TransactionBatch {
-    private val offsets = tail + (head.key -> head.offset)
+    // There is no guarantee that offsets adding callbacks will be called in any particular order.
+    // Decreasing an offset stored for the KTP would mean possible data duplication.
+    // Since `awaitingConfirmation` counter guarantees that all writes finished, we can safely assume
+    // that all all data up to maximal offsets has been wrote to Kafka.
+    private val previousHighest = tail.getOrElse(head.key, -1L)
+    private val offsets = tail + (head.key -> head.offset.max(previousHighest))
 
     def group: String = head.key.groupId
     def offsetMap(): Map[TopicPartition, OffsetAndMetadata] = offsets.map {
--- a/tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
+++ b/tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
@@ -5,6 +5,7 @@
 
 package akka.kafka.scaladsl
 
+import akka.Done
 import akka.kafka.ConsumerMessage.PartitionOffset
 import akka.kafka.Subscriptions.TopicSubscription
 import akka.kafka._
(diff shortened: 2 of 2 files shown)


szymonm:
I know of failing tests... will fix them.

ennru:
Thank you for digging into this down to the bottom -- great find!

seglo:
Good work @szymonm.  Thank you.

I'm curious how you can recreate this reliably in a test without some mechanism to force one callback to return before the other?

szymonm:
@ennru I cannot easily force callbacks to return in specific other. 

However, the probability of the one with the highest order to return last is `1/number of ktps per topic`, so after increasing the number of partitions and the number of consumers I can get pretty good reproducibility of the failure. What do you think of this approach?

seglo:
> @ennru I cannot easily force callbacks to return in specific other.
> 
> However, the probability of the one with the highest order to return last is `1/number of ktps per topic`, so after increasing the number of partitions and the number of consumers I can get pretty good reproducibility of the failure. What do you think of this approach?

@szymonm I assume this question was directed to me?

I think that's probably the best we can do right now without modifying the `TransactionalProducerStage` just for the sake of testing this.  Instead of adding another test, how about we increase the number of partitions for all the tests running in `TransactionsSpec`?  I think this is good enough because if it were already done it would have revealed this issue as transient failures.

szymonm:
You're right @seglo.

We need to have at least two consumers to test that offsets are stored in Kafka correctly (more if we want the test to be more reliable in finding errors). 
We have that in tests 3rd and 4th, so that should work. However, IIUC these tests focus on failure recovery. I would argue that having specialised test for offset committing in case of multiple consumers makes sense.

seglo:
@szymonm I'm not sure I get it.

I'm attempting to understand what's happening in your test provided with this PR.  It seems that it sequentially runs transactional streams for 5 batches of numbers.   i.e.)

```
1..10
11..20
21..30
31..40
41..50
```

The source test data used in each transaction is always coming from a single partition, but is produced to multiple partitions using the round robin partioning strategy because a key isn't defined in the `ProducerRecord`.  Is the idea that one of these batches could potentially commit an earlier offset back to the group, and that the next next transaction may read a duplicate already processed message?

szymonm:
That's right, @seglo .

One thing I would add is that we don't impose any order in which the streams process data (the `runStream` method is non-blocking). It's on Kafka to assign ktps to streams and to ensure that at most one is processing the data at the time. Each stream processes 10 elements and then finishes, so Kafka will assign the ktp with elements to another stream.

szymonm:
FYI:
* fixed the tests so only the last one writes to multiple partitions,
* deleted accidental changes to logging config,
* refactored the tests in TransactionalSpec to avoid code duplication.

Pls review commit by commit.

seglo:
@szymonm Got it, and thanks for DRYing up the tests.  Would you mind adding some comments to your new test so it's more clear to others (and myself in 6 month) what its purpose is?

ennru:
Failure was #745

ennru:
Thank you for digging this deep and fixing it!

PR #744, opened 2019-03-08
eliminate scala compiler warnings

fancywriter:
I believe `private[kafka]` was intended instead of just `private` and few deprecations warnings (by kafka-clients poll) were removed.


core/src/main/scala/akka/kafka/internal/ProducerStage.scala
```diff
@@ -36,7 +36,7 @@ private[kafka] trait ProducerStage[K, V, P, IN <: Envelope[K, V, P], OUT <: Resu
  * INTERNAL API
  */
 @InternalApi
-private object ProducerStage {
+private[kafka] object ProducerStage {
```
ennru: Can you change it to `[internal]`, please.

ennru:
Failure was #745

ennru:
Thank you for fixing those.

Issue #750, opened 2019-03-16
Increasing consumers in consumer group cause rebalance failure with CommitFailedException because of revoked partition


Maatary:
Hi, 

I think this issue is related to #539 but I don't know if it is a bug, or the user is supposed to handle it himself. 

So I have a consumer group, whenever i increase the number of consumer in that group, the revoking of partition is causing the following error: 

```
org.apache.kafka.clients.consumer.CommitFailedException: Commit cannot be completed since the group has already rebalanced and assigned the partitions to another member. This means that the time between subsequent calls to poll() was longer than the configured max.poll.interval.ms, which typically implies that the poll loop is spending too much time message processing. You can address this either by increasing max.poll.interval.ms or by reducing the maximum size of batches returned in poll() with max.poll.records.
	at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.sendOffsetCommitRequest(ConsumerCoordinator.java:778)
	at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.doCommitOffsetsAsync(ConsumerCoordinator.java:617)
	at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.commitOffsetsAsync(ConsumerCoordinator.java:584)
	at org.apache.kafka.clients.consumer.KafkaConsumer.commitAsync(KafkaConsumer.java:1479)
	at akka.kafka.internal.KafkaConsumerActor.akka$kafka$internal$KafkaConsumerActor$$commit(KafkaConsumerActor.scala:430)
	at akka.kafka.internal.KafkaConsumerActor$$anonfun$receive$1.applyOrElse(KafkaConsumerActor.scala:210)
	at akka.actor.Actor.aroundReceive(Actor.scala:517)
	at akka.actor.Actor.aroundReceive$(Actor.scala:515)
	at akka.kafka.internal.KafkaConsumerActor.akka$actor$Timers$$super$aroundReceive(KafkaConsumerActor.scala:142)
	at akka.actor.Timers.aroundReceive(Timers.scala:55)
	at akka.actor.Timers.aroundReceive$(Timers.scala:40)
	at akka.kafka.internal.KafkaConsumerActor.aroundReceive(KafkaConsumerActor.scala:142)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:592)
	at akka.actor.ActorCell.invoke(ActorCell.scala:561)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:258)
[...]

PR #751, opened 2019-03-20
Stash `Commit`s during rebalance

ennru:
## Purpose

By tracking "unfinished" rebalances which are initiated with an `onParitionsRevoked` during `poll`, but not finished with `onParitionsAssigned`. The `KafkaConsumerActor` can collect incoming commits which would fail with `CommitFailedException`.

The commits are issued once the rebalance finishes and partitions get assigned. 

## References

See #750

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -31,7 +31,6 @@ import org.apache.kafka.clients.consumer._
 import org.apache.kafka.common.{Metric, MetricName, TopicPartition}
 
 import scala.collection.JavaConverters._
-import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
 import scala.util.Try
 import scala.util.control.NonFatal
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -17,7 +17,6 @@ import akka.actor.{
   ActorRef,
   DeadLetterSuppression,
   NoSerializationVerificationNeeded,
-  Stash,
   Status,
   Terminated,
   Timers
@@ -83,11 +82,6 @@ import scala.util.control.NonFatal
         with NoSerializationVerificationNeeded
 
     private[KafkaConsumerActor] case object PollTask
-
-    /**
-     * Sent on timer when rebalance is in progress after polling.
-     */
-    private[KafkaConsumerActor] case object RebalanceUnfinished
 
     private val number = new AtomicInteger()
     def nextNumber(): Int =
@@ -224,7 +218,6 @@ import scala.util.control.NonFatal
                                                                  settings: ConsumerSettings[K, V])
     extends Actor
     with ActorLogging
-    with Stash
     with Timers {
   import KafkaConsumerActor.Internal._
   import KafkaConsumerActor._
@@ -248,10 +241,21 @@ import scala.util.control.NonFatal
   private var stopInProgress = false
 
   /**
-   * While `true`, all [[Commit]] messages are stashed.
+   * While `true`, committing is delayed.
    * Changed by `onPartitionsRevoked` and `onPartitionsAssigned` in [[WrappedAutoPausedListener]].
    */
   private var rebalanceInProgress = false
+
+  /**
+   * Keeps commit offsets during rebalances for later commit.
+   */
+  private var rebalanceCommitStash = Map.empty[TopicPartition, OffsetAndMetadata]
+
+  /**
+   * Keeps commit senders that need a reply once stashed commits are made.
+   */
+  private var rebalanceCommitSenders = Vector.empty[ActorRef]
+
   private var delayedPollInFlight = false
 
   def receive: Receive = LoggingReceive {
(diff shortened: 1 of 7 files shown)


core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -224,6 +224,23 @@ import scala.util.control.NonFatal
   private var commitsInProgress = 0
```
ChanglinZhou: this member should be volatile, because it will be update in commitAsync callback.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -468,11 +492,10 @@ import scala.util.control.NonFatal
             log.warning("Kafka commit took longer than `commit-time-warning`: {} ms", duration / 1000000L)
           }
           commitsInProgress -= 1
-          if (exception != null) reply ! Status.Failure(exception)
+          if (exception != null) sendReply(Status.Failure(exception))
           else {
-            val committed = Committed(offsets.asScala.toMap)
-            self ! committed
-            reply ! committed
+            self ! Committed(offsets.asScala.toMap)
```
ChanglinZhou: // 1. commitAsync callback may not be called at the same thread
    // 2. self is stored in the thread local of object ActorCell
    // so, use self in async callback is wrong. use local variable to save the self ActorRef

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
ennru: Looks, like it, but that is not required as it will always be called via `poll()` which is the actor's thread:
```
/**
 * A callback interface that the user can implement to trigger custom actions when a commit request completes. The callback
 * may be executed in any thread calling {@link Consumer#poll(java.time.Duration) poll()}.
 */
public interface OffsetCommitCallback {
```

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
ennru: Thank you for divining into it. Same as above applies here. I should add a comment for that.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -453,7 +477,7 @@ import scala.util.control.NonFatal
     commitRefreshing.updateRefreshDeadlines(assignedOffsets.keySet)
   }
 
-  private def commit(commitMap: Map[TopicPartition, OffsetAndMetadata], reply: ActorRef): Unit = {
+  private def commit(commitMap: Map[TopicPartition, OffsetAndMetadata], sendReply: AnyRef => Unit): Unit = {
     commitRefreshing.updateRefreshDeadlines(commitMap.keySet)
     commitsInProgress += 1
     val startTime = System.nanoTime()
```
ChanglinZhou: What if commitAsync throws exception?

ennru: That exception is passed to the call-back as parameter.

ennru:
With the positive experience of the very similar (if not copied) implementation in [fs2-kafka](https://ovotech.github.io/fs2-kafka/) I'm positive this change is good for Alpakka Kafka, as well.

See https://github.com/ovotech/fs2-kafka/pull/117#issuecomment-480738102

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -605,12 +640,14 @@ import scala.util.control.NonFatal
       val tps = partitions.asScala.toSet
       commitRefreshing.assignedPositions(tps, consumer, positionTimeout)
       listener.onAssign(tps)
+      rebalanceInProgress = false
```
2m: Even without any partitions assigned, this will be called with an empty collection, right? That will make it sure that actor does not get stuck in rebalance-in-progress state.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -473,11 +497,10 @@ import scala.util.control.NonFatal
...
           }
           commitsInProgress -= 1
-          if (exception != null) reply ! Status.Failure(exception)
+          if (exception != null) sendReply(Status.Failure(exception))
           else {
-            val committed = Committed(offsets.asScala.toMap)
-            self ! committed
-            reply ! committed
+            self ! Committed(offsets.asScala.toMap)
+            sendReply(Done)
```
2m: Before the change senders would get back the `Committed` message as a reply. But now `Done` is sent instead? Is that deliberate?

ennru: Very observant.
Yes, I noticed the other side doesn't care about the `Committed`

```scala
    consumerActor
      .ask(Commit(offsetsMap))(Timeout(commitTimeout))
      .map(_ => Done)
```

https://github.com/akka/alpakka-kafka/blob/docs/v1.0.1/core/src/main/scala/akka/kafka/internal/CommittableSources.scala#L114-L116

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
ennru: Yes.

PR #752, opened 2019-03-20
Transactions: Add offsets to be committed directly after `producer.send`

2m:
## Purpose

The transaction flow could produce duplicates (as the test cases showed). By adding the offsets to be committed directly after `producers.send` instead of in the call-back this is solved.

## Background context

The included test-case where multiple transactional source-sink flows are being restarted while moving messages from one topic to another. This test-case showed produced duplicates before the change.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
@@ -118,10 +118,13 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
   }
 
   private def suspendDemand(): Unit =
-    setHandler(stage.out, new OutHandler {
-      // suspend demand while a commit is in process so we can drain any outstanding message acknowledgements
-      override def onPull(): Unit = ()
-    })
+    setHandler(
+      stage.out,
+      new OutHandler {
+        // suspend demand while a commit is in process so we can drain any outstanding message acknowledgements
+        override def onPull(): Unit = ()
+      }
+    )
 
   override protected def onTimer(timerKey: Any): Unit =
     if (timerKey == commitSchedulerKey) {
@@ -141,11 +144,10 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
     }
   }
 
-  override val onMessageAckCb: AsyncCallback[Envelope[K, V, P]] =
-    getAsyncCallback[Envelope[K, V, P]](_.passThrough match {
-      case o: ConsumerMessage.PartitionOffset => batchOffsets = batchOffsets.updated(o)
-      case _ =>
-    })
+  override def postSend(msg: Envelope[K, V, P]): Unit = msg.passThrough match {
+    case o: ConsumerMessage.PartitionOffset => batchOffsets = batchOffsets.updated(o)
+    case _ =>
+  }
 
   override def onCompletionSuccess(): Unit = {
     log.debug("Committing final transaction before shutdown")
--- a/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
@@ -187,7 +187,6 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
   }
 
   private def abortTransaction(): Unit = {
-    println("aborting transaction")
     log.debug("Aborting transaction")
     producer.abortTransaction()
   }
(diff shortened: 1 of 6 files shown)


2m:
The restarted transactional producers should be fine. They provide transactional id which then is used to retrieve the producer id. This, together with calling `producer.initTransactions()` on the [producer stage startup](https://github.com/akka/alpakka-kafka/blob/a8a8268deb40be17a1265378e514d2534ab00097/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala#L104) make sure that existing transaction is aborted, if any. Or if there is a transaction being committed, then it waits until the commit succeeds.

Still investigating the root cause of this. Currently looking into the effects of rebalancing to the transactional consumer, where the duplicates are coming from.

2m:
Added a commit where consumer actor is only initialized when the consumer stage gets demand. This makes sure that `initTransactions()` from the `TransactionalProducerStageLogic.preStart()` is called before initializing the consumer. `initTransactions()` aborts transactions that have been started by another reincarnation of producer with the same transaction-id (producer-id).

However I still noticed some duplicates (but less) after this change.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
```diff
@@ -280,9 +381,13 @@ class TransactionsSpec extends SpecBase(kafkaPort = KafkaPorts.TransactionsSpec)
 
   private def valuesProbeConsumer(settings: ConsumerSettings[String, String],
                                   topic: String): TestSubscriber.Probe[String] =
+    valuesSource(settings, topic)
+      .runWith(TestSink.probe)
+
+  private def valuesSource(settings: ConsumerSettings[String, String],
```
seglo: Does the probe consumer properties enable `isolation.level` `read_committed`?  This will ensure that only successfully committed messages are consumed.

2m: Yes, the consumer has that in its settings:

https://github.com/akka/alpakka-kafka/blob/a93b8b70848395790e76920ab112526cd4156b16/tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala#L380

2m:
Here is another commit that builds onto this, where it checks for duplicate messages in the producer. This would catch any duplicates arriving to the same instance of producer. However there does not seem to be any when running the test-case, meaning that duplication occurs between different instances of the producer. 

https://github.com/2m/alpakka-kafka/commit/bf1ff4438c9af39854c86b38383ff7ba847c8b12

2m:
@seglo @ennru The offset update change fixes this test-case for me and I do not see any duplicates anymore. I have removed the initialization order fix as that seems to not be needed now.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -188,6 +187,7 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
   }
 
   private def abortTransaction(): Unit = {
+    println("aborting transaction")
```
ennru: Remove.

2m:
I also published jars with this fix for anymore that would like to try this out: `1.0.1+11-d381b4f4`

You will need to add Alpakka Kafka snapshot repository to your build to resolve it: https://doc.akka.io/docs/alpakka-kafka/current/snapshots.html

//cc @JustinPihony

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -144,11 +144,10 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
     }
   }
 
-  override val onMessageAckCb: AsyncCallback[Envelope[K, V, P]] =
-    getAsyncCallback[Envelope[K, V, P]](_.passThrough match {
-      case o: ConsumerMessage.PartitionOffset => batchOffsets = batchOffsets.updated(o)
-      case _ =>
-    })
+  override def postSend(msg: Envelope[K, V, P]): Unit = msg.passThrough match {
```
ennru: Looks like the `onMessageAckCb` could be removed altogether.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
```diff
@@ -257,6 +261,103 @@ class TransactionsSpec extends SpecBase(kafkaPort = KafkaPorts.TransactionsSpec)
...
+      val sourcePartitions = 10
+      val destinationPartitions = 4
+      val consumers = 3
+
+      val sourceTopic = createTopic(1, sourcePartitions)
+      val sinkTopic = createTopic(2, destinationPartitions)
+      val group = createGroupId(1)
+
+      givenInitializedTopic(sourceTopic)
+      givenInitializedTopic(sinkTopic)
```
ennru: These are not needed when using `createTopic`.

2m:
Addressed review feedback.

Issue #756, opened 2019-03-22
Transactional source and sink duplicates messages when restarted


2m:
### Versions used 

Alpakka Kafka version: master

### Expected Behavior

No duplicate messages are in the destination topic.

### Actual Behavior

When running the test-case from #752 where three transactional source and sink streams are being periodically restarted, after 1M of messages there are usually up to 50 duplicated messages in the destination topic.

2m:
Here are some of the findings I have discovered while working on this:

* I see duplicate messages in the transactional flow, when it is being restarted. However when the new incarnation of the transactional flow comes alive, the producer calls `producer.initTransactions()` which aborts transactions from the old incarnation of the transactional producer. So those messages are not committed in the initial transaction, but after restart.

* It could be possible that some messages are duplicated when consuming from the destination topic when counting messages and looking duplicates. However I refactored the testcase so transactional processing and then counting for duplicates happen one after another, so consumer is not rebalanced wile counting.

PR #757, opened 2019-03-25
Implement draining logic for transactional flow

szymonm:
# Pull Request Checklist

* [x] Have you added tests for any changed functionality?

## Fixes

Duplications in a transactional flow that occur, when either closing the stream or partitions are revoked by Kafka.

## Purpose

This PR introduces draining logic for `TransactionalSource`. Specifically, for each partition managed by the source, we keep track of the message offsets emitted by the source.
When closing the stage or revoking partitions, we make sure that all the offsets are committed back to the Kafka by the producer.
To know that an offset was committed by the producer we attach `CommittedMarker` to each message produced by the stream. `TransactionalProducerStage` uses this marker to tell the source that an offset was committed.

## Background Context

In a transactional stream processing partition `P`, we need to wait for all commits to be acknowledged by Kafka before closing the consumer.
If we don't, it's possible that another consumer is assigned `P`, which triggers fetching offset that can be stale or overwritten later. This causes data duplication.

## References

https://github.com/akka/alpakka-kafka/issues/758
https://github.com/akka/alpakka-kafka/issues/756

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -2,14 +2,14 @@
 # // #producer-settings
 # Properties for akka.kafka.ProducerSettings can be
 # defined in this section or a configuration section with
-# the same layout. 
+# the same layout.
 akka.kafka.producer {
   # Tuning parameter of how many sends that can run in parallel.
   parallelism = 100
 
   # Duration to wait for `KafkaConsumer.close` to finish.
   close-timeout = 60s
-  
+
   # Fully qualified config path which holds the dispatcher configuration
   # to be used by the producer stages. Some blocking may occur.
   # When this value is empty, the dispatcher configured for the stream
@@ -17,6 +17,7 @@ akka.kafka.producer {
   use-dispatcher = "akka.kafka.default-dispatcher"
 
   # The time interval to commit a transaction when using the `Transactional.sink` or `Transactional.flow`
+  # for exactly-once-semantics processing.
   eos-commit-interval = 100ms
 
   # Properties defined by org.apache.kafka.clients.producer.ProducerConfig
@@ -29,27 +30,29 @@ akka.kafka.producer {
 # // #consumer-settings
 # Properties for akka.kafka.ConsumerSettings can be
 # defined in this section or a configuration section with
-# the same layout. 
+# the same layout.
 akka.kafka.consumer {
   # Tuning property of scheduled polls.
   # Controls the interval from one scheduled poll to the next.
   poll-interval = 50ms
-  
+
   # Tuning property of the `KafkaConsumer.poll` parameter.
   # Note that non-zero value means that the thread that
   # is executing the stage will be blocked. See also the `wakup-timeout` setting below.
   poll-timeout = 50ms
-  
+
   # The stage will delay stopping the internal actor to allow processing of
   # messages already in the stream (required for successful committing).
   # Prefer use of `DrainingControl` over a large stop-timeout.
   stop-timeout = 30s
-  
+
   # Duration to wait for `KafkaConsumer.close` to finish.
   close-timeout = 20s
-  
+
   # If offset commit requests are not completed within this timeout
   # the returned Future is completed `CommitTimeoutException`.
+  # The `Transactional.source` waits this ammount of time for the producer to mark messages as not
+  # being in flight anymore as well as waiting for messages to drain, when rebalance is triggered.
   commit-timeout = 15s
 
   # If commits take longer than this time a warning is logged
@@ -67,7 +70,7 @@ akka.kafka.consumer {
 
   # Not used anymore (since 1.0-RC1)
   # wakeup-debug = true
-  
+
   # Fully qualified config path which holds the dispatcher configuration
   # to be used by the KafkaConsumerActor. Some blocking may occur.
   use-dispatcher = "akka.kafka.default-dispatcher"
@@ -93,6 +96,10 @@ akka.kafka.consumer {
   # This value is used instead of Kafka's default from `default.api.timeout.ms`
   # which is 1 minute.
   metadata-request-timeout = 5s
+
+  # Interval for checking that transaction was completed before closing the consumer.
+  # Used in the transactional flow for exactly-once-semantics processing.
+  eos-draining-check-interval = 30ms
 }
 # // #consumer-settings
 
--- a/core/src/main/scala/akka/kafka/ConsumerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerMessage.scala
@@ -5,15 +5,16 @@
 
 package akka.kafka
 
-import java.util.{Map => JMap}
+import java.util.{Objects, Map => JMap}
 import java.util.concurrent.CompletionStage
 
 import akka.Done
-import akka.annotation.DoNotInherit
-import akka.kafka.internal.CommittableOffsetBatchImpl
+import akka.annotation.{DoNotInherit, InternalApi}
+import akka.kafka.internal.{CommittableOffsetBatchImpl, CommittedMarker}
 import org.apache.kafka.clients.consumer.ConsumerRecord
 
 import scala.concurrent.Future
+import scala.runtime.AbstractFunction2
 
 /**
  * Classes that are used in both [[javadsl.Consumer]] and
@@ -78,15 +79,56 @@ object ConsumerMessage {
   /**
    * Offset position for a groupId, topic, partition.
    */
-  final case class PartitionOffset(key: GroupTopicPartition, offset: Long) {
+  class PartitionOffset(val key: GroupTopicPartition, val offset: Long)
+      extends Product2[GroupTopicPartition, Long]
+      with Serializable {
     def withMetadata(metadata: String) =
       PartitionOffsetMetadata(key, offset, metadata)
+
+    @InternalApi private[kafka] def withCommittedMarker(committedMarker: CommittedMarker) =
+      PartitionOffsetCommittedMarker(key, offset, committedMarker)
+
+    override def toString() = s"PartitionOffset(key=$key,offset=$offset)"
+
+    override def equals(other: Any): Boolean = other match {
+      case that: PartitionOffset =>
+        that.canEqual(this) && Objects.equals(this.key, that.key) &&
+        Objects.equals(this.offset, that.offset)
+      case _ => false
+    }
+
+    override def hashCode(): Int = Objects.hash(key, offset.asInstanceOf[java.lang.Long])
+
+    // This code is for backwards compatibility when PartitionOffset was a case class.
+    override def _1: GroupTopicPartition = key
+
+    override def _2: Long = offset
+
+    override def canEqual(that: Any): Boolean = that.isInstanceOf[PartitionOffset]
+
+    def copy(key: GroupTopicPartition = this.key, offset: Long = this.offset): PartitionOffset =
+      PartitionOffset(key, offset)
+  }
+
+  object PartitionOffset extends AbstractFunction2[GroupTopicPartition, Long, PartitionOffset] {
+    def apply(key: GroupTopicPartition, offset: Long) = new PartitionOffset(key, offset)
+
+    def unapply(arg: PartitionOffset): Option[(GroupTopicPartition, Long)] = Some((arg.key, arg.offset))
   }
 
   /**
    * Offset position and metadata for a groupId, topic, partition.
    */
   final case class PartitionOffsetMetadata(key: GroupTopicPartition, offset: Long, metadata: String)
+
+  /**
+   * Internal Api
+   */
+  @InternalApi private[kafka] final case class PartitionOffsetCommittedMarker(
+      override val key: GroupTopicPartition,
+      override val offset: Long,
+      private[kafka] val committedMarker: CommittedMarker
+  ) extends PartitionOffset(key, offset)
 
   /**
    * groupId, topic, partition key for an offset position.
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -70,6 +70,7 @@ object ConsumerSettings {
     val positionTimeout = config.getDuration("position-timeout").asScala
     val offsetForTimesTimeout = config.getDuration("offset-for-times-timeout").asScala
     val metadataRequestTimeout = config.getDuration("metadata-request-timeout").asScala
+    val drainingCheckInterval = config.getDuration("eos-draining-check-interval").asScala
     new ConsumerSettings[K, V](
       properties,
       keyDeserializer,
@@ -86,6 +87,7 @@ object ConsumerSettings {
       positionTimeout,
       offsetForTimesTimeout,
       metadataRequestTimeout,
+      drainingCheckInterval,
       ConsumerSettings.createKafkaConsumer
     )
   }
@@ -196,6 +198,7 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     val positionTimeout: FiniteDuration,
     val offsetForTimesTimeout: FiniteDuration,
     val metadataRequestTimeout: FiniteDuration,
+    val drainingCheckInterval: FiniteDuration,
     val consumerFactory: ConsumerSettings[K, V] => Consumer[K, V]
 ) {
 
@@ -230,6 +233,7 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     positionTimeout = 5.seconds,
     offsetForTimesTimeout = 5.seconds,
     metadataRequestTimeout = 5.seconds,
+    drainingCheckInterval = 30.millis,
     consumerFactory = ConsumerSettings.createKafkaConsumer
   )
 
@@ -466,6 +470,14 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
   def withMetadataRequestTimeout(metadataRequestTimeout: java.time.Duration): ConsumerSettings[K, V] =
     copy(metadataRequestTimeout = metadataRequestTimeout.asScala)
 
+  /** Scala API: Check interval for TransactionalProducer when finishing transaction before shutting down consumer */
+  def withDrainingCheckInterval(drainingCheckInterval: FiniteDuration): ConsumerSettings[K, V] =
+    copy(drainingCheckInterval = drainingCheckInterval)
+
+  /** Java API: Check interval for TransactionalProducer when finishing transaction before shutting down consumer */
+  def withDrainingCheckInterval(drainingCheckInterval: java.time.Duration): ConsumerSettings[K, V] =
+    copy(drainingCheckInterval = drainingCheckInterval.asScala)
+
   /**
    * Replaces the default Kafka consumer creation logic.
    */
@@ -499,6 +511,7 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
       positionTimeout: FiniteDuration = positionTimeout,
       offsetForTimesTimeout: FiniteDuration = offsetForTimesTimeout,
       metadataRequestTimeout: FiniteDuration = metadataRequestTimeout,
+      drainingCheckInterval: FiniteDuration = drainingCheckInterval,
       consumerFactory: ConsumerSettings[K, V] => Consumer[K, V] = consumerFactory
   ): ConsumerSettings[K, V] =
     new ConsumerSettings[K, V](
@@ -517,6 +530,7 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
       positionTimeout,
       offsetForTimesTimeout,
       metadataRequestTimeout,
+      drainingCheckInterval,
       consumerFactory
     )
 
@@ -539,6 +553,7 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     s"dispatcher=$dispatcher," +
     s"commitTimeWarning=${commitTimeWarning.toCoarsest}," +
     s"waitClosePartition=${waitClosePartition.toCoarsest}," +
-    s"metadataRequestTimeout=${metadataRequestTimeout.toCoarsest}" +
+    s"metadataRequestTimeout=${metadataRequestTimeout.toCoarsest}," +
+    s"drainingCheckInterval=${drainingCheckInterval.toCoarsest}" +
     ")"
 }
--- a/core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
+++ b/core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
@@ -18,6 +18,7 @@ import akka.kafka.ConsumerMessage.{
   _
 }
 import org.apache.kafka.clients.consumer.{ConsumerRecord, OffsetAndMetadata}
+import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.requests.OffsetFetchResponse
 
 import scala.collection.JavaConverters._
@@ -41,15 +42,19 @@ private[kafka] trait PlainMessageBuilder[K, V] extends MessageBuilder[K, V, Cons
 @InternalApi
 private[kafka] trait TransactionalMessageBuilder[K, V] extends MessageBuilder[K, V, TransactionalMessage[K, V]] {
   def groupId: String
+  def committedMarker: CommittedMarker
+  def onMessage(consumerMessage: ConsumerRecord[K, V]): Unit
 
   override def createMessage(rec: ConsumerRecord[K, V]) = {
-    val offset = ConsumerMessage.PartitionOffset(
+    onMessage(rec)
+    val offset = PartitionOffsetCommittedMarker(
       GroupTopicPartition(
         groupId = groupId,
         topic = rec.topic,
         partition = rec.partition
       ),
-      offset = rec.offset
+      offset = rec.offset,
+      committedMarker
     )
     ConsumerMessage.TransactionalMessage(rec, offset)
   }
@@ -94,6 +99,17 @@ private[kafka] trait InternalCommitter {
   // Commit all offsets (of different topics) belonging to the same stage
   def commit(offsets: immutable.Seq[PartitionOffsetMetadata]): Future[Done]
   def commit(batch: CommittableOffsetBatch): Future[Done]
+}
+
+/** Internal API */
+@InternalApi
+private[kafka] trait CommittedMarker {
+
+  /** Marks offsets as already committed */
+  def committed(offsets: Map[TopicPartition, OffsetAndMetadata]): Future[Done]
+
+  /** Marks committing failure */
+  def failed(): Unit
 }
 
 /** Internal API */
--- a/core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
@@ -34,18 +34,11 @@ import scala.concurrent.{Future, Promise}
   final def configureSubscription(): Unit = {
 
     def rebalanceListener(autoSubscription: AutoSubscription): KafkaConsumerActor.ListenerCallbacks = {
-      val partitionAssignedCB = getAsyncCallback[Set[TopicPartition]] { assignedTps =>
-        tps ++= assignedTps
-        log.debug("Assigned partitions: {}. All partitions: {}", assignedTps, tps)
-        requestMessages()
-      }
+      val assignedCB = getAsyncCallback[Set[TopicPartition]](partitionAssignedHandler)
 
-      val partitionRevokedCB = getAsyncCallback[Set[TopicPartition]] { revokedTps =>
-        tps --= revokedTps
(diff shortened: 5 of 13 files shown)


szymonm:
Please don't merge this PR. It's only to demonstrate the issue.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -146,8 +163,10 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
 
   override val onMessageAckCb: AsyncCallback[Envelope[K, V, P]] =
     getAsyncCallback[Envelope[K, V, P]](_.passThrough match {
-      case o: ConsumerMessage.PartitionOffset => batchOffsets = batchOffsets.updated(o)
-      case _ =>
+      case o: ConsumerMessage.PartitionOffsetCommittedMarker =>
+        batchOffsets = batchOffsets.updated(o)
+        println(s"ADDING OFF ${o.offset}")
+      case x => println("Chuj " + x)
```
2m: I think your frustration might be caused by another bug with offset tracking. With a fix for that applied, your test case does not fail for me anymore. The fix is this one: https://github.com/akka/alpakka-kafka/pull/752/commits/d381b4f451e76f125a39424d62ea243eebbae0f3

szymonm: Sorry, this should not be included in the PR. Will rewrite history.

szymonm:
Some afterthoughts:
IIUC, we need to ensure that consumer is not closed before producer finishes committing transaction also in the case of stream failure.
IIUC, in this case the failure message will reach source stage and stop the consumer. However, it will not reach `TransactionalProducerStage` if `producer.commitTransaction` will be in progress (it's blocking). This can lead to race conditions, when producer commits after some other consumer is assigned the partition.

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
```diff
@@ -36,9 +81,73 @@ private[kafka] final class TransactionalSource[K, V](consumerSettings: ConsumerS
     IsolationLevel.READ_COMMITTED.toString.toLowerCase(Locale.ENGLISH)
   )
 
+  case object Drained
+  case class Drain[T](ack: ActorRef, msg: T)
+  case class Committed(offsets: Iterable[ConsumerMessage.PartitionOffset])
```
2m: I do not see that this one would be handled in the stage actor

2m:
Rebased on top of the latest master, and also fixed a bug, where a wrong `Committed` message was being handled in the Trnsational Source stage.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -151,9 +151,9 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
   private def maybeCommitTransaction(beginNewTransaction: Boolean = true): Unit = {
     val awaitingConf = awaitingConfirmation.get
     batchOffsets match {
-      case batch: NonemptyTransactionBatch if awaitingConf == 0 =>
+      case batch: NonemptyTransactionBatch if awaitingConf == 0 && !hasBeenPulled(stage.in) =>
```
szymonm: Can you explain why we need this logic?

szymonm:
Thanks, @2m

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
2m: With this PR, the transaction commit is split into two synchronous blocks in the Producer stage. That means that in between, a message could be sent to the stage which would be sent to the Kafka producer, but the transaction has not yet been started.

So this guard was needed to be sure that no message will come in between the the committing of transaction and beginning of the new one.

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
```diff
@@ -36,9 +81,73 @@ private[kafka] final class TransactionalSource[K, V](consumerSettings: ConsumerS
...
+      override def partitionRevokedCB(revokedTps: Set[TopicPartition]): Unit = {
+        waitForDraining()
+        sourceActor.ref ! Revoked(revokedTps.toList)
+        super.partitionRevokedCB(revokedTps)
+      }
+
+      def waitForDraining(): Unit = {
+        import akka.pattern.ask
+        implicit val timeout = Timeout(5.seconds)
+        Await.result(ask(stageActor.ref, Drain(stageActor.ref, Drained)), 10.seconds)
```
jotek7: I'm afraid if the behaviour in case of a timeout is correct.
IIUC TimeoutException is thrown and it's caught in kafka code by ConsumerCoordinator,
here:

```
try {
            listener.onPartitionsRevoked(revoked);
        } catch (WakeupException | InterruptException e) {
            throw e;
        } catch (Exception e) {
            log.error("User provided listener {} failed on partition revocation", listener.getClass().getName(), e);
        }
```

So due to `catch (Exception e)` we won't fail the stream and we'll just continue like nothing happened and the stream was drained.

Therefore if it takes too long to drain then we still allow messages to remain in the stream. In the meantime some other consumer might start processing data from the same partition because we just released it.

@szymonm @2m Could you verify if my reasoning is correct?

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
szymonm: Still didn't quite get it. I was hoping that suspending demand when transaction is in progress will be enough.
Also, I'm wondering if we just shouldn't block waiting for internal commit.
`kafkaProducer.commitTrasnaction()` is already blocking the whole stage waiting on a network call. And `internalCommit` is just waiting for local memory change, so it shouldn't add much...
That would simplify the code a bit.

szymonm: Making `internalCommit`just send a message would also be possible. Seems like `TransactionalProducerStage` does not have to wait for the SourceActor to acknowledge getting the message, right?

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
szymonm: You're right @jotek7. In this case we should fail the stream starting from the producer. Which is similar to what we should do generally in the failure case (see my comments)

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
2m: > I was hoping that suspending demand when transaction is in progress will be enough.

Calling `suspendDemand` does make sure that `stage.in` is not going to be pulled anymore from that point on, but it could have been pulled from before.

> Seems like TransactionalProducerStage does not have to wait for the SourceActor to acknowledge getting the message, right?

Yes, currently it looks like that's the case. I'll need to think about it a bit more.

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
```diff
@@ -36,9 +81,73 @@ private[kafka] final class TransactionalSource[K, V](consumerSettings: ConsumerS
...
+      override def shuttingDownReceive = super.shuttingDownReceive.orElse(drainHandling)
+
+      def drainHandling: PartialFunction[(ActorRef, Any), Unit] = {
+        case (sender, Committed(offsets)) =>
+          inFlightRecords.committed(offsets.mapValues(_.offset() - 1))
+          sender ! Done
+        case (_, Drain(ack, msg)) =>
+          if (inFlightRecords.empty()) {
+            log.debug("Source drained")
+            ack ! msg
```
2m: This should also send an answer back to the `sender` otherwise the `ask` on line 139 will not be completed.

2m:
I added a commit where it fails the consumer stage if the draining timeout happens.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
```diff
@@ -266,7 +266,7 @@ class TransactionsSpec extends SpecBase(kafkaPort = KafkaPorts.TransactionsSpec)
     "provide consistency when multiple transactional streams are being restarted" in {
       val sourcePartitions = 10
       val destinationPartitions = 4
-      val consumers = 3
+      val consumers = 1
```
szymonm: why did you need to change this here?

szymonm:
> I am not sure if this is a problem. If the stream is being torn down, then after the blocking `producer.commitTransaction` is done, it will complete the stage with `onCompletionFailure` and will not commit the next transaction, which might have duplicates. Am I missing something here?
There is still a chance that consumer will be closed before the producer finishes `producer.commitTransaction`. This can cause the partition to be picked up by another consumer that will start from the offsets that are no yet committed.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
2m: I noticed that this test-case does not finish when running with 3 consumers. The stopping of the stage on drain timeout does not work nicely with how this test uses the RestartSource and it gets restarted all the time and the test never finishes. I also noticed that in Travis it did not finish even with 1 consumer. I need to improve that test-case.

szymonm:
What are the requirements for binary compatibility? After turning `PartitionOffset` case class to a normal class we have some warnings. Should we fix them all?

2m:
> Should we fix them all?

Yes, otherwise we will not be able to include this in a patch release.

szymonm:
Last travis run reports: `The job exceeded the maximum time limit for jobs, and has been terminated.`

The tests for `TransacionsSpec` take a lot of time: 
```[info] - must provide consistency when multiple transactional streams are being restarted (8 minutes, 18 seconds)
[info] - must drain stream on partitions rebalancing (2 minutes, 10 seconds)```

**Comment by szymonm** [comment]
Looks like the test `provide consistency when multiple transactional streams are being restarted` takes 8 mins. We run it twice (for each scala version) and thus we increase the test times from ~30mins to over 50mins (the test for draining takes another 2 mins for each run).

Can we make the test case smaller, @2m ?

**Comment by 2m** [comment]
Yes, lets decrease the number of messages to 100k and see how long does that take.

File: core/src/main/scala/akka/kafka/internal/TransactionalSource.scala [hunk 7]
```diff
@@ -36,9 +82,85 @@ private[kafka] final class TransactionalSource[K, V](consumerSettings: ConsumerS
...
+          sourceActor.ref ! Revoked(revokedTps.toList)
+        } else {
+          sourceActor.ref ! Failure(new Error("Timeout while draining"))
+        }
+        super.partitionRevokedCB(revokedTps)
+      }
+
+      def waitForDraining(): Boolean = {
+        import akka.pattern.ask
+        // Shall we use commitTimeout here?
```
2m: Yes, it should be at least commit timeout. Possible a bit larger as well. Can you refactor this from hardcoded constant to configuration value?

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
```diff
@@ -36,9 +82,85 @@ private[kafka] final class TransactionalSource[K, V](consumerSettings: ConsumerS
...
+          sender ! Done
+        case (sender, Drain(ack, msg)) =>
+          if (inFlightRecords.empty()) {
+            log.debug("Source drained")
+            ack.getOrElse(sender) ! msg
+          } else {
+            log.debug(s"Draining partitions {}", inFlightRecords)
+            // TODO: This timeout should be somehow related to the committing interval.
+            // For instance having 10.millis makes no sense if we commit transaction every second
+            materializer.scheduleOnce(30.millis, new Runnable {
```
2m: Would be good to move this to configuration as well.

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
szymonm: After some thoughts I think using commitTimeout is just ok, will use that.

2m:
Two of the other tests in the TransactionalSpec started failing. Could be the new timing configuration not playing nicely with the test-cases.

szymonm:
This seems to be a problem with `expectN` timeout of 10s by default. Is there a pattern how we could change it on a per-test basis, @2m?

szymonm:
Alternatively, we can change it for all tests (not nice)...

szymonm:
These tests actually catch a valid issue now:
We keep track of offsets emitted by the source and now we are waiting for all emitted offsets. But we should not wait for the offset emitted if processing of this element caused an exception.

szymonm:
If produced is closed after committing a transaction we can safely close the consumer.

szymonm:
Looks like `IntegrationSpec` `signal rebalance events to actor` test is flakky.

szymonm:
TBH there is no reason why this test should work reliably. 
It uses `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest"` config with no offsets storing.

This means that after a rebalancing consumers pick up the latest element from the queue. If there is more elements written to a ktp during the rebalancing this elements are never read.

szymonm:
Can you run tests again?

2m:
> We keep track of offsets emitted by the source and now we are waiting for all emitted offsets. But we should not wait for the offset emitted if processing of this element caused an exception.

Great catch. That was the reason, I saw the multiple "waiting for draining" messages in the logs when the stage was stopping because of failure.

> Looks like IntegrationSpec signal rebalance events to actor test is flaky.

I do not see it failing anymore, but if we see that again, lets create an issue and fix it in a separate PR.

szymonm:
> I do not see it failing anymore, but if we see that again, lets create an issue and fix it in a separate PR.
👍

szymonm:
What are the next steps with the PR?

2m:
I am currently running the test with `consumers = 3` and after a little tweak, where we let the copy flow to restart a couple more times, it is passing. I am going to push that and if that looks green on Travis, this will be good to go in.

2m:
With multiple consumers the test fails with missing messages.

core/src/main/resources/reference.conf
```diff
@@ -93,6 +93,10 @@ akka.kafka.consumer {
   # This value is used instead of Kafka's default from `default.api.timeout.ms`
   # which is 1 minute.
   metadata-request-timeout = 5s
+
+  # Interval for cheking that transaction was completed before closing the consumer.
```
seglo: I don't see any more documentation for the end user about his new functionality.  I suggest updating the Transactions section in the docs.

seglo: [Minor] Typo on "checking".

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -159,21 +190,35 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
...
-      scheduleOnce(commitSchedulerKey, commitInterval)
-    }
+    batch
+      .internalCommit()
+      .onComplete { _ =>
+        onInternallCommitAckCb.invoke(beginNewTransaction)
+      }(materializer.executionContext)
+  }
+
+  val onInternallCommitAckCb: AsyncCallback[Boolean] = {
```
seglo: [Minor] Typo `onInternalCommitAckCb`

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -159,21 +190,35 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
   override def onCompletionFailure(ex: Throwable): Unit = {
     log.debug("Aborting transaction due to stage failure")
     abortTransaction()
+    batchOffsets.committingFailed()
     super.onCompletionFailure(ex)
   }
 
   private def commitTransaction(batch: NonemptyTransactionBatch, beginNewTransaction: Boolean): Unit = {
     val group = batch.group
     log.debug("Committing transaction for consumer group '{}' with offsets: {}", group, batch.offsetMap())
```
seglo: [Minor] Use the local variable offsetMap so we don't have to to build for each log line?

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
```diff
@@ -6,24 +6,73 @@
...
+                                                     commitTimeout: FiniteDuration)
     extends KafkaSourceStage[K, V, TransactionalMessage[K, V]](
       s"TransactionalSource ${subscription.renderStageAttribute}"
     ) {
   require(consumerSettings.properties(ConsumerConfig.GROUP_ID_CONFIG).nonEmpty, "You must define a Consumer group.id.")
 
+  type Offset = Long
+
+  private trait InFlightRecords {
+    // Assumes that offsets per topic partition are added in the increasing order
```
ennru: What makes this assumption hold? Can you explain it in the comment, please.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -111,32 +132,44 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
     setHandler(stage.out, new OutHandler {
```
ennru: These `setHandler` and `demandSuspended = false` should possibly become a method to stress they belong together.

core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -96,6 +101,16 @@ private[kafka] trait InternalCommitter {
   def commit(batch: CommittableOffsetBatch): Future[Done]
 }
 
+/** Internal API */
+@InternalApi
+private[kafka] trait CommittedMarker {
+  // Marks offsets as already committed
```
ennru: These should be scaladoc-style comments.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -540,5 +554,6 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     s"commitTimeWarning=${commitTimeWarning.toCoarsest}," +
     s"waitClosePartition=${waitClosePartition.toCoarsest}," +
     s"metadataRequestTimeout=${metadataRequestTimeout.toCoarsest}" +
```
ennru: Comma missing at the end.

core/src/main/resources/reference.conf
```diff
@@ -93,6 +93,10 @@ akka.kafka.consumer {
   # This value is used instead of Kafka's default from `default.api.timeout.ms`
   # which is 1 minute.
   metadata-request-timeout = 5s
+
+  # Interval for checking that transaction was completed before closing the consumer.
+  # Used in the transactional flow.
+  draining-check-interval = 30ms
```
ennru: I wonder if the setting name should contain `eos` or something. There are so many settings now it is hard to see which belong together.

core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -78,16 +79,53 @@ object ConsumerMessage {
...
   }
 
   /**
    * Offset position and metadata for a groupId, topic, partition.
    */
   final case class PartitionOffsetMetadata(key: GroupTopicPartition, offset: Long, metadata: String)
 
+  final case class PartitionOffsetCommittedMarker(override val key: GroupTopicPartition,
+                                                  override val offset: Long,
+                                                  private[kafka] val committer: CommittedMarker)
```
ennru: I'd prefer `committedMarker` as name to distinguish it from `Committer`.

tests/src/test/scala/akka/kafka/scaladsl/TransactionsSpec.scala
```diff
@@ -349,6 +356,120 @@ class TransactionsSpec extends SpecBase(kafkaPort = KafkaPorts.TransactionsSpec)
...
+      val probeConsumerGroup = createGroupId(2)
+      val probeConsumer = valuesProbeConsumer(probeConsumerSettings(probeConsumerGroup), sinkTopic)
+
+      periodicalCheck("Wait for elements written to Kafka", maxTries = 30, 1.second) {
+        elementsWrote.get
+      }(_ > 10)
+
+      probeConsumer
+        .request(elements)
+        .expectNextUnorderedN((1 to elements).map(_.toString))
```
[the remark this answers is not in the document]
szymonm: We are reading from multiple partitions, kafka gives us a guarantee that messages in a single partition are ordered.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
szymonm: We have methods `resumeDemand` and `suspendDemand` now. Is this what you were thinking of?

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
szymonm: Added:
```
    // The assumption is true for a Kafka consumer that guarantees that elements are emitted
    // per partition in offset-increasing order.```

File: core/src/main/resources/reference.conf (hunk 16 above)
Comment (szymonm) [reply core/src/main/resources/reference.conf]: tbh... `eos` is cryptic even for me...

Comment (szymonm) [reply core/src/main/resources/reference.conf]: but we may have some distinct section for transactional processing?

**Comment by szymonm** [comment]
The test for the failure case is still failing showing both duplication and data loss.

My intuition what happens is as follows:
1. Data loss
Since we send messages and offsets in transaction the only possibility to loose data is to loose some records while they are in the stream and commit records with higher offsets.

2. Duplication
We close consumer before the producer in a single stream. Thus another consumer starts consuming from the partition before the previous producer flushes last offsets.

**Comment by szymonm** [comment]
I'm also seeing some problems with the draining mechanism, specifically it happens that it waits fvor the last message that is never drained.

**Comment by szymonm** [comment]
Also setting higher backoff for restarts increases the probability of test success (from what I observed)
```RestartSource
          .onFailuresWithBackoff(1.second, 5.seconds, 0.2)```

**Comment by 2m** [comment]
> Since we send messages and offsets in transaction the only possibility to loose data is to loose some records while they are in the stream and commit records with higher offsets.

I only noticed lost messages when running that test-case with multiple consumers. Investigating that right now.

> Duplication

I have not seen test-case failing because of duplication for quite some time now. The duplication check is first, and at least in Travis it passes now for quite some time before the test-case fails with missing elements check.

**Comment by 2m** [comment]
Pushed two more commits that improve the failing test-case in a couple of ways. When it comes to duplicates, it could very well be, that we get duplicated messages when rebalance happens while consuming messages after transactional processing. For that I added offset tracking which will ignore duplicates that have same offsets. We should only care about duplicates with different offsets, because those would be from the transactions.

Also I moved the place where the transactional processing flow is terminated in the test stream, so the stream is always terminated with an error (from upstream) instead of cancellation (from downstream). Lets see what Travis thinks about that when it comes to missing messages.

**Comment by szymonm** [comment]
I'm looking into it too, and what worries me know is seeing `java.lang.Error: Timeout while draining`. My intuition is that with embedded kafka and timeout set to 15s we have plenty of time to drain.
Maybe timeout indicates that we don't pull correctly while draining?

**Comment by szymonm** [comment]
Also, when the timeout happens, we should make sure that close the producer before closing the consumer. This doesn't happen, right?

**Comment by 2m** [comment]
> Also, when the timeout happens, we should make sure that close the producer before closing the consumer.

Not explicitly, but there is a `akka.kafka.consumer.stop-timeout` that delays stopping the ConsumerStage so late commits can happen when using a CommittableSource.

I see that we are overriding `stopConsumerActor` in TransactionalSource and not doing the delay anymore. We should probably bring the delay back so the producer gets closed first.

> Maybe timeout indicates that we don't pull correctly while draining?

Yea, noticed those timeouts as well. But I also see `Source drained.` in the logs. Could it be that the reply after draining does not come back to complete the ask Future?

**Comment by szymonm** [comment]
> Yea, noticed those timeouts as well. But I also see Source drained. in the logs. Could it be that the reply after draining does not come back to complete the ask Future?

That was another idea I had!

**Comment by szymonm** [comment]
After increasing the parallelism of sending messages to kafka we don't have to wait for messages to be flushed. As a result we are not failing stream draining that often, which gives very good chance of the test to pass (didn't see a single failure on my machine after changing that).

It also speeds up the test a lot!

**Comment by 2m** [comment]
Looks good. I also like the simplification in the Producer where the transaction is started sooner. That simplification means that these two commits are not needed anymore, because transaction is committed and new transaction is started in the same pass.

https://github.com/akka/alpakka-kafka/pull/757/commits/7be3450b28f140d75a0d0ff626b9a64f66178945
https://github.com/akka/alpakka-kafka/pull/757/commits/9c04dcf2a8d730b54ac9e9cd35b025ff4a544e6c

**Comment by 2m** [comment]
Pushed a revert of those two commits. I'll do a docs update and then this will be good to go.

**Comment by szymonm** [comment]
Yeap, happy to see this simplified again!

File: core/src/main/scala/akka/kafka/ConsumerMessage.scala [hunk 19]
```diff
@@ -78,16 +79,53 @@ object ConsumerMessage {
...
+
+    def unapply(arg: PartitionOffset): Option[(GroupTopicPartition, Long)] = Some((arg.key, arg.offset))
   }
 
   /**
    * Offset position and metadata for a groupId, topic, partition.
    */
   final case class PartitionOffsetMetadata(key: GroupTopicPartition, offset: Long, metadata: String)
 
+  final case class PartitionOffsetCommittedMarker(override val key: GroupTopicPartition,
```
ennru: Internal API.

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -78,15 +71,30 @@ import scala.concurrent.{Future, Promise}
...
+      completeStage()
+  }
+
+  protected def stopConsumerActor(): Unit =
     materializer.scheduleOnce(settings.stopTimeout, new Runnable {
       override def run(): Unit =
         consumerActor.tell(KafkaConsumerActor.Internal.Stop, sourceActor.ref)
     })
+
+  protected def partitionAssignedCB(assignedTps: Set[TopicPartition]): Unit = {
```
ennru: I find the `CB` strange as this is not an `AsyncCallback`, maybe `Signal` or `Handler`?

docs/src/main/paradox/transactions.md
```diff
@@ -30,12 +30,14 @@ Transactions are committed on an interval which can be controlled with the produ
 
 When the stream is materialized the producer will initialize the transaction for the provided `transactional.id` and a transaction will begin.  Every commit interval (`eos-commit-interval`) we check if there are any offsets available to commit.  If offsets exist then we suspend backpressured demand while we drain all outstanding messages that have not yet been successfully acknowledged (if any) and then commit the transaction.  After the commit succeeds a new transaction is begun and we re-initialize demand for upstream messages.
 
+Messages are also drained from the stream when the consumer gets a rebalance of partitions. In that case, the consumer will wait in the `onPartitionsRevoked` callback until all of the messages have been drained from the stream and transaction is committed before allowing the rebalance to continue. The amount of total time the consumer will wait for draining is controlled by the `akka.kafka.consumer.commit-timeout`, and the interval between checks is controlled by the `akka.kafka.consuner.eos-draining-check-interval` configuration settings.
```
ennru: "the" transaction

```suggestion
Messages are also drained from the stream when the consumer gets a rebalance of partitions. In that case, the consumer will wait in the `onPartitionsRevoked` callback until all of the messages have been drained from the stream and the transaction is committed before allowing the rebalance to continue. The amount of total time the consumer will wait for draining is controlled by the `akka.kafka.consumer.commit-timeout`, and the interval between checks is controlled by the `akka.kafka.consuner.eos-draining-check-interval` configuration settings.
```

docs/src/main/paradox/transactions.md
```diff
@@ -30,12 +30,14 @@ Transactions are committed on an interval which can be controlled with the produ
...
+
 To gracefully shutdown the stream and commit the current transaction you must call `shutdown()` on the `Control` (@scala[@scaladoc[API](akka.kafka.scaladsl.Consumer$$Control)]@java[@scaladoc[API](akka.kafka.javadsl.Consumer$$Control)]) materialized value to await all produced message acknowledgements and commit the final transaction.  
 
 ### Simple Example
 
 Scala
-: @@ snip [snip](/tests/src/test/scala/docs/scaladsl/TransactionsExample.scala) { #transactionalSink }
+: @@ snip [snip](/tests/src/test/scala/docs/4scaladsl/TransactionsExample.scala) { #transactionalSink }
```
ennru:
```suggestion
: @@ snip [snip](/tests/src/test/scala/docs/scaladsl/TransactionsExample.scala) { #transactionalSink }
```

2m:
Rebased on top of master and addressed the latest feedback.

PR #759, opened 2019-03-25
Commit refreshing: avoid call to position when disabled

ennru:
## Purpose

When commit-refreshing is disabled (which is the default), the calls to `consumer.position` in the rebalance handler's `onPartitionAssigned` can be avoided.
This further isolates the commit refreshing so it doesn't obstruct the `KafkaConsumerActor`.

## Background Context

A follow-up to #753

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -117,12 +117,13 @@ import scala.util.control.NonFatal
   }
 
   private[KafkaConsumerActor] trait CommitRefreshing {
+    def enabled: Boolean
     def add(offsets: Map[TopicPartition, OffsetAndMetadata]): Unit
     def committed(offsets: Map[TopicPartition, OffsetAndMetadata]): Unit
     def revoke(revokedTps: Set[TopicPartition]): Unit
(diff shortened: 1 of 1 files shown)


core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -227,13 +231,15 @@ import scala.util.control.NonFatal
...
+      checkOverlappingRequests("Assign", sender(), assignedTps)
       val previousAssigned = consumer.assignment()
-      consumer.assign((tps.toSeq ++ previousAssigned.asScala).asJava)
-      val assignedOffsets = tps.map(tp => tp -> consumer.position(tp)).toMap
-      partitionAssigned(assignedOffsets)
+      consumer.assign((assignedTps.toSeq ++ previousAssigned.asScala).asJava)
+      if (commitRefreshing.enabled) {
+        val assignedOffsets = assignedTps.map(tp => tp -> consumer.position(tp, positionTimeout)).toMap
+        commitRefreshing.assignedPositions(assignedTps, assignedOffsets)
+      }
```
2m: Instead of checking `commitRefreshing.enabled` we could call `commitRefreshing.assignedPositions` with `assignedTps` and `consumer`, where it would do the right thing, and in the NoOp implementation it would do nothing.

PR #761, opened 2019-03-29
Synchronous partition assignment handler (internal)

ennru:
## Purpose

**For now this PR just introduces the synchronous partition assignment handler internally without a user-facing API as the use-cases for this need to be understood better.**

Introduce a new callback for changes in partitions assigned to an Alpakka Kafka consumer. This allows executing code before a rebalance continues, and querying the underlying Kafka consumer.

These callbacks are much more powerful (and dangerous) than the current rebalance listener actor. The callbacks are executed on the same thread that called `poll()` (the actor thread) so they are allowed to call into the Kafka consumer eg. to query current offsets, commit offsets or to seek to an offset position. The callbacks add to the time `poll()` is executed and block the internal Kafka consumer actor from doing any other work. As this can easily make influence other Kafka consumer actors running on the same dispatcher, the time the callbacks take is checked and warnings are issued to the logs if they pass a configurable threshold.

This allows for new ways to handle offsets which might show more efficient for certain use-cases. The tests show an example for external offset storage and one example which commits on rebalance only.

## Background Context

It has been discussed in #539 that the lack of blocking user-defined call-backs to react on Kafka's partition rebalancing makes certain use cases impossible with Alpakka Kafka.

In particular, it is impossible to combine Kafka's partition assignment with external offset storage.

## References

References #539 
Follow up #841

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -273,11 +273,19 @@ lazy val tests = project
         "org.slf4j" % "jul-to-slf4j" % slf4jVersion % Test,
         "org.mockito" % "mockito-core" % "2.24.5" % Test
       ) ++ {
-        if (scalaBinaryVersion.value == "2.13") Seq()
-        else
-          Seq(
-            "io.github.embeddedkafka" %% "embedded-kafka-schema-registry" % embeddedKafkaSchemaRegistry % Test exclude ("log4j", "log4j") exclude ("org.slf4j", "slf4j-log4j12")
-          )
+        scalaBinaryVersion.value match {
+          case "2.13" =>
+            Seq()
+          case "2.12" =>
+            Seq(
+              "com.typesafe.akka" %% "akka-actor-typed" % akkaVersion % Test,
+              "io.github.embeddedkafka" %% "embedded-kafka-schema-registry" % embeddedKafkaSchemaRegistry % Test exclude ("log4j", "log4j") exclude ("org.slf4j", "slf4j-log4j12")
+            )
+          case "2.11" =>
+            Seq(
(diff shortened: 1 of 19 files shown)


seglo:
This looks really promising @ennru .  When I was considering an API I was thinking about providing strategies that users could opt in with (i.e. "commit on revoked partitions"), and then provide a super user interface which just exposed the consumer itself, but I like what I like the idea behind `RestrictedConsumer` better.

build.sbt
```diff
@@ -215,11 +215,18 @@ lazy val tests = project
...
-            "io.github.embeddedkafka" %% "embedded-kafka-schema-registry" % "5.2.1" % Test exclude ("log4j", "log4j") exclude ("org.slf4j", "slf4j-log4j12")
-          )
+        scalaBinaryVersion.value match {
+          case "2.13" =>
+            Seq()
+          case "2.12" =>
+            Seq(
+              "io.github.embeddedkafka" %% "embedded-kafka-schema-registry" % "5.2.1" % Test exclude ("log4j", "log4j") exclude ("org.slf4j", "slf4j-log4j12")
+            )
+          case "2.11" =>
```
2m: This could be made a bit shorter: `case "2.11" | "2.12" => ...`

2m:
Looking good. I like how TransactionalSource was able to use this without any code changes.

ennru:
For now, I removed the user-facing API so that the new partition assignment handler is used internally, only.
The plan is to continue to explore the scenarios this kind of partition assignment handler improves as eg. #841 and #843

Issue #773, opened 2019-04-19
Benchmarks work only for 1-partitioned topics (low performance for more partitions)


mowczare:
The benchmarks included in `akka.kafka.benchmarks` package do not work when local Kafka cluster config has `num.partitions` > 1. Both producer and consumer hang waiting for an offset equal to number of messages sent (which can never happen in the aforementioned test case).

What's more, I was concerned of the "at-least once batched" consumer performance on my production environment so I was looking for the benchmark results. I found this:
https://docs.google.com/document/d/101VfFarQNFBW23PYaT09MSLtmtsK03r7mViGt2PCa0Q/edit#
and this:
https://docs.google.com/spreadsheets/d/1SJDQsPHaKyCLBEkPQvBay3thq6I3Xvm_EzuIh1nuKMA/edit#gid=0
456,695 messages per second was an out-of-space number comparing to my 1k per second production (64-partitions topics) experiences. 

So I hotfixed the benchmarks and run the `XKafkaBatchedConsumer`'like tests locally on both 1-partition and 8-partition topics.

Below are the results (considering the benchmark doesn't work on any other number of partitions than 1, I assume the previously reported results pertained to the topic with 1 partition, please verify):

| name of the test | number of partitions | msgs/s previously reported in sheet | msgs/s on my local machine |
| - | - | - | - |
| ApacheKafkaBatchedConsumer | 1 | 1,194,331 | 1,108,237 |
| ApacheKafkaBatchedConsumer | 8 | - | 1,642,036 |
| AlpakkaKafkaBatchedConsumer | 1 | 456,695 | 75,956 |
| AlpakkaKafkaBatchedConsumer | 8 | - | 1017 |

The nature of the consumption is also very curious. I've managed to observe it using the Kafka Tool consumers lag feature and only one partition is consumed immediately, whereas others are consumed very slowly (500 messages per half a second, sequentially).

This seems to affect only this connector, similar ones (like https://github.com/monix/monix-kafka) work fine and performant.

2m:
Really appreciate the analysis. We have dedicated time in this sprint to work on the performance of this connector so this is very timely and helpful.

Also would you mind creating a PR with fixed benchmarks for the multi-partition topics?

mowczare:
Sure:
https://github.com/akka/alpakka-kafka/pull/774/files

I stayed with the "number of partitions should be choosed with kafka brokers config" approach.

2m:
[...]

PR #774, opened 2019-04-19
Benchmarks now support multi-partitioned topics tests

mowczare:
## Purpose

Current benchmarks do not support multi-partitioned topic tests.

## References

https://github.com/akka/alpakka-kafka/issues/773


2m:
Test failure was from the already fixed TransactionsSpec.

PR #784, opened 2019-05-06
TransactionalSource: drain only partitions that were revoked

szymonm:
## Purpose
Currently, we drain all partitions when receiving revoke signal from kafka. It's better to drain only those that were revoked. Especially, given that Kafka sends also the signal with empty partitions list.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
+++ b/core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
@@ -45,8 +45,9 @@ private[kafka] final class TransactionalSource[K, V](consumerSettings: ConsumerS
     def committed(offsets: Map[TopicPartition, Offset]): Unit
     def revoke(revokedTps: Set[TopicPartition]): Unit
     def reset(): Unit
+    def assigned(): Set[TopicPartition]
 
-    def empty(): Boolean
+    def empty(partitions: Set[TopicPartition]): Boolean
   }
 
   private object InFlightRecords {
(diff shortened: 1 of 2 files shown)


szymonm:
I see binary compatibility failure, but I'm changing only Internal api, right?

2m:
Yes, but the internal classes are not ignored automatically. Please create a file `core/src/main/mima-filters/1.0.2.backwards.excludes` that contains:

```
ProblemFilters.exclude[Problem]("akka.kafka.internal.*")
```

core/src/main/scala/akka/kafka/internal/TransactionalSource.scala
```diff
@@ -138,11 +141,12 @@ private[kafka] final class TransactionalSource[K, V](consumerSettings: ConsumerS
...
 
       override protected def stopConsumerActor(): Unit =
-        sourceActor.ref.tell(Drain(Some(consumerActor), KafkaConsumerActor.Internal.Stop), sourceActor.ref)
+        sourceActor.ref.tell(Drain(inFlightRecords.assigned(), Some(consumerActor), KafkaConsumerActor.Internal.Stop),
+                             sourceActor.ref)
 
       // This is invoked in the KafkaConsumerActor thread when doing poll.
       override def partitionRevokedHandler(revokedTps: Set[TopicPartition]): Unit = {
-        if (waitForDraining()) {
+        if (waitForDraining(revokedTps)) {
```
2m: AFAIK we always get all partitions revoked, and then get assigned again the ones, that are kept to the same consumer. But anyways, this can come in anticipating future Kafka client improvements.

2m:
Failure was #788

Issue #785, opened 2019-05-06
waitForDraining always times out in TransactionalSource


charlibot:
### Versions used 

1.0.2

### Expected Behavior

The `waitForDraining` method of `TransactionalSource` always seems to cause a timeout after a partition rebalancing. To reproduce:

1. Create an input topic with more than 1 partition
2. Start the graph below and wait for it to start
3. With the first graph running, start the graph again
4. Wait for `commit-timeout` duration where first graph will fail with timeout.

No need to send any messages.

The logs actually indicate the `Source drained` log does happen but after the timeout so my guess is the `Await.result` blocks the message handling from happening until after the timeout.

```scala
Transactional
  .source(consumerSettings, Subscriptions.topics(topic))
  .groupBy(4, _.partitionOffset.key.partition)
  .map { msg =>
    ProducerMessage.single(new ProducerRecord[String, String](outTopic, msg.record.value()),
      msg.partitionOffset)
  }
  .via(Flow.lazyInit(event => {
    Future.successful(
      Transactional.flow(producerSettings, "txid-" + event.passThrough.key.partition)
    )
  },
    () => Future.successful(NotUsed)
  ))
  .mergeSubstreams
  .toMat(Sink.ignore)(Keep.both)
  .mapMaterializedValue(DrainingControl.apply)
  .run()
```

### Actual Behavior

After starting the second graph, the first graph should drain and not fail.

### Reproducible Test Case

The `drain stream on partitions rebalancing` test in `TransactionsSpec` can be changed to create the source topic with partitions = 4 for example.

charlibot:
Hi @szymonm, did you run into the above when you working on your PR?

szymonm:
Yes, I'm also seeing increased number of `Timeout while draining`. I'm not sure if it times out every time. I see that a PR with a fix: https://github.com/akka/alpakka-kafka/pull/784 helps a bit.

Does it fix your test case?

charlibot:
Unfortunately not, and for me it times out every time even though I've not sent any data.

charlibot:
Can the stage actor only handle one `getAsyncCallback` at a time? If so, because the `waitForDraining` is within a `getAsyncCallback` and blocks, the Drain message is not handled until after that `waitForDraining` callback has completed. Underneath the `getStageActor`, `The messages are looped through the [[getAsyncCallback]] mechanism of [[GraphStage]]`.

szymonm:
[...]

PR #787, opened 2019-05-06
Transactional revoke on consumer thread

szymonm:
## Purpose

As observed by @charlibot waiting for draining on the source actor thread is wrong (this happens because of using `asyncCallback`), because we never receive a message that the stream is drained. This fix is intended to fix that by forcing waiting for the draining on the consumer thread.
It also blocks the consumer from polling/releasing partitions, which was my intention from the beginning.

This is a quick fix. We should implement a proper fix on top of #761 

## References

References #785 #784 #761


2m:
Rebased on top of master.

2m:
Failure was #788

Issue #788, opened 2019-05-06
Failed AssignmentSpec must consume from the specified topic pattern 


2m:
Fails in Travis:

```
[info] - must consume from the specified topic pattern *** FAILED *** (16 seconds, 55 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 16 times, sleeping 1000000000 nanoseconds between each query. (AssignmentSpec.scala:71)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at docs.scaladsl.AssignmentSpec$$anonfun$1$$anonfun$apply$mcV$sp$2$$anonfun$apply$5.apply(AssignmentSpec.scala:71)
[info]   at docs.scaladsl.AssignmentSpec$$anonfun$1$$anonfun$apply$mcV$sp$2$$anonfun$apply$5.apply(AssignmentSpec.scala:49)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:30)
[info]   at docs.scaladsl.AssignmentSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(AssignmentSpec.scala:49)
[info]   at docs.scaladsl.AssignmentSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(AssignmentSpec.scala:49)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[...]

PR #792, opened 2019-05-08
Use multiple partitions in draining test

2m:
As suggested in https://github.com/akka/alpakka-kafka/issues/785#issuecomment-489627986


2m:
Failure was #788

PR #799, opened 2019-05-09
Add Committer.batchFlow

2m:
## Purpose

Adds `Committer.batchFlow` that emits `CommittableOffsetBatch` for every committed batch.

## Changes

* renamed benchmark directories that contained spaces
* increased message count in batched benchmarks
* added `Committer.batchFlow`
* refactored `apache-kafka-batched-consumer` to use the `CommittableOffsetBatch` which ended up doubling message throughput.


build.sbt
```diff
@@ -302,7 +302,7 @@ lazy val benchmarks = project
     kafkaScale := 1,
     buildInfoPackage := "akka.kafka.benchmarks",
     buildInfoKeys := Seq[BuildInfoKey](kafkaScale),
-    dockerComposeTestLogging := true,
```
ennru: Is that intended to go in?

2m: Yes. When this is true, it prints all of the logs from docker containers (so Kafka starting up and shutting down) to the console. This makes it hard to locate the bench results.

PR #802, opened 2019-05-10
Emit only messages from TransactionalFlow only after committing them

### LINKED ISSUE #791: Transactional.flow should emit only after a message is committed
### Short description

Currently `Transactional.flow` emits elements downstream only after they are acked by a Kafka broker, but not necessarily after they are committed.
This is counter intuitive, because it may produce messages even if transaction that contains them is aborted or fails.

Speaking more practically, currently there is no way to verify that a record was successfully committed. This means that I cannot count the number of elements that were written to Kafka.

I also believe that this behavior is never described in the documentation and can be confusing to the users.

What are your thoughts around this @2m @ennru @seglo ?

### Possible solutions
1. Change the behaviour of transactional flow to emit only committed elements.
2. Implement a new sink that will emit only after committing successfully.
3. Add a way to know which elements were committed.

szymonm:
## Purpose

Resolves #791 

## Changes

Introduces waiting for commit message before emitting from an element from the `TransactionalProducerStage`.

This has the following effect on the logic:
After the `TransactionalProducerStage` we have `mapAsync(txSettings.parallelism)`.
Before in the `mapAsync` stage, we used to wait for Kafka ack message (`producer.send` callback).
Now we are going to wait for the transaction to complete (`producer.commitTransaction`).
Thus we need to increase the `txSettings.parallelism` to accommodate committing many messages in the same transaction. 

What do you think about this change @2m, @ennru ?

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
@@ -114,8 +114,7 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
           r.success(Result(metadata, msg))
         }))
         postSend(msg)
-        val future = r.future.asInstanceOf[Future[OUT]]
-        push(stage.out, future)
+        emit(r.future.asInstanceOf[Future[OUT]])
 
       case multiMsg: MultiMessage[K, V, P] =>
         val promises = for {
@@ -131,13 +130,11 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
         val res = Future.sequence(promises).map { parts =>
           MultiResult(parts, multiMsg.passThrough)
         }
-        val future = res.asInstanceOf[Future[OUT]]
-        push(stage.out, future)
+        emit(res.asInstanceOf[Future[OUT]])
 
       case passthrough: PassThroughMessage[K, V, P] =>
         postSend(passthrough)
-        val future = Future.successful(PassThroughResult[K, V, P](in.passThrough)).asInstanceOf[Future[OUT]]
-        push(stage.out, future)
+        emit(Future.successful(PassThroughResult[K, V, P](in.passThrough)).asInstanceOf[Future[OUT]])
 
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -167,6 +176,13 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
     }
   }
 
+  override protected def emit(future: Future[Results[K, V, P]]): Unit = {
+    implicit val ec: ExecutionContext = this.materializer.executionContext
+    val committed: Future[Results[K, V, P]] = future.zip(batchOffsets.isCommitted()).map(_._1)
```
2m: `batchOffsets.isCommitted().flatMap(_ => future)` might be a bit cleaner

2m:
Interesting approach! It would be useful to play with the parallelism and see how it affects performance, as the smaller it is the quicker `mapAsync` buffer will fill in and the stream will backpressure until transaction is committed.

Transaction benchmarks will be usable for this after https://github.com/akka/alpakka-kafka/pull/804 gets in.

patriknw:
@szymonm Seems like this has been stale for a long while. Shall we close as obsolete?

szymonm:
I think that #791 is valid and this is the right solution.

I can resolve the conflicts and improve the code as @2m suggested, if you agree with me on the principals.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -72,8 +77,10 @@ private object TransactionalProducerStage {
       case (gtp, offset) => new TopicPartition(gtp.topic, gtp.partition) -> new OffsetAndMetadata(offset + 1)
     }
 
-    def internalCommit(): Future[Done] =
+    def internalCommit(): Future[Done] = {
+      committed.success(Done)
```
patriknw, line 81: are we sure this promise is not completed elsewhere, or use `trySuccess`

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -159,6 +156,9 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     }
   }
 
+  protected def emit(future: Future[OUT]): Unit =
```
patriknw, line 159: it's somewhat confusing to name this `emit` since that is a method in Akka's GraphStage. Can we come up with another name? `pushOut`?

patriknw (commented):
this looks like the right approach

PR #809, opened 2019-05-16
Use Produer Send API with callback

2m:
## Purpose

Alpakka Kafka uses Kafka Producer Send API with callback support. We should use the same in raw Apache Kafka benchmarks.


2m:
Ahh, I thought we do not compile benchmarks with 2.11. Maybe we should change that.

PR #828, opened 2019-06-17
Use Graal JIT compiler when running tests with JDK11

2m:
## Purpose

Use Graal JIT compiler when running tests with JDK11.

## Changes

Command flags are added to the `JAVA_OPTS` environment flag which is automatically picked up by sbt.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ jobs:
 
     - env:
       - JDK="adopt@~1.11.0-2"
+      - _JAVA_OPTIONS="-XX:+UnlockExperimentalVMOptions -XX:+UseJVMCICompiler"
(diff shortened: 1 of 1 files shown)


ennru:
Should we add something to https://doc.akka.io/docs/alpakka-kafka/current/production.html ?

2m:
From the Travis logs, I see that JVM_OPTS is overwritten, but I am not sure from where (sbt-extras?):

```
$ export JVM_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseJVMCICompiler"
...
$ export JVM_OPTS=@/etc/sbt/jvmopts
```

Lets try `_JAVA_OPTIONS` instead.

PR #829, opened 2019-06-18
Remove doc hostname logic

2m:
## Purpose

Removes a switch of how the hostname to the Alpakka Kafka API docs is determined. Instead it uses absolute URL.


2m:
Failure was #832

PR #830, opened 2019-06-18
Enable fatal warnings in CI

2m:
## Purpose

Enables fatal warnings in CI, so unused import and other warnings do not slip through to master.


ennru:
For Scala 2.13 there are a lot of warnings right now. Some will be easier to solve once 2.13.1 comes along. I suggest not enabling it for Scala 2.13.

Maybe the contributor's advice needs to talk about this. The Travis stage names should probably show that it is enabled to make it less surprising.

2m:
Alright.

I also noticed warnings like these:

```
[warn] /home/martynas/projects/alpakka-kafka/tests/src/test/scala/akka/kafka/internal/ProducerSpec.scala:534:20: method close in trait Producer is deprecated: see corresponding Javadoc for more information.
[warn]       .when(result.close(mockito.ArgumentMatchers.any[Long], mockito.ArgumentMatchers.any[TimeUnit]))
[warn]             
```

but when browsing into the Apache Kafka `Producer` code from IntelliJ no deprecation tag was on the methods. There was a mismatch between versions somewhere.

ennru:
That is interesting. I see `kafka-clients-2.2.0` being pulled in. There it is deprecated in favour of `close(Duration)`.

ennru:
The schema registry pulls in Kafka 2.2.0 see `tests/Test/dependencyTree`.
The fatal warnings can't be easily applied to `doc` tasks right now...

ennru:
But some more work to do...

2m:
Collection convertion deprecation warnings will go away, once https://github.com/scala/scala-collection-compat/pull/217 is released and we start using it.

ennru:
What about merging this without enabling `fatal-warnings` for now?

2m:
Validation should pass now after disabling fatal-warnings in scaladoc.

Issue #674, opened 2018-12-05
Detect if broker is available (when using Kafka 2.x client)


ennru:
When using the new Kafka consumer API from Kafka 2.x, we won't use `WakeupException`s anymore to stop the client from blocking the actor thread.
Until now theses exceptions served as an indication that there was now Kafka broker available, or that it disappeared. This won't be detected automatically with the new API use.
For use cases where we actually want to fail the consuming stream when there is no Kafka broker, a regular call to `listTopics(Duration)` would be able to check the connection.
(As suggested on [StackOverflow](https://stackoverflow.com/questions/37920923/how-to-check-whether-kafka-server-is-running))

mriehl:
Hi,

We noticed that when a broker is assigned a new IP (due to machine failure for example), the source never fails anymore. This is an issue because the kafka consumer library will cache the broker `InetSocketAddress`es on startup, and never update them again.
We have a retry backoff supervision around the consumer source that will re-create it if the stream blows up. This used to work because the source would blow up, and then our supervision would recreate it which causes the kafka consumer to resolve the broker IPs anew.

The new API is pretty terrible, because now the source never fails. And since it doesn't fail, we don't recreate it and the kafka consumer keeps its resolved broker `InetSocketAddress` forever.

mriehl:
Damn, I was pretty sure I was using the latest from maven central but apparently still on `1.0`. I'll upgrade and see what happens, but it really looks like they solved it. Thanks!

mriehl:
Can confirm using 1.0.1 solves the issue

ennru:
I'm glad to hear that. Thanks for getting in touch.

gabrielgiussi:
@ennru could I give it a try?

ennru:
Sure, that would be great.

tayvs:
@gabrielgiussi @ennru Is there any updates?

ennru:
This could possibly be implemented together with https://github.com/akka/alpakka-kafka/issues/765

tayvs:
As were mentioned, `listTopics` can be used to check connectivity. 
What if there was a scheduler that periodically checks `listTopics` and in case of failure retries exponentially configurable number of attempts. After attempts exceeded, `KafkaConsumerActor` stops and `processErrors` signals all stages, thus stream fails.

ennru:
Sounds viable. But that should be implemented in another actor so that it can be amended with another dispatcher and easily replaced with other strategies.

How would it work for producers?

[...]

PR #831, opened 2019-06-19
Connection checker

tayvs:
## Purpose
Kafka connection checker
Checking connection liveness using listTopics. On first failure start exponentially retrying. When maxRetries exceeded then KafkaConsumerActor and dependent streams will fail with 'KafkaConnectionFailed' error.

## References

Feature, that described in https://github.com/akka/alpakka-kafka/issues/674


tayvs:
Review by @2m @ennru

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,30 @@ akka.kafka.consumer {
   # Interval for checking that transaction was completed before closing the consumer.
   # Used in the transactional flow for exactly-once-semantics processing.
   eos-draining-check-interval = 30ms
+
+  # Settings for connection checker. Connection checking using listTopics request with timeout
```
ennru:
```suggestion
  # Settings for checking the connection to the Kafka broker. Connection checking uses `listTopics` requests with the timeout
```

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,30 @@ akka.kafka.consumer {
...
+  # configurated by akka.kafka.consumer.metadata-request-timeout
```
ennru:
```suggestion
  # configured by `consumer.metadata-request-timeout`
```

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,30 @@ akka.kafka.consumer {
...
+  connection-checker {
+
+    #Flag to turn on connection checker
+    enable = false
+
+    # Fully qualified config path which holds the dispatcher configuration
+    # to be used by the KafkaConnectionChecker. Some blocking may occur.
+    dispatcher = ${akka.kafka.consumer.use-dispatcher}
```
ennru: The connection checker doesn't do any blocking itself.
So the dispatcher setting can be removed.

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,30 @@ akka.kafka.consumer {
...
+
+    # Ammount of attempts that would be perform after first connection failure occurs
+    # Required not negative number
+    max-retries = 3
+
+    # Intreval between listTopics response and new request. Also used as base for exponential retry
+    check-interval = 1s
```
ennru: The default should be larger to align with the other defaults, maybe 15 seconds.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -71,6 +71,12 @@ object ConsumerSettings {
     val offsetForTimesTimeout = config.getDuration("offset-for-times-timeout").asScala
     val metadataRequestTimeout = config.getDuration("metadata-request-timeout").asScala
     val drainingCheckInterval = config.getDuration("eos-draining-check-interval").asScala
+
+    val enablePath = s"${KafkaConnectionCheckerSettings.configPath}.enable"
+    val kafkaConnectionCheckerConfig =
+      if (config.hasPath(enablePath) && config.getBoolean(enablePath)) Some(KafkaConnectionCheckerSettings(config))
```
ennru: You need to use `config.getConfig(KafkaConnectionCheckerSettings.configPath)`, right?

core/src/main/scala/akka/kafka/KafkaConnectionCheckerSettings.scala
```diff
@@ -0,0 +1,82 @@
...
+  def withCheckInterval(checkInterval: FiniteDuration): KafkaConnectionCheckerSettings =
+    copy(checkInterval = checkInterval)
+
+  /** Java API */
+  def withCheckInterval(checkInterval: java.time.Duration): KafkaConnectionCheckerSettings =
+    copy(checkInterval = checkInterval.asScala)
+
+  def getMaxRetries: Int = maxRetries
+  def getCheckInterval: java.time.Duration = checkInterval.asJava
+  def getFactor: Double = factor
```
ennru: The Java getters are not required.

core/src/main/scala/akka/kafka/internal/KafkaConnectionChecker.scala
```diff
@@ -0,0 +1,85 @@
...
+
+@InternalApi private class KafkaConnectionChecker(config: KafkaConnectionCheckerSettings)
+    extends Actor
+    with ActorLogging
+    with Timers {
+
+  import KafkaConnectionChecker.Internal._
+  import config.{dispatcher => _, _}
+
+  startRegularTimer()
```
ennru: Start the timer in `preStart`.

core/src/main/scala/akka/kafka/internal/KafkaConnectionChecker.scala
```diff
@@ -0,0 +1,85 @@
...
+   */
+  def startBackoffTimer(previousInterval: FiniteDuration): FiniteDuration = {
+    val backoffCheckInterval = (previousInterval * factor).asInstanceOf[FiniteDuration]
+    timers.startSingleTimer(BackoffCheck, CheckConnection, backoffCheckInterval)
+    backoffCheckInterval
+  }
+
+}
+
+object KafkaConnectionChecker {
```
ennru: This should be marked `@InternalApi`.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -199,7 +206,8 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     val offsetForTimesTimeout: FiniteDuration,
     val metadataRequestTimeout: FiniteDuration,
     val drainingCheckInterval: FiniteDuration,
-    val consumerFactory: ConsumerSettings[K, V] => Consumer[K, V]
+    val consumerFactory: ConsumerSettings[K, V] => Consumer[K, V],
+    val kafkaConnectionCheckerSettings: Option[KafkaConnectionCheckerSettings]
```
ennru: Instead of the none option, use an "empty" implementation which doesn't do the checking.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -234,7 +242,45 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     offsetForTimesTimeout = 5.seconds,
     metadataRequestTimeout = 5.seconds,
     drainingCheckInterval = 30.millis,
-    consumerFactory = ConsumerSettings.createKafkaConsumer
+    consumerFactory = ConsumerSettings.createKafkaConsumer,
+    kafkaConnectionCheckerSettings = None
+  )
+
+  def this(properties: Map[String, String],
```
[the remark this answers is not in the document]
tayvs: Compatibility check failing without this constructor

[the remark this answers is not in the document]
tayvs: Compatibility checking failed without this constructor

core/src/main/scala/akka/kafka/ConsumerSettings.scala
tayvs: You mean some kind of config with single value enable:false?

core/src/main/scala/akka/kafka/KafkaConnectionCheckerSettings.scala
```diff
@@ -0,0 +1,82 @@
...
+
+package akka.kafka
+
+import akka.actor.ActorSystem
+import akka.util.JavaDurationConverters._
+import com.typesafe.config.Config
+
+import scala.concurrent.duration.FiniteDuration
+
+class KafkaConnectionCheckerSettings private[kafka] (val maxRetries: Int,
```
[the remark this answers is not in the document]
tayvs: ok, then should I rename 'KafkaConnectionChecker' to 'ConnectionChecker' too?

core/src/main/scala/akka/kafka/KafkaConnectionCheckerSettings.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  def apply(config: Config): KafkaConnectionCheckerSettings = {
+    val retries = config.getInt("max-retries")
+    val factor = config.getDouble("backoff-factor")
+    val checkInterval = config.getDuration("check-interval").asScala
+    val dispatcher = config.getString("dispatcher")
+    apply(retries, checkInterval, factor, dispatcher)
+  }
+
+  def apply(system: ActorSystem): KafkaConnectionCheckerSettings =
```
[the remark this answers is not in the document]
tayvs: Not sure. I just implement similar to 'ConsumerSettings' Api

core/src/main/scala/akka/kafka/KafkaConnectionCheckerSettings.scala
tayvs: remove them all or just 'getFactor' and 'getMaxRetries'?

core/src/main/scala/akka/kafka/KafkaConnectionCheckerSettings.scala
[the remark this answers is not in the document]
ennru: Yes, makes sense.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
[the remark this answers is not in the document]
ennru: Add an exclusions file as described in https://github.com/akka/alpakka/blob/master/CONTRIBUTING.md#binary-compatibility-mima

core/src/main/scala/akka/kafka/KafkaConnectionCheckerSettings.scala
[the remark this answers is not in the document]
ennru: Remove it.

core/src/main/scala/akka/kafka/KafkaConnectionCheckerSettings.scala
ennru: Remove all of them.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
ennru: Yes, add enabled to the settings and test it to decide if the actor is started. The default value is `false` already.

build.sbt
```diff
@@ -247,6 +247,7 @@ lazy val tests = project
         "SerializationSpec.scala" ||
         "PartitionExamples.scala" ||
         "TransactionsExample.scala" ||
+        "KafkaConnectionCheckerSpec.scala" ||
```
tayvs: file was renamed

ennru:
FYI: There seem to be major issues around scalafmt downloads on Travis.

tayvs:
> FYI: There seem to be major issues around scalafmt downloads on Travis.

I'm already notice. Is my changes

tayvs:
> FYI: There seem to be major issues around scalafmt downloads on Travis.

Is this issue result of my changes?

ennru:
No, it is not. We don't know why what is causing those problems.

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,26 @@ akka.kafka.consumer {
...
   eos-draining-check-interval = 30ms
+
+  # Settings for checking the connection to the Kafka broker. Connection checking uses `listTopics` requests with the timeout
+  # configured by `consumer.metadata-request-timeout`
+  connection-checker {
+
+    #Flag to turn on connection checker
+    enable = false
+
+    # Ammount of attempts that would be perform after first connection failure occurs
```
ennru:
```suggestion
    # Amount of attempts to be performed after a first connection failure occurs
```

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,26 @@ akka.kafka.consumer {
...
+    # Required not negative number
```
ennru:
```suggestion
    # Required, non-negative integer
```

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,26 @@ akka.kafka.consumer {
...
+    max-retries = 3
+
+    # Intreval between listTopics response and new request. Also used as base for exponential retry
```
ennru:
```suggestion
    # Interval for the connection check. Used as the base for exponential retry.
```

core/src/main/resources/reference.conf
```diff
@@ -100,6 +100,26 @@ akka.kafka.consumer {
...
+    check-interval = 15s
+
+    # Check interval multiplier for backoff interval
+    # Required finite positive number
```
ennru:
```suggestion
    # Required, positive number
```

core/src/main/scala/akka/kafka/ConnectionCheckerSettings.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  val configPath: String = "connection-checker"
+  val fullConfigPath: String = ConsumerSettings.configPath + "." + configPath
+
+  /**
+   * Create settings from a configuration with layout `connection-checker`.
+   */
+  def apply(config: Config): ConnectionCheckerSettings = {
+    val enable = if (config.hasPath("enable")) config.getBoolean("enable") else false
+    if (enable) EnabledConnectionCheckerSettings(config) else DisabledConnectionCheckerSettings
```
ennru: Thank you for trying this pattern out. I believe it would be just enough to have one implementation with an `enabled` field to decide if the actor should spin up. This API looks a bit too large for the use-case.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -439,6 +448,20 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
...
+   */
+  def withConnectionChecker(
+      kafkaConnectionCheckerConfig: ConnectionCheckerSettings
+  ): ConsumerSettings[K, V] =
+    copy(connectionCheckerConfig = kafkaConnectionCheckerConfig)
+
+  /**
+   * Disable kafka connection checker
+   */
+  def withoutConnectionChecker: ConsumerSettings[K, V] =
```
ennru: This is not needed.

core/src/main/scala/akka/kafka/internal/ConnectionChecker.scala
```diff
@@ -0,0 +1,84 @@
...
+        context.parent ! KafkaConnectionFailed(te, maxRetries)
+        context.stop(self)
+      } else context.become(backoff(failedAttempts + 1, startBackoffTimer(interval)))
+
+    case Metadata.Topics(Success(_)) =>
+      startRegularTimer()
+      context.become(regular)
+  }
+
+  def startRegularTimer(): Unit = timers.startSingleTimer(RegularCheck, CheckConnection, checkInterval)
```
ennru:
```suggestion
  def startTimer(): Unit = timers.startSingleTimer(RegularCheck, CheckConnection, checkInterval)
```

core/src/main/scala/akka/kafka/ConnectionCheckerSettings.scala
tayvs: I don't understand what you exactly mean. I can't create one settings class because in one case I expect configs and `enabled: true`, in another configs can not be present at all. So I can't create Settings like 
```
ConnectionCheckerSettings(
val enabled: Boolean.
val maxRetries: Int,
val checkInterval: FiniteDuration,
val factor: Double)
```
Or you mean use connectionCheckerEnabled flag instead ConnectionCheckerSettings . And parse configs on actor start. But if config values invalid and ConnectionChecker actor fall with ActorCreationException then ConsumerActor and streams continue work like nothing happens

tayvs: Hmmm.. Maybe first case make sense because we provide default parameters and config must be valid. But I have feeling that someone would create actorSystem using ConfigFactory.parseString and even if they don't want use ConnectionChecker them must provide `connection-checker` configs

ennru: Even the use of `ConfigFactory.parseString` needs to be resolved with `reference.conf`. New required settings can be introduced even in patch releases.

tayvs: @ennru Ok, then reimplement like
```
ConnectionCheckerSettings(
val enabled: Boolean.
val maxRetries: Int,
val checkInterval: FiniteDuration,
val factor: Double)
```
?

ennru: Yes, please.

core/src/main/scala/akka/kafka/ConnectionCheckerSettings.scala
```diff
@@ -0,0 +1,87 @@
...
+object ConnectionCheckerSettings {
+
+  val configPath: String = "connection-checker"
+  val fullConfigPath: String = ConsumerSettings.configPath + "." + configPath
+
+  def apply(enable: Boolean,
+            maxRetries: Int,
+            checkInterval: FiniteDuration,
+            factor: Double): ConnectionCheckerSettings = {
+    require(factor > 0, "Backoff factor for connection checker must be finite positive number")
```
ennru: The validations should move to the class instead.
Remove `enable` from `apply` and `create`. You could add a `val Disabled: ConnectionCheckerSettings = ...` if you think that is useful.

tayvs: good idea

tayvs: good idea

tayvs: I Have `The command "jabba install adopt@~1.8.202-08" failed and exited with 1 during .` error could you please restart build

ennru:
Failure was https://github.com/akka/alpakka-kafka/issues/842

core/src/main/mima-filters/1.0.5.backwards.excludes
```diff
@@ -0,0 +1,3 @@
+# PR #831
```
ennru: Just noticed that the file name is wrong. It should be `1.0.4.backwards.excludes`.

tayvs: done

tayvs: Am I right understand that I need 2 approvals from you and @2m

ennru:
Thank you for this great piece of functionality.

PR #837, opened 2019-06-25
Test case illustrating dropping of fetched data (update)

seglo:
This is an updated `SlowConsumerSpec` of #551 

> I used this test case to investigate #549 it does not assert anything useful but shows the reported misbehaviour.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -248,6 +248,7 @@ lazy val tests = project
         "EmbeddedKafkaSampleSpec.scala" ||
         "TransactionsSpec.scala" ||
         "SerializationSpec.scala" ||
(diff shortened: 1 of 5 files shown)


ennru:
You need to exclude that test from the Scala 2.13 build.

seglo:
PR #1014 will allow us to track various metrics related to Kafka consumer performance (consumer/broker network traffic, consumer throughput, fetch requests, etc.)

seglo:
Dropping of pre-fetched data was fixed upstream in Kafka in https://issues.apache.org/jira/browse/KAFKA-7548 .  Released in 2.4.0.

Issue #832, opened 2019-06-19
Failed IntegrationSpec must signal rebalance events to actor


2m:
```
[info] - must signal rebalance events to actor *** FAILED *** (37 seconds, 646 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 61 times, sleeping 500000000 nanoseconds between each query. (IntegrationSpec.scala:120)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at akka.kafka.scaladsl.IntegrationSpec.$anonfun$new$6(IntegrationSpec.scala:120)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
[info]   at akka.kafka.scaladsl.IntegrationSpec.$anonfun$new$5(IntegrationSpec.scala:52)
```

https://travis-ci.org/akka/alpakka-kafka/jobs/547197965

ennru:
Not all messages are received for some reason
https://travis-ci.org/akka/alpakka-kafka/jobs/580201215#L890

Issue #819, opened 2019-06-03
Use Scala 2.13 for the benchmarks


ennru:
### Short description

The benchmarks should be able to work on Scala 2.13.

### Details

The benchmarks project is excluded from Scala 2.13 compilation for now.

ennru:
Fixed with #844

PR #844, opened 2019-07-03
Get closer to using Scala 2.13 as primary version

ennru:
## Purpose

Build the documentation with Scaladoc 2.13 

## Changes

* set `scalaVersion` to 2.13.0 for the Travis publish documentation job
  * use new scaladoc flags for better links to source and canonical URLs
* check if jabba was cached
* fix? the notorious failure #832 
* Run `IntegrationSpec` with Testcontainers


ennru:
This will enable Scala 2.13 for the benchmarks as well. See #819

Issue #845, opened 2019-07-03
Allow user to commit offsets without handling response when using committable sources


seglo:
### Short description

Allow the user to commit offsets back to Kafka asynchronously without waiting for a response.

### Details

The `Committer.batchFlow` will use `mayAsync` to backpressure offset commit callbacks.  When we call `commitScalaDsl` we will use the Ask pattern to ask the `KafkaConsumerActor` to commit the offset map immediately (using `KafkaConsumer.commitAsync`) and reply with a response when the commit response is handled during a subsequent poll.  This implementation has the benefit of making sure we lose as few messages as possible in the event of a failure while using at least once message delivery, but it can also considerably slow down the whole consuming stream.  A compromise would be to commit and not wait for a reply which would allow the stream to process records faster, with the tradeoff of potentially reprocessing more records after a failure or rebalance.

It would be nice if there was a way to "fire and forget" the commit so we don't back-pressure in the `Committer.batchFlow` waiting for replies.  We also wouldn't create the overhead of additional callback threads.

seglo:
FYI I'm going to pick this up!

ennru:
Implemented with #883

Issue #841, opened 2019-07-01
Committing: explore different implementation


ennru:
# Committing: explore different implementation

The current Alpakka Kafka implementation for committing offsets to Kafka relies on collecting offsets for processed messages in the stream. In case of a change in the consumer group, Kafka revokes and assigns partitions which implies that the consumer can't commit any further offsets to partitions that are revoked (and not re-assigned). This means already in-stream collected offsets can't be committed and those messages will be re-emitted to the consumer that gets that partition assigned by Kafka.

### Committing on partition revoke

The experiment of introducing a `PartitionAssignmentHandler` in [#761](https://github.com/akka/alpakka-kafka/pull/761) explores how such a handler could be used to implement committing outside of the current Alpakka Kafka handling. A key idea is to leverage the callback from Kafka when partitions are revoked to commit any offsets which are known to have been processed successfully.

If committing can be handled just before a partition gets revoked, much fewer messages will get reprocessed during consumer rebalancing.

### Potential improvements

1. Fewer messages get reprocessed on rebalance
1. It would become possible to allow for some out-of-order processing if the committing logic keeps track of holes in the offset list and just commits segments that are done.

### Potential negative changes

1. More actor messaging as every message sends its commit to the collector
1. Larger change in the Alpakka Kafka architecture

### Necessary changes

1. New "Commit collector" actor as child of `KafkaConsumerActor`
1. Change `commitScaladsl` to send the offset to the "Commit collector" (fire and forget?)
1. Schedule regular commits from "Commit collector" (using committer `max-interval` and `max-batch`)
1. Commit any uncommitted offsets from the partition revoke by asking the "Commit collector"
1. Remove batching from the "Committer"

ennru:
#862 and #868 implement the aggregation of commits.
I'll create a new issue to discuss dropping of commit back-pressure.

PR #846, opened 2019-07-04
New DSL for offset commit with no callback when using a committable source

seglo:
## Purpose

Allow the user to commit offsets back to Kafka asynchronously without waiting for a response.

## References

Issue #845

## Changes

* Add more options to the `Committer` and `Committable` DSLs to commit offsets without waiting for a callback

## Background Context

The `Committer.batchFlow` will use `mayAsync` to backpressure offset commit callbacks.  When we call `commitScalaDsl` we will use the Ask pattern to ask the `KafkaConsumerActor` to commit the offset map immediately (using `KafkaConsumer.commitAsync`) and reply with a response when the commit response is handled during a subsequent poll.  This implementation has the benefit of making sure we lose as few messages as possible in the event of a failure while using at least once message delivery, but it can also considerably slow down the whole consuming stream.  A compromise would be to commit and not wait for a reply which would allow the stream to process records faster, with the tradeoff of potentially reprocessing more records after a failure or rebalance.

This implementation provides the user with a DSL to "fire and forget" the commit so we don't back-pressure in the `Committer.flow` waiting for replies.  We also don't create the overhead of additional unnecessary callback threads, though the Kafka consumer will create threads internally for its default callback which logs errors when they occur.

I'm on the fence about calling the implementation "WithNoCallback".  I also considered calling it "Async", but since the original impl. is actually async as well this didn't seem appropriate.  I'm open to suggestions.

I can add doc updates once an initial review has been done.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -63,7 +63,11 @@ import scala.util.control.NonFatal
     final case class RequestMessages(requestId: Int, topics: Set[TopicPartition])
         extends NoSerializationVerificationNeeded
     val Stop = akka.kafka.KafkaConsumerActor.Stop
-    final case class Commit(offsets: Map[TopicPartition, OffsetAndMetadata]) extends NoSerializationVerificationNeeded
+    sealed trait Commit extends NoSerializationVerificationNeeded {
+      def offsets: Map[TopicPartition, OffsetAndMetadata]
+    }
+    final case class CommitWithCallback(offsets: Map[TopicPartition, OffsetAndMetadata]) extends Commit
+    final case class CommitWithNoCallback(offsets: Map[TopicPartition, OffsetAndMetadata]) extends Commit
     //responses
     final case class Assigned(partition: List[TopicPartition]) extends NoSerializationVerificationNeeded
     final case class Revoked(partition: List[TopicPartition]) extends NoSerializationVerificationNeeded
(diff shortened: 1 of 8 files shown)


core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -518,16 +528,27 @@ import scala.util.control.NonFatal
...
+
+  private def commitWithNoCallback(commitMap: Map[TopicPartition, OffsetAndMetadata]): Unit = {
+    // passing a `null` callback will force the consumer to use a default which will log errors, but nothing else
+    consumer.commitAsync(commitMap.asJava, null)
+    pollAfterCommit()
+  }
+
+  /**
+   * When many requestors, e.g. many partitions with committablePartitionedSource the
+   * performance is much by collecting more requests/commits before performing the poll.
```
agemooij: "much slower"? "much faster"? "much better"? "much worse"?

ennru:
I wonder if this is the best cure to the problem. I believe my idea in #841 will be able to provide both parts of this but would be a huge change in the Alpakka Kafka committing architecture.

As an intermediate step, I'd like to add the scenario this tries to cure into test cases, possibly in the benchmarks.
We should try to come up with a solution that aggregates the committed offsets a bit more depending on the polling, eg. make sure we commit only once within every poll cycle. That would lower the pressure on the Kafka client for high-throughput streams.

ennru:
Aggregating commits could use the same commit stash as the rebalance stashing does.

seglo:
Closing this in favour of a more fundamental chance to how committing could be done, with fire and forget possibly as a default.  #850

Issue #859, opened 2019-08-12
Static Membership support.


def1ne:
### Short description

Add support of static membership in consumer.

### Details

Static membership was introduced In Kafka 2.3.0. It allows to evade rebalances if consumer group is static.

To support this feature we need to update Kafka client to 2.3.0 and add `withGroupInstanceId` function in `ConsumerSettings`.

### Links
[static membership description](https://kafka.apache.org/documentation/#static_membership)
[group.instance.id](http://kafka.apache.org/23/javadoc/org/apache/kafka/clients/consumer/ConsumerConfig.html#GROUP_INSTANCE_ID_CONFIG)
[KIP-345: Introduce static membership protocol to reduce consumer rebalances](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances)

ennru:
Thank you for this suggestion. Please feel free to suggest the necessary changes in a Pull Request.

jewertow:
Since PR is merged this issue should be closed.

ennru:
Thanks for the ping.
Fixed with #861

PR #861, opened 2019-08-13
ConsumerSetting: Add group instance id setter

def1ne:
## Purpose

Update Kafka dependency and add method `withGroupInstanceId` to `ConsumerSettings`.

## References

References #859


ennru:
Thank you for this PR.
As it changes the Kafka client library version, we won't include it in patch releases, but need to defer it until version 2.0.

Are you aware that you can achieve this feature by overriding the Kafka client library version yourself and setting the property with the `withProperty` method?

def1ne:
We updated kafka client and set property through withProperty. All works fine.

ennru:
I've rebased this PR on master which will go into Alpakka Kafka 2.0.0-M1.

ennru:
Thank you for your contribution!

Issue #849, opened 2019-07-05
Commit only once per poll-interval


ennru:
### Short description

Do at most one commit per poll-cycle to lower the commit strain on the Kafka client.

### Details

For high-throughput consumers, Alpakka Kafka may fire off commits much more often than it polls. The commit callbacks won't be called before a poll is made.

The default poll-intervall is 50 ms.

Instead the offsets to be committed should be collected in the `KafkaConsumerActor` in the same way as they are for `rebalanceInProgress`. The collected commits can be passed to Kafka just before the `poll` in the same way as `refreshOffsets` are sent there.

#### Potential improvements

This puts an upper limit on how often commit is called, thus avoids strain on the Kafka consumer.

#### Potential negative changes

As the `Committer` sends the commits by default with a parallelism of 1, this internal batching will make a high-throughput stream with default committer settings seemingly slow as backpressure is applied until the commit is reported to be done.

The committer settings still need to be balanced with the expected throughput. 

### References

#841 
#845 & #846

ennru:
This is now implemented with a bit differently with #862 and #868.

PR #862, opened 2019-08-14
Aggregate offsets and commit before poll

ennru:
## Purpose

Reduce calls to `consumer.commitAsync` by aggregating commits in the actor and sending them just before polling.

## References

Idea explained in #849
Implementation partly stolen from #851 

## Changes

The almost unchanged tests indicate that this doesn't change the behaviour a lot.

This change may make a commit wait until the next poll (default 50 ms). To fully make use of this, the committer defaults should be changed.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -205,6 +205,8 @@ import scala.util.control.NonFatal
   private val positionTimeout = settings.getPositionTimeout
 
   private var requests = Map.empty[ActorRef, RequestMessages]
+
+  /** ActorRefs to all stages that requested messages from this actor (removed on their termination). */
   private var requestors = Set.empty[ActorRef]
   private var consumer: Consumer[K, V] = _
   private var subscriptions = Set.empty[SubscriptionRequest]
@@ -224,7 +226,7 @@ import scala.util.control.NonFatal
   /**
    * Aggregates commit offsets until the next poll.
    */
-  private var commitStash = Map.empty[TopicPartition, OffsetAndMetadata]
+  private var commitAggregate = Map.empty[TopicPartition, OffsetAndMetadata]
 
   /**
    * Keeps commit senders that need a reply once stashed commits are made.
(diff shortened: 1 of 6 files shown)


core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -270,14 +272,9 @@ import scala.util.control.NonFatal
...
-    case Commit(offsets) if rebalanceInProgress =>
-      rebalanceCommitStash ++= offsets
-      rebalanceCommitSenders = rebalanceCommitSenders :+ sender()
-
     case Commit(offsets) =>
-      commitRefreshing.add(offsets)
-      val replyTo = sender()
-      commit(offsets, replyTo ! _)
+      commitAggregate ++= offsets
+      commitSenders = commitSenders :+ sender()
```
ennru: A thing to consider would be to request an extra poll by setting up a "delayed poll in flight" as the `RequestMessages` do.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -463,14 +461,20 @@ import scala.util.control.NonFatal
         log.error(e, "Exception when polling from consumer, stopping actor: {}", e.toString)
         context.stop(self)
     }
-    checkRebalanceState(initialRebalanceInProgress)
```
ennru: This could possibly stay. It applies if commits arrive during a rebalance, but they'll be handled with the next poll.

seglo: The `checkRebalanceState` method seems redundant with `commitAggregatedOffsets`.  Also, I don't see how we could have any pending commits to make anyway since the actor is currently handling the poll message.

ennru: Yes, I came to that conclusion, as well.

ennru:
Pushed an update so that the order in which the offsets are received doesn't matter anymore. With this we can increase the parallelism in the Committer and switch to `mapAsyncUnordered`.

ennru:
Failure was #832

PR #865, opened 2019-08-19
Illustrate how max.poll.records affects buffer

ennru:
## Purpose

Once messages are fetched from the Kafka broker even a rebalance doesn't drop them from Alpakka Kafka's internal buffer. That buffer can be reduced by setting the `max.poll.records` to some value lower than the 500 messages.

This test case illustrates this.

## Background Context

This behaviour leads to duplicated messages to different Alpakka Kafka streams after a rebalance. It would be interesting to explore ways to minimise this behaviour without lowering the `max.poll.records`.


ennru:
Tests fail as the cross Scala running leaves messages from the first run on the topic.

ennru:
Issue described in https://github.com/akka/alpakka-kafka/issues/872

PR #866, opened 2019-08-19
Extra poll after receiving commit data

ennru:
## Purpose

Trigger a poll after receiving commit data so that the commit will be executed earlier than the scheduled poll, but still allowing more commits to arrive before the poll is made.

## References

Follow-up on #862 as the "at-most-once" benchmarks showed very slow consumption when committing one message at a time.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -286,7 +286,7 @@ import scala.util.control.NonFatal
       // prepending, as later received offsets most likely are higher
       commitMaps = offsets :: commitMaps
       commitSenders = commitSenders :+ sender()
-      requestExtraPoll()
+      requestDelayedPoll()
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -416,6 +411,19 @@ import scala.util.control.NonFatal
...
 
+  /**
+   * Sends an extra `Poll` request to self.
+   * Enqueing an extra poll via the actor mailbox allows other requests to be handled
+   * before the actual poll is executed.
+   *  When many requestors, e.g. many partitions with committablePartitionedSource the
+   * performance is much improved by collecting more requests/commits before performing the poll.
+   */
+  private def requestExtraPoll(): Unit =
+    if (!delayedPollInFlight) {
```
2m: I think the flag should match the method, so `extraPollInFlight` or `requestDelayedPoll()`.

PR #867, opened 2019-08-19
Commit refreshing: avoid extra work

ennru:
## Purpose

Instead of sending a message to self, the committed offset tracking can be updated directly. Furthermore, converting to a Scala map can be avoided if commit refreshing isn't even enabled.


ennru:
Failure was https://github.com/akka/alpakka-kafka/issues/832

PR #870, opened 2019-08-23
Offset batch: remove intermediary structure 

ennru:
## Purpose

Convert the `Map[GroupTopicPartition, OffsetAndMetadata]` directly into a `Map[TopicPartition, OffsetAndMetadata]` to be sent to the consumer actor.

## Changes

* Move the location where the `offset + 1` is applied
* Change `toString` back to just show the offsets

## Background Context

Before sending offsets in `KafkaAsyncConsumerCommitterRef` the map was first transformed to a `Seq[PartitionOffsetMetadata]` (unpacking the `OffsetWithMetaData`) and than back to a `Map[TopicPartition, OffsetAndMetadata]` in the `commit` method.
By mapping directly less intermediary objects are created. The `OffsetAndMetadata` can be kept, but the `offset + 1` needed to be applied when building the batch in `CommittableOffsetBatchImpl` instead. The public `offsets` method compensates for that.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/CommittableSources.scala
+++ b/core/src/main/scala/akka/kafka/internal/CommittableSources.scala
@@ -7,12 +7,7 @@ package akka.kafka.internal
 
 import akka.actor.ActorRef
 import akka.annotation.InternalApi
-import akka.kafka.ConsumerMessage.{
-  CommittableMessage,
-  CommittableOffset,
-  CommittableOffsetBatch,
-  PartitionOffsetMetadata
-}
+import akka.kafka.ConsumerMessage.{CommittableMessage, CommittableOffset, CommittableOffsetBatch, GroupTopicPartition}
 import akka.kafka._
 import akka.kafka.internal.KafkaConsumerActor.Internal.{Commit, CommitSingle}
 import akka.kafka.scaladsl.Consumer.Control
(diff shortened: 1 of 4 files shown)


core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -180,7 +180,7 @@ private[kafka] final class CommittableOffsetBatchImpl(
     }
 
     val newOffsets =
-      offsetsAndMetadata.updated(key, new OffsetAndMetadata(committableOffset.partitionOffset.offset, metadata))
+      offsetsAndMetadata.updated(key, new OffsetAndMetadata(committableOffset.partitionOffset.offset + 1L, metadata))
```
2m: I do not see immediately, why the change in offset increment and decrement above.

ennru: The `OffsetAndMetadata` is now passed through as-is. Before it was constructed, decomposed and constructed again with the `+ 1`.

core/src/main/scala/akka/kafka/internal/CommittableSources.scala
```diff
@@ -139,26 +133,20 @@ private[kafka] class KafkaAsyncConsumerCommitterRef(consumerActor: ActorRef, com
       )
     )
 
-  private def commit(offsets: immutable.Seq[PartitionOffsetMetadata]): Future[Done] = {
-    val offsetsMap: Map[TopicPartition, OffsetAndMetadata] = offsets.map { offset =>
-      new TopicPartition(offset.key.topic, offset.key.partition) ->
-      new OffsetAndMetadata(offset.offset + 1, offset.metadata)
```
ennru: Here was the `+ 1` before.

PR #876, opened 2019-08-28
Add producer benchmark with partitions

ennru:
## Purpose

Add another producer benchmark that writes to 8 partitions.
Make partition use more explicit for all benchmarks.

## Background Context

Producing with Alpakka Kafka should be able to perform better.

### DIFF
--- a/benchmarks/src/it/scala/akka/kafka/benchmarks/Benchmarks.scala
+++ b/benchmarks/src/it/scala/akka/kafka/benchmarks/Benchmarks.scala
@@ -139,7 +139,7 @@ class ApacheKafkaPlainProducer extends BenchmarksBase() {
   }
 
   it should "bench with normal messages written to 8 partitions" in {
-    val cmd = RunTestCommand(prefix + "-normal-msg", bootstrapServers, 2000 * factor, 5000, numberOfPartitions = 8)
+    val cmd = RunTestCommand(prefix + "-normal-msg-8-partitions", bootstrapServers, 2000 * factor, 5000, numberOfPartitions = 8)
     runPerfTest(cmd, KafkaProducerFixtures.initializedProducer(cmd), KafkaProducerBenchmarks.plainFlow)
   }
 }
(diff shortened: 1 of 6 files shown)


benchmarks/src/it/scala/akka/kafka/benchmarks/Benchmarks.scala
```diff
@@ -125,25 +126,39 @@ class AlpakkaKafkaAtMostOnceConsumer extends BenchmarksBase() {
...
   }
 
   it should "bench with normal messages" in {
-    val cmd = RunTestCommand("apache-kafka-plain-producer-normal-msg", bootstrapServers, 2000 * 1000, 5000)
+    val cmd = RunTestCommand(prefix + "-normal-msg", bootstrapServers, 2000 * factor, 5000)
+    runPerfTest(cmd, KafkaProducerFixtures.initializedProducer(cmd), KafkaProducerBenchmarks.plainFlow)
+  }
+
+  it should "bench with normal messages written to 8 partitions" in {
+    val cmd = RunTestCommand(prefix + "-normal-msg", bootstrapServers, 2000 * factor, 5000, numberOfPartitions = 8)
```
2m: The name of the command should be distinct from the one above. Add `-8-partitions`. This is important, because this name is used in the result output file name.

benchmarks/src/it/scala/akka/kafka/benchmarks/Benchmarks.scala
```diff
@@ -125,25 +126,39 @@ class AlpakkaKafkaAtMostOnceConsumer extends BenchmarksBase() {
...
   }
 
   it should "bench with normal messages" in {
-    val cmd = RunTestCommand("alpakka-kafka-plain-producer-normal-msg", bootstrapServers, 2000 * 1000, 5000)
+    val cmd = RunTestCommand(prefix + "-normal-msg", bootstrapServers, 2000 * factor, 5000)
+    runPerfTest(cmd, ReactiveKafkaProducerFixtures.flowFixture(cmd), ReactiveKafkaProducerBenchmarks.plainFlow)
+  }
+
+  it should "bench with normal messages written to 8 partitions" in {
+    val cmd = RunTestCommand(prefix + "-normal-msg", bootstrapServers, 2000 * factor, 5000, numberOfPartitions = 8)
```
2m: Same here, add `-8-partitions` to the name.

benchmarks/src/it/scala/akka/kafka/benchmarks/Benchmarks.scala
ennru: Good catch!

PR #878, opened 2019-08-30
Update to Lightbend Community Code of Conduct

TimMoore:
https://www.lightbend.com/conduct


ennru:
We'll need to fix links in the published documentation before merging this to avoid 404s.

TimMoore:
Good point. I can update the content rather than removing the file.

TimMoore:
Updated

PR #879, opened 2019-08-30
IntegrationSpec: simplify counting to understand failures

ennru:
Trying to improve on #832


ennru:
It seems to be a Heisenbug. Now that I try to inspect it, it doesn't show.
But I'd say it is worth merging this anyway so we can inspect it if it shows again.

Issue #882, opened 2019-09-02
Promote non-backpressuring committing to official API


ennru:
The new support to not backpressure for commit replies should be promoted to official API or deprecated and be removed again.

This is a discussion issue for reference from the source code.

ennru:
Our initial tests do not show an improvement in throughput for non-backpressureing commits.

PR #883, opened 2019-09-02
Introduce API for committing without backpressure (Alternative 2)

ennru:
## Purpose

Explore an API to send commits to Kafka without waiting for an acknowledgement.

This PR is an alternative to #874. This alternative uses a new flag in `CommitterSettings` to enable committing without backpressure on the existing `Committer` flows.

## References

#874 offers different factory methods to use committing without backpressure. 
See #845 

## Changes

* Add `CommitterSettings.delivery` (`CommitDelivery`)
* Add internal `CommitWithoutReply` message
* Add internal `CommittableBatch.commitAndForget`

Unrelated:
* The Raw Kafka benchmarks now commit to the correct partition. 

## Background Context

It is not obvious why one would backpressure reading from a consumer when commits are not acknowledged by Kafka at the same rate. This new API opens for a new way of committing without waiting for the commit to get a reply from the Kafka broker.

### DIFF
--- a/core/src/main/mima-filters/1.0.5.backwards.excludes
+++ b/core/src/main/mima-filters/1.0.5.backwards.excludes
@@ -1,0 +1,3 @@
+# PR #883 Committing without backpressure
+# https://github.com/akka/alpakka-kafka/pull/883
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.CommitterSettings.this")
--- a/core/src/main/mima-filters/1.0.5.backwards.excludes
+++ b/core/src/main/mima-filters/1.0.5.backwards.excludes
@@ -1,3 +1,0 @@
-# PR #883 Committing without backpressure
-# https://github.com/akka/alpakka-kafka/pull/883
-ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.CommitterSettings.this")
--- a/core/src/main/mima-filters/1.0.x.backwards.excludes
+++ b/core/src/main/mima-filters/1.0.x.backwards.excludes
@@ -1,2 +1,5 @@
 # always exclude changes to the internal APIs
 ProblemFilters.exclude[Problem]("akka.kafka.internal.*")
+
+# PR #874 committing without backpressure
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.tellCommit")
--- a/core/src/main/mima-filters/1.0.x.backwards.excludes
+++ b/core/src/main/mima-filters/1.0.x.backwards.excludes
@@ -2,4 +2,6 @@
 ProblemFilters.exclude[Problem]("akka.kafka.internal.*")
 
 # PR #883 committing without backpressure
+# https://github.com/akka/alpakka-kafka/pull/883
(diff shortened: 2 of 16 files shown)


ennru:
This is the difference between this PR and #874: 

https://github.com/ennru/alpakka-kafka/compare/ennru:commit-and-forget...ennru:commit-and-forget-settings

core/src/main/scala/akka/kafka/CommitterSettings.scala
```diff
@@ -7,11 +7,51 @@ package akka.kafka
...
+ * Selects how the stream delivers commits to the internal actor.
+ */
+@ApiMayChange(issue = "https://github.com/akka/alpakka-kafka/issues/882")
+object CommitDelivery {
+
+  /**
+   * Expect replies for commits, and backpressure the stream if replies do not
+   * arrive.
+   */
+  case object Ask extends CommitDelivery
```
seglo: Looks good.  I don't want to get hung up on the naming, but tell/ask does rely on the users knowledge of Akka message delivery to understand.  The scaladoc comments are descriptive though so I think it's good as-is.

It may be worth noting that there are several Kafka configurations that use the terminology `acks`.  In the producer client properties it's used for acknowledgements of message receipt across all replicas.  In the broker there's a setting `offsets.commit.required.acks`.  TBH I'm not really sure what it's used for.

ennru: I'm open to naming suggestions. I believe `ack` interferes with how things are called in Kafka. 
Maybe `ExpectResponses` and `SendAndForget` are better? But that's is incorporated in the Akka Actor inspired names, of course.

core/src/main/mima-filters/1.0.x.backwards.excludes
```diff
@@ -1,2 +1,5 @@
 # always exclude changes to the internal APIs
 ProblemFilters.exclude[Problem]("akka.kafka.internal.*")
+
+# PR #874 committing without backpressure
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.tellCommit")
```
ennru: This should go.

core/src/main/scala/akka/kafka/CommitterSettings.scala
ennru: We agreed on `WaitForAck` and `SendAndForget`.

ennru:
I opened #882 as follow up to decide about this new feature's future, once it is merged.

core/src/main/mima-filters/1.0.x.backwards.excludes
ennru: No, I misread my own code... This has to stay.

core/src/main/mima-filters/1.0.5.backwards.excludes
```diff
@@ -0,0 +1,3 @@
+# PR #883 Committing without backpressure
+# https://github.com/akka/alpakka-kafka/pull/883
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.CommitterSettings.this")
```
2m: The mima exclusion files can be merged to one.

PR #898, opened 2019-09-06
Benchmarks: Create topics with data only once per run

ennru:
## Purpose

To shorten the time it takes to run the benchmarks, the different topics are created and filled with data only once. If the topic exists, the next spec reads the existing data.

## Background Context

Especially the tests with 5000 bytes messages require a lot of disk space during the benchmarks.


ennru:
The build times of the benchmarks were cut by a third with this change.

Issue #497, opened 2018-06-07
Add wrapper for Metadata requests


ennru:
The new feature to request Kafka Metadata from the internal `KafkaConsumerActor` (see #491) would be even nicer with thin wrappers (javadsl and scaladsl) hiding the `ask` calls and mapping to the correct response types.

2m:
The wrapper should also have Scala and Java DSLs with appropriate conversions between Java and Scala types and collections.

jewertow:
I will take care of it.

jewertow:
I started working on it but I wonder what return types do expect from the wrapper, e.g. in the case of `EndOffsets` the wrapper should return `Future[Long]` or `Future[Option[Long]]`?

ennru:
With the current `GetEndOffsets(Set[TopicPartition])` the response is `EndOffsets(Map[TopicPartition, Long])`. So I'd expect `Future[Map[TopicPartition, Long]]`.

ennru:
Implemented in #497 by @jewertow

PR #886, opened 2019-09-02
WIP: proposal to implement a subscription factory with relative offset.

jewertow:
## Purpose

Introduction of a subscription factory with relative offset definitions.

## References

See #538

## Changes

- Add `RelativeOffsetSubscriptions.assignmentForLastN`

## Background Context

There has been a need to query certain Kafka topics with relative offsets, for example last offset, or last offset minus 100, or last offset minus 10 minutes.

PR #900, opened 2019-09-08
Add wrapper for Metadata requests #497

jewertow:
## Purpose

Introduction of a wrapper for `Metadata` requests.

## References

See #497 

## Changes

- Add `MetadataClient`

## Background Context

The new feature to request Kafka Metadata from the internal KafkaConsumerActor (see #491) would be even nicer with thin wrappers (javadsl and scaladsl) hiding the ask calls and mapping to the correct response types.


jewertow:
I think this feature will be useful for #886, so I would like to implement it first. At the moment I have prepared a draft of the target implementation. I need an opinion whether this is a good way or not. If this is a good approach, I will continue this way. Otherwise, please give me some hints.

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
```diff
@@ -0,0 +1,49 @@
...
+import akka.kafka.Metadata.{BeginningOffsets, GetBeginningOffsets}
+import akka.pattern.ask
+import akka.util.Timeout
+import org.apache.kafka.common.TopicPartition
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object MetadataClient {
+
+  private val consumerActors = scala.collection.mutable.Map[ConsumerSettings[Any, Any], ActorRef]()
```
ennru: It should be left to the API user to create and track the consumer actors. The Kafka client runs a number of threads that conscious users might want to avoid.

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def getBeginningOffsets[K, V](
+      consumerSettings: ConsumerSettings[K, V],
+      partitions: Set[TopicPartition],
+      timeout: Timeout
+  )(implicit system: ActorSystem, ec: ExecutionContext): Future[Map[TopicPartition, Long]] = {
```
ennru: The `ask` and `map` operations are safe to be executed on the ActorSystem's dispatcher, you can get away without the separate `ExecutionContext` here.

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
jewertow: I removed it. Please look at it again and tell me if this is a right way.

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
```diff
@@ -0,0 +1,34 @@
...
+object MetadataClient {
+
+  def getBeginningOffsets(
+      consumerActor: ActorRef,
+      partitions: Set[TopicPartition],
+      timeout: Timeout
+  )(implicit ec: ExecutionContext): Future[Map[TopicPartition, Long]] =
+    (consumerActor ? GetBeginningOffsets(partitions))(timeout)
+      .mapTo[BeginningOffsets]
+      .map(_.response.get)
```
ennru: This will throw an exception if there was an error and fail the future that way. Instead, you should `transform` the `Future`. To be Scala 2.11 compatible you need to do it like this 
```
    .flatMap {
      case Success(res) => Future.successful(res)
      case Failure(e) => Future.failed(e)
    }(ExecutionContexts.sameThreadExecutionContext)
```
With this it is OK to use Akka's `sameThreadExecutionContext` to run the `flatMap` on.

tests/src/test/scala/akka/kafka/scaladsl/MetadataClientSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+    "fetch beginning offsets for given partitions" in assertAllStagesStopped {
+      val topic1 = createTopic(1)
+      val group1 = createGroupId(1)
+      val partition0 = new TopicPartition(topic1, 0)
+      val consumerSettings = consumerDefaults.withGroupId(group1)
+      val consumerActor = system.actorOf(KafkaConsumerActor.props(consumerSettings))
+
+      val beginningOffsetsFuture = MetadataClient
+        .getBeginningOffsets(consumerActor, Set(partition0), 1 seconds)
+      val beginningOffsets = Await.result(beginningOffsetsFuture, 1 seconds)
```
ennru: Please add some failing call eg. by using an illegal topic name.
You may mix in Scalatest's `ScalaFutures` and use the `.futureValue` instead of `Await`.

tests/src/test/java/docs/javadsl/MetadataClientTest.java
```diff
@@ -0,0 +1,83 @@
...
+        consumerDefaults().withGroupId(group1);
+    final Set<TopicPartition> partitions = Collections.singleton(partition0);
+    final Timeout timeout = new Timeout(1, TimeUnit.SECONDS);
+    final ActorRef consumerActor = system().actorOf(KafkaConsumerActor.props(consumerSettings));
+
+    final CompletionStage<Map<TopicPartition, Long>> response =
+        MetadataClient.getBeginningOffsets(consumerActor, partitions, timeout, ec);
+    final Map<TopicPartition, Long> beginningOffsets = response.toCompletableFuture().join();
+
+    assertThat(beginningOffsets.get(partition0), is(0L));
```
ennru: Please add stopping the actor by sending a `Stop` message to it.

tests/src/test/java/docs/javadsl/MetadataClientTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Set;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class MetadataClientTest extends EmbeddedKafkaJunit4Test {
```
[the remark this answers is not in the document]
jewertow: I tried to use `TestcontainersKafkaLike` and it didn't compile. I used `TestcontainersKafkaJunit4Test` instead and it compiles successfully.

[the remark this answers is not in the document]
ennru: Ah sorry, Java requires that of course. Scala 2.13 compiles now, but some class cast fails during the tests.

tests/src/test/java/docs/javadsl/MetadataClientTest.java
jewertow: Done

tests/src/test/scala/akka/kafka/scaladsl/MetadataClientSpec.scala
jewertow: Done

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
jewertow: Do we really need this? I added failing test cases and the code behaves the same for:
```
.map(_.response.get)
```
as well as for:
```
.map(_.response)
      .flatMap {
        case Success(res) => Future.successful(res)
        case Failure(exception) => Future.failed(exception)
      }
```

ennru: To run it safely on the `sameThreadExecutionContext` we may not allow any exceptions happening in the operator.

ennru:
Have you thought about making the `MetadataClient` an instance which would keep the actor ref and timeout? That would open up to have a convenience factory method which would take `ConsumerSettings` and create the actor internally (a bit like your first proposal).

jewertow:
Yes, I think it's a good way.

jewertow:
I pushed new proposal. Please look again and tell me if I can continue this way for the other requests.

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
```diff
@@ -0,0 +1,61 @@
...
+import akka.kafka.{ConsumerSettings, KafkaConsumerActor}
+import akka.kafka.Metadata.{BeginningOffsets, GetBeginningOffsets}
+import akka.pattern.ask
+import akka.util.Timeout
+import org.apache.kafka.common.TopicPartition
+
+import scala.concurrent.{ExecutionContextExecutor, Future}
+import scala.util.{Failure, Success}
+
+class MetadataClient(actorSystem: ActorSystem, timeout: Timeout) {
```
ennru: It would be preferable to have one `MetadataClient` instance per consumer actor/settings.
Make the primary constructor `private` and let it take the consumer actor's `ActorRef`.
Create factory methods in the companion object, one taking the `ActorRef` and one taking `ConsumerSettings`.

jewertow: Thanks for this hint. I pushed new implementation.

jewertow: Thanks for the hint. I pushed new implementation.

jewertow: @ennru could you look at it again and say if you meant it?

ennru: Yes, that looks good.
The actor should only be stopped by the metadata client if it was created by it, you'd possibly need to pass that information to the client. So two constructors on that class might make sense.
`close` might be better than `stop`.

ennru:
It would be great if you could get back to this to push it over the line.

jewertow:
@ennru I pushed updated docs. Look at this please.

core/src/main/scala/akka/kafka/javadsl/MetadataClient.scala
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 - 2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 - 2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
```
ennru: Duplicated.

tests/src/test/java/docs/javadsl/MetadataClientTest.java
```diff
@@ -0,0 +1,240 @@
+/*
+ * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+ * Copyright (C) 2016 - 2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+/// *
+// * Copyright (C) 2014 - 2016 Softwaremill <http://softwaremill.com>
+// * Copyright (C) 2016 - 2019 Lightbend Inc. <http://www.lightbend.com>
+// */
+//
```
ennru: Remove this, please.

docs/src/main/paradox/metadata-client.md
```diff
@@ -0,0 +1,42 @@
+---
+project.description: Access Kafka consumer metadata by calling MetadataClient.
+---
+# Metadata Client
+
+`MetadataClient` is a thin wrapper for `KafkaConsumerActor` hiding the ask calls and mapping to the correct response types.
+
+To access the Kafka consumer metadata you need to create the `KafkaConsumerActor` as described in the @ref[Consumer documentation](consumer.md#sharing-the-kafkaconsumer-instance) pass it to `MetadataClient`'s factory method `create`.
+
+Another approach to create metadata client is passing the `ConsumerSettings` and `ActorSystem` objects to factory method. Then the metadata client manages the internal actor and stops it when the `close` method is called.
```
ennru:
```suggestion
Another approach to create metadata client is passing the `ConsumerSettings` and `ActorSystem` objects to the factory method. Then the metadata client manages the internal actor and stops it when the `stop` method is called.  
```
close or stop?

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
```diff
@@ -0,0 +1,106 @@
...
+  def getCommittedOffset(partition: TopicPartition): Future[OffsetAndMetadata] =
+    (consumerActor ? GetCommittedOffset(partition))(timeout)
+      .mapTo[CommittedOffset]
+      .map(_.response)
+      .flatMap {
+        case Success(res) => Future.successful(res)
+        case Failure(e) => Future.failed(e)
+      }(ExecutionContexts.sameThreadExecutionContext)
+
+  def stop(): Unit =
```
ennru: I'd still prefer to call it `close` instead. The user of this API should not think about as an actor being stopped, it is more like closing a file.

docs/src/main/paradox/metadata-client.md
```diff
@@ -0,0 +1,42 @@
...
+
+The supported metadata are
+
+| Metadata | Response type |
+|-------| ------- |
+| Topics list | `Future[Map[String, List[PartitionInfo]]]` | 
+| Partitions | `Future[List[PartitionInfo]]` |
+| Beginning offsets | `Future[Map[TopicPartition, Long]]` |
+| End offsets | `Future[Map[TopicPartition, Long]]` |
+| Committed offset | `Future[OffsetAndMetadata]` |
```
ennru: You may use the `@scala[..]@java[...]` directives to show the type in the reader's preferred language.

core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
jewertow: I fully agree with you. I forgot to rename this method before.

docs/src/main/paradox/metadata-client.md
jewertow: It should definitely be called `close`.

docs/src/main/paradox/metadata-client.md
jewertow: Done

jewertow:
I agree with you that it should be in the `consumer-metadata.md` documentation, but current implementation of `MetadataClient` does not support requests of the type `GetOffsetsForTimes`. 
I'm going to add support for the `GetOffsetsForTimes` in the foreseeable future.
For now I would add to `consumer-metadata` docs the following statement:
```
@@@ warning

Since the `KafkaConsumerActor` is the internal mechanism, the recommended method to get metadata is using the [metadata-client](metadata-client.md).
```

ennru:
Please join the new docs into the `consumer-metadata.md` page.
Show the `MetadataClient` first and have the raw experience after it.

jewertow:
Done.

ennru:
Thank you for implementing this!

jewertow:
Thank you for the opportunity to contribute to alpakka-kafka. It was the best experience I’ve ever had in open source! Thank you for your help and advices.

PR #902, opened 2019-09-16
Benchmarks: producer writing 500 bytes messages

ennru:
Add another producer benchmark writing 500b messages to understand the Akka Streams overhead a bit better.


ennru:
The Integration tests fail far more often than we'd like to. We haven't investigated it lately. Travis is a bit hard to debug.

PR #903, opened 2019-09-18
Clarify exactly-once-semantics limitations

raboof:
Though this wording is still rather confusing - when would you not be using
a set of consumed partitions and a set of partitions that are produced to?


seglo:
@raboof Yes, I think maybe this statement should be expanded for clarity.  The point it's trying to make is that you can only get transactional guarantees when doing stateless transformations in a consume, transform, produce workflow, on the same Kafka cluster.  Workflows that wouldn't work with transactions include: 

* Simply producing to partitions from a 3rd party source
* Consuming partitions from a Kafka cluster A and producing to partitions on Kafka cluster B

PR #906, opened 2019-09-25
Benchmarks: try producing with ridiculous parallelism

ennru:
## Purpose

Experiment with producer benchmarks and how they get affected by the backpressure.


ennru:
Force merged this to get fast feedback.

PR #913, opened 2019-09-29
CI: update Adopt Open JDK versions

ennru:
The former JDK 8 version became invalid for use with Jabba.


ennru:
Merged to make builds work again.

PR #914, opened 2019-09-29
Akka 2.6.0-M8 for nightly builds

ennru:


ennru:
Force merged as it only affects nightly/cron builds.

PR #917, opened 2019-09-30
Benchmarks: create new topics for producers tests

ennru:
## Purpose

Try if topic name reuse causes the benchmarks to show the unexpected drop in Raw Kafka performance since build 550.

image

## References

This drop must be connected to https://github.com/akka/alpakka-kafka/pull/902

## Changes

Use a fresh topic name for producer benchmarks so they do not interfere with consumer benchmarks.

### DIFF
--- a/benchmarks/src/it/scala/akka/kafka/benchmarks/Producer.scala
+++ b/benchmarks/src/it/scala/akka/kafka/benchmarks/Producer.scala
@@ -8,23 +8,23 @@ class ApacheKafkaPlainProducer extends BenchmarksBase() {
   private val prefix = "apache-kafka-plain-producer"
 
   it should "bench with small messages" in {
-    val cmd = RunTestCommand(prefix, bootstrapServers, topic_2000_100)
+    val cmd = RunTestCommand(prefix, bootstrapServers, topic_2000_100.freshTopic)
     runPerfTest(cmd, KafkaProducerFixtures.initializedProducer(cmd), KafkaProducerBenchmarks.plainFlow)
   }
 
   it should "bench with 500b messages" in {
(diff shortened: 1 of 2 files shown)


benchmarks/src/it/scala/akka/kafka/benchmarks/Producer.scala
```diff
@@ -8,23 +8,23 @@ class ApacheKafkaPlainProducer extends BenchmarksBase() {
...
   it should "bench with normal messages" in {
-    val cmd = RunTestCommand(prefix + "-normal-msg", bootstrapServers, topic_2000_5000)
+    val cmd = RunTestCommand(prefix + "-normal-msg", bootstrapServers, topic_2000_5000.freshTopic)
     runPerfTest(cmd, KafkaProducerFixtures.initializedProducer(cmd), KafkaProducerBenchmarks.plainFlow)
   }
 
   it should "bench with normal messages written to 8 partitions" in {
     val cmd =
-      RunTestCommand(prefix + "-normal-msg-8-partitions", bootstrapServers, topic_2000_5000_8)
+      RunTestCommand(prefix + "-normal-msg-8-partitions", bootstrapServers, topic_2000_5000_8.freshTopic)
```
seglo: I don't see what changed here between #902 and here.  If it was reusing the same topic, it was doing that before as well.  The new benchmark for 500b messages would have had its own random topic id.

ennru:
I agree this is weird. It didn't have the effect I hoped for, but suddenly the Plain Consumer hopped back to where it was earlier?! https://jenkins.akka.io:8498/job/alpakka-kafka-perf-master/plot/#1446559666.csv

PR #922, opened 2019-10-03
Benchmarks: don't reuse topics for raw batched consumers

ennru:
As https://github.com/akka/alpakka-kafka/pull/919 removed the problem for the 5000k version.


seglo:
CI failure looks unrelated.  Logged an issue #923.

PR #926, opened 2019-10-08
sbt 1.3.5

ennru:
Use sbt 1.3.2 in the build and adapt caching directories in Travis.

### DIFF
--- a/.jvmopts-travis
+++ b/.jvmopts-travis
@@ -1,5 +1,7 @@
 # This is used to configure the sbt instance that Travis launches
 
+-Dfile.encoding=UTF8
+-Dsbt.color=always
 -Xms1G
 -Xmx1G
 -Xss2M
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,10 +107,9 @@ before_cache:
 
 cache:
   directories:
-    - $HOME/.coursier
+    - $HOME/.cache/coursier
     - $HOME/.ivy2/cache
-    - $HOME/.sbt/boot
-    - $HOME/.sbt/launchers
+    - $HOME/.sbt
     - $HOME/.jabba
 
 env:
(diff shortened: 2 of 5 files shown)


ennru:
After restarting a few build steps for strange lock errors everything went through, but the docker-compose testing seems broken. It could be the exclusions I added.
```
[info] akka.kafka.PartitionedSourceFailoverSpec *** ABORTED *** (30 milliseconds)
[info]   java.lang.NoClassDefFoundError: org/glassfish/hk2/api/MultiException
[info]   at akka.kafka.PartitionedSourceFailoverSpec.<init>(PartitionedSourceFailoverSpec.scala:30)
```

ennru:
The new version of MiMa discovers the change in generics on `flowWithOffsetContext` from #931

ennru:
The integration tests still fail for a missing class. I need to revisit the dependency exclusions.

ennru:
The test run further this time but failed for another missing class
```
[info]   Cause: java.lang.IllegalStateException: InjectionManagerFactory not found.
[info]   at org.glassfish.jersey.internal.inject.Injections.lambda$lookupInjectionManagerFactory$0(Injections.java:98)
```
https://travis-ci.org/akka/alpakka-kafka/jobs/599684381#L3634

For some reason, the transitive dependencies of the "docker-client" library are not added automatically by coursier.

With the work in #939 we might be able to do this testing without the "docker-client", so let's wait for that.

ennru:
With sbt 1.3.4 the Dependency Graph works again. This is good to go, I think.

ennru:
Failure is https://github.com/akka/alpakka-kafka/issues/611

ennru:
The terminal colours are lost here, as well.

ennru:
This looks like a regression in sbt https://travis-ci.org/akka/alpakka-kafka/jobs/625607995#L1065

ennru:
An even here shows #957 https://travis-ci.org/akka/alpakka-kafka/jobs/625607991#L978

PR #930, opened 2019-10-11
Add partitioned transactional source phase 1

seglo:
## Purpose

This PR adds a new partitioned source with support for Kafka transactions, `Transactions.partitionedSource`.  A `Source` is created per partition in a consumer group subscription.  Each `Source` must have a downstream `Transactional.sink|flow` to commit messages transactionally.  A derived `transactional.id` is used per `Transactional.sink|flow` that's based on a user-provided `transactional.id` along with the `group.id`, topic and partition of the originally consumed message.  `Transactions.partitionedSource` will make it easier to run transactional workloads across consumer groups with multiple members.

This PR is based on the extensive work already completed by @charlibot earlier this year. It was rebased by @raboof and updated to accommodate transactional consistency work completed by @2m and @szymonm.  This new branch will serve as the integration point to get this functionality into Alpakka Kafka 2.0.0.

**Update 27/11/2019**

Due to the large number of changes introduced by this PR it has been repurposed as a "phase 1" to lay the groundwork to continue this work.  This PR will not expose the `Transactional.partitionedSource` to the end user.

## References

* Original PR: #705 
* Transactions draining logic PR: #757 
* Original Transactions PR: #420

## Phase 1 tasks

- [x] Core impl.
- [x] Confirm producer per substream is acceptable (https://github.com/akka/alpakka-kafka/pull/705#discussion_r330509381)
- [x] Review docs.  Include details about KIP-447.
- [x] Multi-broker integration test (blocked by https://github.com/akka/alpakka-kafka/pull/939)
- [x] Test: If sub source fails, rebalance occurs and new source is created (https://github.com/akka/alpakka-kafka/pull/705#discussion_r251468274)
- [x] Test: Consistency after rebalance of regular transactions
- [x] Update docs. (_to be continued in phase 2_)
- [x] Test retry infrastructure

## Phase 2 tasks (preliminary)

- Test: Consistency of partitioned source when calling shutdown using draining control
- Set `ConsumerSettings.stopTimeout` to 0 to allow for faster shutdowns when a stream completes/fails.
- Update docs. (_to be continued in phase 2_)
- Add more low level testing of draining control logic

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/CommittableSources.scala
+++ b/core/src/main/scala/akka/kafka/internal/CommittableSources.scala
@@ -211,7 +211,7 @@ private[kafka] class KafkaAsyncConsumerCommitterRef(private val consumerActor: A
 }
 
 @InternalApi
-private class CommittableSubSourceStageLogic[K, V](
+private final class CommittableSubSourceStageLogic[K, V](
     shape: SourceShape[CommittableMessage[K, V]],
     tp: TopicPartition,
     consumerActor: ActorRef,
--- a/core/src/main/scala/akka/kafka/internal/CommittableSources.scala
+++ b/core/src/main/scala/akka/kafka/internal/CommittableSources.scala
@@ -10,11 +10,12 @@ import akka.annotation.InternalApi
 import akka.kafka.ConsumerMessage.{CommittableMessage, CommittableOffset, CommittableOffsetBatch}
 import akka.kafka._
 import akka.kafka.internal.KafkaConsumerActor.Internal.{Commit, CommitSingle, CommitWithoutReply}
+import akka.kafka.internal.SubSourceLogic._
 import akka.kafka.scaladsl.Consumer.Control
 import akka.pattern.AskTimeoutException
 import akka.stream.SourceShape
 import akka.stream.scaladsl.Source
-import akka.stream.stage.GraphStageLogic
+import akka.stream.stage.{AsyncCallback, GraphStageLogic}
 import akka.util.Timeout
 import akka.{Done, NotUsed}
 import org.apache.kafka.clients.consumer.{ConsumerConfig, ConsumerRecord, OffsetAndMetadata}
@@ -102,17 +103,34 @@ private[kafka] final class CommittableSubSource[K, V](
     ) {
   override protected def logic(
       shape: SourceShape[(TopicPartition, Source[CommittableMessage[K, V], NotUsed])]
-  ): GraphStageLogic with Control =
-    new SubSourceLogic[K, V, CommittableMessage[K, V]](shape, settings, subscription, getOffsetsOnAssign, onRevoke)
-      with CommittableMessageBuilder[K, V]
-      with MetricsControl {
-      override def metadataFromRecord(record: ConsumerRecord[K, V]): String = _metadataFromRecord(record)
-      override def groupId: String = settings.properties(ConsumerConfig.GROUP_ID_CONFIG)
-      lazy val committer: KafkaAsyncConsumerCommitterRef = {
-        val ec = materializer.executionContext
-        new KafkaAsyncConsumerCommitterRef(consumerActor, settings.commitTimeout)(ec)
-      }
+  ): GraphStageLogic with Control = {
+
+    val factory = new SubSourceStageLogicFactory[K, V, CommittableMessage[K, V]] {
+      def create(
+          shape: SourceShape[CommittableMessage[K, V]],
+          tp: TopicPartition,
+          consumerActor: ActorRef,
+          subSourceStartedCb: AsyncCallback[(TopicPartition, ControlAndStageActor)],
+          subSourceCancelledCb: AsyncCallback[(TopicPartition, SubSourceCancellationStrategy)],
+          actorNumber: Int
+      ): SubSourceStageLogic[K, V, CommittableMessage[K, V]] =
+        new CommittableSubSourceStageLogic(shape,
+                                           tp,
+                                           consumerActor,
+                                           subSourceStartedCb,
+                                           subSourceCancelledCb,
+                                           actorNumber,
+                                           settings,
+                                           _metadataFromRecord)
+
     }
+    new SubSourceLogic[K, V, CommittableMessage[K, V]](shape,
+                                                       settings,
+                                                       subscription,
+                                                       getOffsetsOnAssign,
+                                                       onRevoke,
+                                                       subSourceStageLogicFactory = factory)
+  }
 }
 
 /**
@@ -190,4 +208,30 @@ private[kafka] class KafkaAsyncConsumerCommitterRef(private val consumerActor: A
         this.consumerActor == that.consumerActor && this.commitTimeout == that.commitTimeout
       case _ => false
     }
+}
+
+@InternalApi
+private class CommittableSubSourceStageLogic[K, V](
+    shape: SourceShape[CommittableMessage[K, V]],
+    tp: TopicPartition,
+    consumerActor: ActorRef,
+    subSourceStartedCb: AsyncCallback[(TopicPartition, ControlAndStageActor)],
+    subSourceCancelledCb: AsyncCallback[(TopicPartition, SubSourceCancellationStrategy)],
+    actorNumber: Int,
+    consumerSettings: ConsumerSettings[K, V],
+    _metadataFromRecord: ConsumerRecord[K, V] => String = CommittableMessageBuilder.NoMetadataFromRecord
+) extends SubSourceStageLogic[K, V, CommittableMessage[K, V]](shape,
+                                                                tp,
+                                                                consumerActor,
+                                                                subSourceStartedCb,
+                                                                subSourceCancelledCb,
+                                                                actorNumber)
+    with CommittableMessageBuilder[K, V] {
+
+  override def metadataFromRecord(record: ConsumerRecord[K, V]): String = _metadataFromRecord(record)
+  override def groupId: String = consumerSettings.properties(ConsumerConfig.GROUP_ID_CONFIG)
+  lazy val committer: KafkaAsyncConsumerCommitterRef = {
+    val ec = materializer.executionContext
+    new KafkaAsyncConsumerCommitterRef(consumerActor, consumerSettings.commitTimeout)(ec)
+  }
 }
--- a/core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
@@ -51,6 +51,7 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
   private lazy val decider: Decider =
     inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
   protected val awaitingConfirmation = new AtomicInteger(0)
+
   private var inIsClosed = false
   private var completionState: Option[Try[Done]] = None
 
@@ -60,7 +61,7 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
 
   override def preStart(): Unit = {
     super.preStart()
-    resolveProducer()
+    resolveProducer(stage.settings)
   }
 
   def checkForCompletion(): Unit =
@@ -85,11 +86,14 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     failStage(ex)
   }
 
-  def postSend(msg: Envelope[K, V, P]) = ()
+  def filterSend(msg: Envelope[K, V, P]): Boolean = true
+
+  def postSend(msg: Envelope[K, V, P]): Unit = ()
 
   override protected def producerAssigned(): Unit = resumeDemand()
 
   protected def resumeDemand(tryToPull: Boolean = true): Unit = {
+    log.debug("Resume demand")
     setHandler(stage.out, new OutHandler {
       override def onPull(): Unit = tryPull(stage.in)
     })
@@ -99,21 +103,28 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     }
   }
 
-  protected def suspendDemand(): Unit =
+  protected def suspendDemand(fromStageLogicConstructor: Boolean = false): Unit = {
+    // not permitted to access stage logic members from constructor
+    if (!fromStageLogicConstructor) log.debug("Suspend demand")
     setHandler(
       stage.out,
       new OutHandler {
         override def onPull(): Unit = ()
       }
     )
+  }
 
   // suspend demand until a Producer has been created
-  suspendDemand()
+  suspendDemand(fromStageLogicConstructor = true)
 
   setHandler(
     stage.in,
     new InHandler {
-      override def onPush(): Unit = produce(grab(stage.in))
+      override def onPush(): Unit = {
+        val msg = grab(stage.in)
+        if (filterSend(msg))
+          produce(msg)
+      }
 
       override def onUpstreamFinish(): Unit = {
         inIsClosed = true
@@ -183,5 +194,4 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     closeProducer()
     super.postStop()
   }
-
 }
--- a/core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
@@ -59,6 +59,22 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
 
   final override val producerSettings: ProducerSettings[K, V] = stage.settings
 
+  protected class DefaultInHandler extends InHandler {
+    override def onPush(): Unit = produce(grab(stage.in))
+
+    override def onUpstreamFinish(): Unit = {
+      inIsClosed = true
+      completionState = Some(Success(Done))
+      checkForCompletion()
+    }
+
+    override def onUpstreamFailure(ex: Throwable): Unit = {
+      inIsClosed = true
+      completionState = Some(Failure(ex))
+      checkForCompletion()
+    }
+  }
+
   override def preStart(): Unit = {
     super.preStart()
     resolveProducer(stage.settings)
@@ -86,9 +102,7 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     failStage(ex)
   }
 
-  def filterSend(msg: Envelope[K, V, P]): Boolean = true
-
-  def postSend(msg: Envelope[K, V, P]): Unit = ()
+  protected def postSend(msg: Envelope[K, V, P]): Unit = ()
 
   override protected def producerAssigned(): Unit = resumeDemand()
 
@@ -103,9 +117,13 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     }
   }
 
-  protected def suspendDemand(fromStageLogicConstructor: Boolean = false): Unit = {
-    // not permitted to access stage logic members from constructor
-    if (!fromStageLogicConstructor) log.debug("Suspend demand")
+  protected def suspendDemand(): Unit = {
+    log.debug("Suspend demand")
+    suspendDemandOutHandler()
+  }
+
+  // factored out of suspendDemand because logging is not permitted when called from the stage logic constructor
+  private def suspendDemandOutHandler(): Unit = {
     setHandler(
       stage.out,
       new OutHandler {
@@ -114,33 +132,14 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     )
   }
 
+  protected def initialInHandler(): Unit = producingInHandler()
+  protected def producingInHandler(): Unit = setHandler(stage.in, new DefaultInHandler())
+
   // suspend demand until a Producer has been created
-  suspendDemand(fromStageLogicConstructor = true)
+  suspendDemandOutHandler()
+  initialInHandler()
 
-  setHandler(
-    stage.in,
-    new InHandler {
-      override def onPush(): Unit = {
-        val msg = grab(stage.in)
-        if (filterSend(msg))
-          produce(msg)
-      }
-
-      override def onUpstreamFinish(): Unit = {
-        inIsClosed = true
-        completionState = Some(Success(Done))
-        checkForCompletion()
-      }
-
-      override def onUpstreamFailure(ex: Throwable): Unit = {
-        inIsClosed = true
-        completionState = Some(Failure(ex))
-        checkForCompletion()
-      }
-    }
-  )
-
-  def produce(in: Envelope[K, V, P]): Unit =
+  protected def produce(in: Envelope[K, V, P]): Unit =
     in match {
       case msg: Message[K, V, P] =>
         val r = Promise[Result[K, V, P]]
--- a/core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
+++ b/core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
@@ -45,6 +45,8 @@ private[kafka] trait TransactionalMessageBuilderBase[K, V, Msg] extends MessageB
   def committedMarker: CommittedMarker
 
   def onMessage(consumerMessage: ConsumerRecord[K, V]): Unit
+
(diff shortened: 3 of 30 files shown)


seglo:
I received confirmation that Kafka Streams EoS creates one producer per source partition.  This will introduce performance implications which will be addressed in [KIP-447, Producer scalability for exactly once semantics](https://cwiki.apache.org/confluence/display/KAFKA/KIP-447%3A+Producer+scalability+for+exactly+once+semantics).  This work is in progress and will land in a future version of Kafka (not 2.4.0).

* [Talk: Revisiting Exactly One Semantics (EOS) by Jason Gustafson](https://www.youtube.com/watch?v=j0l_zUhQaTc&list=PLa7VYi0yPIH3kDYOrar5B26WkslCGfDy_&index=7)
* kafka-users thread, [Questions about Producer per Task/Partition in Streams EoS Impl.](https://lists.apache.org/thread.html/491b89840776498009fa773a737bf512a2c144558cce961b4ff8c768@%3Cusers.kafka.apache.org%3E)

In the meantime we can proceed with our current implementation to have one producer per partitioned source.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -31,19 +32,24 @@ import scala.util.{Failure, Success}
...
     getOffsetsOnAssign: Option[Set[TopicPartition] => Future[Map[TopicPartition, Long]]] = None,
-    onRevoke: Set[TopicPartition] => Unit = _ => ()
+    onRevoke: Set[TopicPartition] => Unit = _ => (),
+    msgSourceLogicFactory: (SourceShape[Msg],
+                            TopicPartition,
+                            ActorRef,
+                            AsyncCallback[(TopicPartition, (Control, ActorRef))],
+                            AsyncCallback[(TopicPartition, Option[ConsumerRecord[K, V]])],
+                            Int) => MessageSubSourceLogic[K, V, Msg]
 ) extends TimerGraphStageLogic(shape)
```
ennru: This is worth a named type `SourceLogicFactory`.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -29,28 +29,31 @@ import scala.util.{Failure, Success, Try}
...
 
 /**
  * Internal API.
  *
  * Used by [[DefaultProducerStage]], extended by [[TransactionalProducerStageLogic]].
  */
-private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <: Results[K, V, P]](
-    stage: ProducerStage[K, V, P, IN, OUT],
-    producer: Producer[K, V],
+private abstract class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <: Results[K, V, P], S](
```
ennru: The new type parameter is not obvious, add a `@tparam` to the Scaladoc.

core/src/main/scala/akka/kafka/internal/CommittableSources.scala
```diff
@@ -102,22 +102,39 @@ private[kafka] final class CommittableSubSource[K, V](
...
-        val ec = materializer.executionContext
-        new KafkaAsyncConsumerCommitterRef(consumerActor, settings.commitTimeout)(ec)
+  ): GraphStageLogic with Control = {
+    def factory(shape: SourceShape[CommittableMessage[K, V]],
+                tp: TopicPartition,
+                consumerActor: ActorRef,
+                subSourceStartedCb: AsyncCallback[(TopicPartition, (Control, ActorRef))],
+                subSourceCancelledCb: AsyncCallback[(TopicPartition, Option[ConsumerRecord[K, V]])],
+                actorNumber: Int): MessageSubSourceLogic[K, V, CommittableMessage[K, V]] =
+      new MessageSubSourceLogic[K, V, CommittableMessage[K, V]](shape,
```
ennru: This could get a named type.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -29,28 +29,31 @@ import scala.util.{Failure, Success, Try}
...
+    stage: ProducerStage[K, V, P, IN, OUT, S],
+    producerProvider: S => Producer[K, V],
     inheritedAttributes: Attributes
 ) extends TimerGraphStageLogic(stage.shape)
     with StageLogging
     with MessageCallback[K, V, P]
     with ProducerCompletionState {
 
+  protected var producer: Producer[K, V]
```
ennru: Can this be encapsulated better? So that it becomes more obvious that the producer is created once and never used before it is created.

core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -45,6 +45,8 @@ private[kafka] trait TransactionalMessageBuilderBase[K, V, Msg] extends MessageB
   def committedMarker: CommittedMarker
 
   def onMessage(consumerMessage: ConsumerRecord[K, V]): Unit
+
+  val fromPartitionedSource: Boolean
```
ennru: Better use `def` here and leave the overrides as `val`.

core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -53,15 +55,13 @@ private[kafka] trait TransactionalMessageBuilder[K, V]
...
   override def createMessage(rec: ConsumerRecord[K, V]): TransactionalMessage[K, V] = {
     onMessage(rec)
-    val offset = PartitionOffsetCommittedMarker(
-      GroupTopicPartition(
-        groupId = groupId,
-        topic = rec.topic,
-        partition = rec.partition
-      ),
-      offset = rec.offset,
-      committedMarker
```
ennru: I find the old notation easier to grasp.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -59,7 +65,7 @@ private abstract class SubSourceLogic[K, V, Msg](
 
   /** We have created a source for these partitions, but it has not started up and is not in subSources yet. */
   var partitionsInStartup: immutable.Set[TopicPartition] = immutable.Set.empty
-  var subSources: Map[TopicPartition, Control] = immutable.Map.empty
+  var subSources: Map[TopicPartition, (Control, ActorRef)] = immutable.Map.empty
```
ennru: It is not obvious which actor this is (source stage actor or consumer actor). I had to dig back and forth a while. Maybe a case class instead?

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -222,14 +229,14 @@ private abstract class SubSourceLogic[K, V, Msg](
...
       partitionsInStartup += tp
-      val subSource = Source.fromGraph(
-        new SubSourceStage(tp,
-                           consumerActor,
-                           subsourceStartedCB,
-                           subsourceCancelledCB,
-                           messageBuilder = this,
-                           actorNumber)
-      )
+      val graphStage = new GraphStage[SourceShape[Msg]] {
```
ennru: Create a names class for this.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -278,90 +291,92 @@ private object SubSourceLogic {
 }
 
 /** Internal API */
-private final class SubSourceStage[K, V, Msg](
+@InternalApi
+private abstract class MessageSubSourceLogic[K, V, Msg](
```
ennru: Is this a `SubSourcLogic` with `MessageBuilder`? Is there a better name?

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -99,18 +100,22 @@ private object TransactionalProducerStage {
  * Transaction (Exactly-Once) Producer State Logic
  */
 private final class TransactionalProducerStageLogic[K, V, P](stage: TransactionalProducerStage[K, V, P],
-                                                             producer: Producer[K, V],
+                                                             producerProvider: String => Producer[K, V],
```
ennru: We'll get conflicts with the Akka Discovery PR here...

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -99,18 +100,22 @@ private object TransactionalProducerStage {
...
+                                                                                            producerProvider,
+                                                                                            inheritedAttributes)
     with StageLogging
     with MessageCallback[K, V, P]
     with ProducerCompletionState {
 
   import TransactionalProducerStage._
 
+  override protected var producer: Producer[K, V] = _
+  private var initiated: Boolean = _
```
ennru: put the `false` explicitly right here.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -222,6 +246,6 @@ private final class TransactionalProducerStageLogic[K, V, P](stage: Transactiona
 
   private def abortTransaction(): Unit = {
     log.debug("Aborting transaction")
-    producer.abortTransaction()
+    Option(producer).foreach(_.abortTransaction())
```
ennru: A null-check is nicer.

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -48,6 +54,21 @@ object Transactional {
...
+  /**
+   * The `partitionedSource` is a way to track automatic partition assignment from kafka.
+   * Each source is setup for for Exactly Only Once (EoS) kafka message semantics.
+   * To enable EoS it's necessary to use the [[Transactional.sink]] or [[Transactional.flow]] (for passthrough).
+   * When Kafka rebalances partitions, all sources complete before the remaining sources are issued again.
+   *
+   * By generating the `transactionalId` from the [[TopicPartition]], multiple instances of your application can run
+   * without having to manually assign partitions to each instance.
+   */
+  def partitionedSource[K, V](
```
ennru: Mark it `@ApiMayChange`.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
seglo: I renamed `MessageSubSourceLogic` to `SubSourceStageLogic` and created a type called `SubSourceStageLogicFactory`.  I also implemented concrete types for Transactional, Committable, and Plain sub sources that subclass `SubSourceStageLogic`.

core/src/main/scala/akka/kafka/internal/CommittableSources.scala
seglo: Concrete type created `CommittableSubSourceStageLogic`

https://github.com/akka/alpakka-kafka/pull/930#discussion_r340823268

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
seglo: This type param is no longer applicable.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
seglo: I didn't encapsulate the variable itself, but I did add the `ProducerAssignmentLifecycle` enum/workflow to make it more clear what state the producer in the logs and when working with messages in the producer stage.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
seglo: I cleaned this up.

https://github.com/akka/alpakka-kafka/pull/930#discussion_r340823268

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
seglo: It's the control of the `SubSourceStageLogic` and the stage actor for that stage.  I created a case class `SubSourceStageLogic.ControlAndStageActor`.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
seglo: I cleaned this up.

https://github.com/akka/alpakka-kafka/pull/930#discussion_r340823268

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
seglo: We sure did!  It's been handled now.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
seglo: No longer applicable.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -36,6 +36,13 @@ private[kafka] class DefaultProducerStage[K, V, P, IN <: Envelope[K, V, P], OUT
     new DefaultProducerStageLogic(this, inheritedAttributes)
 }
 
+private[kafka] object DefaultProducerStageLogic {
+  sealed trait ProducerAssignmentLifecycle
+  case object Unassigned extends ProducerAssignmentLifecycle
+  case object AsyncCreateRequestSent extends ProducerAssignmentLifecycle
+  case object Assigned extends ProducerAssignmentLifecycle
```
ennru: Some API docs on these? Or somewhere else where the concept can be explained.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -273,98 +290,160 @@ private abstract class SubSourceLogic[K, V, Msg](
...
+ *
+ * SubSourceStageLogic [[akka.kafka.scaladsl.Consumer.Control]] and the stage actor [[ActorRef]]
+ */
+@InternalApi
+private final case class ControlAndStageActor(control: Control, stageActor: ActorRef)
+
+sealed trait SubSourceCancellationStrategy
+final case class SeekToOffsetAndReEmit(offset: Long) extends SubSourceCancellationStrategy
+case object ReEmit extends SubSourceCancellationStrategy
+case object DoNothing extends SubSourceCancellationStrategy
```
ennru: Explain this concept.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
seglo: I'll add some docs.  Basically it's used within the `TransactionalProducerStage` to handle messages differently depending on the state of the async producer assignment.  It's also useful to log state transitions for troubleshooting, so I included that too.  I incorporated it into your new trait `DeferredProducer`.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
seglo: This decides if the sub source is re-emmitted downstream (after possibly re-seeking to last committed offsets).  For the transactional partitioned source we want to stop, fail, or shutdown the outer `SubSourceStage` if any sub source cancels.  I didn't change the default behaviour which is now implemented as a combination of `SeekToOffsetAndReEmit` and `ReEmit`.  I'll add comments.

tests/src/it/scala/akka/kafka/TransactionsPartitionedSourceSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+          })(Keep.left)
+          .run()
+
+      val controls: Seq[UniqueKillSwitch] = (0 until consumers)
+        .map(_.toString)
+        .map(runStream)
+
+      while (completedCopy.get() < consumers) {
+        Thread.sleep(2000)
+      }
```
ennru:
```
      eventually(Interval(2.seconds)) {
        completedCopy.get() shouldBe consumers
      }
```

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -328,6 +329,91 @@ class IntegrationSpec extends SpecBase with TestcontainersKafkaLike with Inside
       }
     }
 
+    "rebalance safely using transactional partitioned flow" in assertAllStagesStopped {
```
ennru: Would it make sense to have this in `TransactionsSpec` or a new one? `IntegrationSpec` is very unspecific.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -99,21 +103,28 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     }
   }
 
-  protected def suspendDemand(): Unit =
+  protected def suspendDemand(fromStageLogicConstructor: Boolean = false): Unit = {
+    // not permitted to access stage logic members from constructor
+    if (!fromStageLogicConstructor) log.debug("Suspend demand")
```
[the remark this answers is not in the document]
seglo: I agree it's weird.  Maybe we can suspend demand directly in the constructor instead of calling this method.  I think the logging under normal circumstances is useful.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -85,11 +86,14 @@ private class DefaultProducerStageLogic[K, V, P, IN <: Envelope[K, V, P], OUT <:
     failStage(ex)
   }
 
-  def postSend(msg: Envelope[K, V, P]) = ()
+  def filterSend(msg: Envelope[K, V, P]): Boolean = true
```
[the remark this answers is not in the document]
seglo: Good idea.  I've changed this logic set a new handler which is only overridden in `TransactionalProducerStage`.

core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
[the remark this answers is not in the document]
seglo: I factored out setting the `OutHandler` so it can be called from `suspendDemand` and the stage logic constructor.

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
```diff
@@ -178,6 +183,91 @@ private[internal] abstract class TransactionalSourceLogic[K, V, Msg](shape: Sour
...
+          override def onAssign(assignedTps: Set[TopicPartition], consumer: RestrictedConsumer): Unit = ()
+
+          // This is invoked in the KafkaConsumerActor thread when doing poll.
+          override def onRevoke(revokedTps: Set[TopicPartition], consumer: RestrictedConsumer): Unit =
+            if (revokedTps.isEmpty) ()
+            else if (waitForDraining(revokedTps)) {
+              subSources.values.map(_.stageActor).foreach(_ ! Revoked(revokedTps.toList))
+            } else {
+              sourceActor.ref ! Status.Failure(new Error("Timeout while draining"))
+              consumerActor ! KafkaConsumerActor.Internal.Stop
```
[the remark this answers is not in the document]
seglo: > Hmm, `TransactionalSourceLogic` above uses `consumerActor` as sender, that doesn't make sense to me.

I think this has something to do with the draining logic.  It's a complicated asynchronous dance of stage logic timers and async callbacks.  It is interesting that the transactional sub source logic isn't using the consumer actor too.  Either it was copied before that change was made, or it was intentional.  Hard to say which.

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
```diff
@@ -178,6 +183,91 @@ private[internal] abstract class TransactionalSourceLogic[K, V, Msg](shape: Sour
...
+        implicit val timeout = Timeout(txConsumerSettings.commitTimeout)
+        try {
+          val drainCommandFutures = subSources.values.map(_.stageActor).map(ask(_, Drain(partitions, None, Drained)))
+          implicit val ec = executionContext
+          Await.result(Future.sequence(drainCommandFutures), timeout.duration)
+          true
+        } catch {
+          case t: Throwable =>
+            false
+        }
```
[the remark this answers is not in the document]
seglo: Yes.  We need to understand this better through tests.

> the `ec` could be applied explicitly

`Future.sequence` takes several implicit params.  If you define one then you have to define them all.  I'm not sure where `CanBuildFrom` is coming from.

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
```diff
@@ -246,3 +335,90 @@ private object TransactionalSourceLogic {
...
+      inFlightRecords.committed(offsets.view.mapValues(_.offset() - 1).toMap)
+      sender ! Done
+    case (sender, CommittingFailure) => {
+      log.info("Committing failed, resetting in flight offsets")
+      inFlightRecords.reset()
+    }
+    case (sender, Drain(partitions, ack, msg)) =>
+      if (inFlightRecords.empty(partitions)) {
+        log.debug(s"Partitions drained ${partitions.mkString(",")}")
+        ack.getOrElse(sender) ! msg
```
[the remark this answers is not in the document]
seglo: It seems the `sender` for the ack can change.  I'm hesitant about modifying `Drain` in this PR since it's used by the current transactional logic.

tests/src/test/scala/akka/kafka/TransactionsOps.scala
```diff
@@ -0,0 +1,212 @@
...
+      maxPartitions: Int,
+      restartAfter: Option[Int] = None
+  ): Source[ProducerMessage.Results[String, String, PartitionOffset], Control] =
+    Transactional
+      .partitionedSource(consumerSettings, Subscriptions.topics(sourceTopic))
+      .flatMapMerge(
+        maxPartitions, {
+          case (_, source) =>
+            val results: Source[ProducerMessage.Results[String, String, PartitionOffset], NotUsed] = source
+              .zip(Source.unfold(1)(count => Some((count + 1, count))))
```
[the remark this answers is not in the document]
seglo: I believe that's the only difference.  This logic was copied verbatim from `TransactionsSpec` to be used across that test class and the integration tests.

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
seglo: `TransactionsSpec` works.

seglo: The test requires a little more relaxed "patience" config due to the rebalancing.

core/src/main/scala/akka/kafka/javadsl/Transactional.scala
```diff
@@ -54,6 +55,31 @@ object Transactional {
...
+   * Each source is setup for for Exactly Only Once (EoS) kafka message semantics.
+   * To enable EoS it's necessary to use the [[Transactional.sink]] or [[Transactional.flow]] (for passthrough).
+   * When Kafka rebalances partitions, all sources complete before the remaining sources are issued again.
+   *
+   * By generating the `transactionalId` from the [[TopicPartition]], multiple instances of your application can run
+   * without having to manually assign partitions to each instance.
+   */
+  @ApiMayChange
+  @InternalApi
+  private[kafka] def partitionedSource[K, V](
```
ennru: We should comment out the Java API. Java doesn't support scoped private visibility and it will seem public to Java users.

ennru:
I commented it out and added `final`, `private`, and `protected` at a few places.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -157,6 +170,47 @@ private final class TransactionalProducerStageLogic[K, V, P](
...
+          throw new IllegalStateException("Cannot reapply first message")
+        }
+        // stash the first message so it can be sent after the producer is assigned
+        firstMessage = Some(msg)
+        // initiate async async producer request _after_ first message is stashed in case future eagerly resolves
+        // instead of asynccallback
+        resolveProducer(generatedTransactionalConfig(msg))
+        // suspend demand after we receive the first message until the producer is assigned
+        suspendDemand()
+        false
```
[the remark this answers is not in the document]
ennru: I split it into two `case` and a guard.

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
ennru: If needed, you can define a test case specific `PatienceConfig`.

PR #931, opened 2019-10-15
Committer: change type bound to allow offset batches

ennru:
Change the type bound on the new `Committer.flowWithOffsetContext` and `Committer.sinkWithOffsetContext` from `CommittableOffset` to `Comittable` so that commit offsets can be accepted pre-aggreaget in `CommittableOffsetBatch` (which does not inherit `CommittableOffset`).

### DIFF
--- a/core/src/main/scala/akka/kafka/javadsl/Committer.scala
+++ b/core/src/main/scala/akka/kafka/javadsl/Committer.scala
@@ -36,7 +36,7 @@ object Committer {
    * `CommittableOffsetBatch` as context.
    */
   @ApiMayChange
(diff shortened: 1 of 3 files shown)


ennru:
That would be https://github.com/akka/alpakka-kafka/blob/98a4b966f31a1fc75315b250cb5762a0de62bad1/core/src/main/scala/akka/kafka/ConsumerMessage.scala#L152 

That could take `Committable`, but it would not be useful to send in `CommittableOffsetBatch` to create a new `CommittableOffsetBatch` out of it.

That change would not be binary compatible, either.
The changes I suggest -- which are on methods marked as "may change" -- are both binary and source compatible as they make a generic type wider than before.

PR #932, opened 2019-10-15
Create `committableSink` as combination of Producer.flexiFlow an…

ennru:
## Purpose

For the common consume-produce flow, a combination of producer and committer is useful. It conceptually consists of a `Producer.flexiFlow` with a `Committer.sink`, but could be implemented as a specific sink stage later.

## Changes

* Add `committableSink` which combines `flexiFlow` with `Committer.sink`
* Add `committableSinkWithOffsetContext` which combines `flowWithOffsetContext` with `Committer.sinkWithOffsetContext` (but is implemented simpler than that)

## Background Context

The committer uses commit batching internally while the producer `committableSink` sends commits one-by-one.

### DIFF
--- a/core/src/main/scala/akka/kafka/javadsl/Producer.scala
+++ b/core/src/main/scala/akka/kafka/javadsl/Producer.scala
@@ -8,10 +8,11 @@ package akka.kafka.javadsl
 import java.util.concurrent.CompletionStage
 
 import akka.annotation.ApiMayChange
-import akka.{Done, NotUsed}
+import akka.kafka.ConsumerMessage.{Committable, CommittableOffset}
 import akka.kafka.ProducerMessage._
-import akka.kafka.{scaladsl, ConsumerMessage, ProducerSettings}
-import akka.stream.javadsl.{Flow, FlowWithContext, Sink}
+import akka.kafka.{scaladsl, CommitterSettings, ConsumerMessage, ProducerSettings}
+import akka.stream.javadsl.{Flow, FlowWithContext, Keep, Sink}
+import akka.{japi, Done, NotUsed}
 import org.apache.kafka.clients.producer.ProducerRecord
 
 import scala.compat.java8.FutureConverters.FutureOps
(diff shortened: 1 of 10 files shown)


ennru:
As an alternative to this, the combination could be pre-packaged for now and get combined implementation with a specific stage later.

ennru:
Re-targeted this PR to introduce new sinks for the produce-and-commit combination.

The existing `committableSink`s could be deprecated now.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
```diff
@@ -152,6 +153,116 @@ object Producer {
       producer: org.apache.kafka.clients.producer.Producer[K, V]
   ): Sink[Envelope[K, V, ConsumerMessage.Committable], CompletionStage[Done]] = committableSink(settings, producer)
 
+  /**
+   * Create a sink that is aware of the [[ConsumerMessage.CommittableOffset committable offset]]
```
seglo: Typo `[[ConsumerMessage.CommittableOffset committable offset]]`.  The other new factory methods in the Scala and Java DSLs have the same typo.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
```diff
@@ -152,6 +153,116 @@ object Producer {
...
+   * - [[akka.kafka.ProducerMessage.Message Message]] publishes a single message to its topic, and commits the offset
+   *
+   * - [[akka.kafka.ProducerMessage.MultiMessage MultiMessage]] publishes all messages in its `records` field, and commits the offset
+   *
+   * - [[akka.kafka.ProducerMessage.PassThroughMessage PassThroughMessage]] does not publish anything, but commits the offset
+   *
+   * Note that there is a risk that something fails after publishing but before
+   * committing, so it is "at-least once delivery" semantics.
+   */
+  def sinkWithOffsetContext[K, V, IN <: Envelope[K, V, _], C <: CommittableOffset](
```
seglo: Should it be called `committingSinkWithOffsetContext`?  I know it's a mouthful..

core/src/main/scala/akka/kafka/javadsl/Producer.scala
ennru: It's not a typo a link to a class or method can use any text, but I figured these links should link to `Committable` instead.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
ennru: I changed to `sinkWithCommitting` and `sinkWithCommittingOffsetContext` instead.

ennru:
I had to upgrade to Scala 2.13.1 as scaladoc hit some bug.

core/src/main/scala/akka/kafka/javadsl/Producer.scala
seglo: Got it.  I misunderstood the scaladocs syntax.

PR #936, opened 2019-10-16
[WIP] Testkit: start configurable Testcontainer via mixin

ennru:
## Purpose

The current Testcontainers helpers in Alpakka Kafka Testkit always start the Kafka container with standard configuration.
Certain tests require a different configuration of the Kafka broker.

WIP: I haven't tried this any further

## References

Suggested in #933

PR #939, opened 2019-10-17
Use testcontainers for multi-broker integration tests and benchmarks

seglo:
## Purpose

The `sbt-docker-compose` plugin is abandoned and difficult to manage.  The only reason it was used was to support multi-broker Kafka clusters for integration testing.  Now that we can setup equivalent clusters using testcontainers there's no need to include this plugin.

## References

* https://github.com/testcontainers/testcontainers-java/pull/1984
* https://github.com/akka/alpakka-kafka/pull/936

## Changes

* Integration Tests: Set minimum in sync replicas for topics to 2 so produced messages get retried when a broker is unexpectedly shutdown
* Integration Tests: Use `takeWhile` to signal consumer stream shutdowns.
* Uses testcontainers-java to spin up Kafka clusters.  This required changes to `KafkaContainer`.  I created an [upstream PR to add this support to testcontainer-java](https://github.com/testcontainers/testcontainers-java/pull/1984), but I've copied the implementations into our internal testkit DSL in the meantime.
* Add `TestcontainersKafkaPerClassLike` to set testcontainer lifetime to test class
* Remove `sbt-docker-compose` and `sbt-build-info` SBT plugins
* Testcontainers settings `KafkaTestkitTestcontainersSettings`

TODO

* ~~Enable logs from running containers?~~
* Compare benchmark test results before and after this change (after merge)


testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainer.java
```diff
@@ -0,0 +1,179 @@
...
+                    + "/etc/confluent/docker/run")
+            .exec();
+
+    try {
+      dockerClient
+          .execStartCmd(execCreateCmdResponse.getId())
+          .exec(new ExecStartResultCallback())
+          .awaitStarted(10, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
seglo: @ennru This was my crappy workaround for the Lombok sneakythrows usage in the upstream PR.  sneakythrows allows the user to not define checked exceptions (kind of like how Scala methods work), but if you don't have Lombok available at compile time with your build tool then the sneakythrows attribute can't be used, which means the exception must be handled, which I can't add as a checked exception in this case or I wouldn't be able to override this method `containerIsStarting`.  Does that make sense?

seglo: https://github.com/seglo/testcontainers-java/blob/seglo/multi-broker-kafka/modules/kafka/src/main/java/org/testcontainers/containers/KafkaContainer.java#L95..L120

seglo: Discussed offline.  Will rethrow `RuntimeException` with inner throwable.

testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private Stream<GenericContainer> allContainers() {
+    Stream<GenericContainer> genericBrokers = this.brokers.stream().map(b -> (GenericContainer) b);
+    Stream<GenericContainer> zookeeper = Stream.of(this.zookeeper);
+    return Stream.concat(genericBrokers, zookeeper);
+  }
+
+  public void startAll() {
+    allContainers().parallel().forEach(GenericContainer::start);
```
bsideup: FYI there is `Startables.deepStart(allContainers()).{get, join}()`, that takes `GenericContainer#dependsOn` into consideration, but otherwise starts everything in parallel (without using the default forkjoin pool as `.parallel()` does)

seglo: Thanks for checking out the PR!  I'll follow up on the upstream `testcontainers-java` PR.

docs/src/main/paradox/testing.md
```diff
@@ -123,6 +123,30 @@ The Testcontainers support is new to Alpakka Kafka since 1.0.2 and may evolve a
 
 @@@
```
ennru: You may remove this section now.

testkit/src/main/java/akka/kafka/testkit/javadsl/TestcontainersKafkaTest.java
```diff
@@ -22,23 +23,38 @@
  */
 public abstract class TestcontainersKafkaTest extends KafkaTest {
```
ennru: How do users select between per-test-run and per-class Kafka brokers with this?

testkit/src/main/scala/akka/kafka/testkit/KafkaTestkitTestcontainersSettings.scala
```diff
@@ -0,0 +1,116 @@
...
+ * Copyright (C) 2016 - 2019 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.kafka.testkit
+
+import akka.actor.ActorSystem
+import com.typesafe.config.Config
+import org.testcontainers.containers.GenericContainer
+
+class KafkaTestkitTestcontainersSettings private (val confluentPlatformVersion: String,
```
ennru: `final`?

testkit/src/main/scala/akka/kafka/testkit/KafkaTestkitTestcontainersSettings.scala
```diff
@@ -0,0 +1,116 @@
...
+   */
+  def getNumBrokers(): Int = numBrokers
+
+  /**
+   * Java Api
+   */
+  def getInternalTopicsReplicationFactor(): Int = internalTopicsReplicationFactor
+
+  /**
+   * Replaces the default Confluent Platform Version
```
ennru:
```suggestion
   * Sets the Confluent Platform Version
```

testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
seglo: Updated `KafkaContainers` to use `dependsOn` and `deepStart` as a `Startable`.

testkit/src/main/java/akka/kafka/testkit/javadsl/TestcontainersKafkaTest.java
seglo: For per class they would include the `void afterClass()` implementation to stop the cluster and for all tests they would omit it.  I created an implementation for the docs (`TestkitTestcontainersTest`) that demonstrates the former, but I'll add a comment to indicate how it can be used for all tests too.

testkit/src/main/scala/akka/kafka/testkit/internal/KafkaTestKit.scala
```diff
@@ -5,11 +5,12 @@
 
 package akka.kafka.testkit.internal
 
-import java.util
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.{Arrays, Properties}
 
+import scala.collection.JavaConverters._
```
ennru: I have to re-learn as well, this should be `scala.jdk.CollectionConverters._` now.

docs/src/main/paradox/testing-testcontainers.md
```diff
@@ -0,0 +1,84 @@
+---
+project.description: Alpakka Kafka provides Testcontainers support for running a Kafka cluster locally using Docker containers.
+---
+# Testing with a Docker Kafka cluster
```
ennru: This becomes the title in the ToC, as well. 
Maybe "Testing using Testcontainers".
Or have it as a sub-page to `testing.md` might be even better.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
```diff
@@ -5,12 +5,12 @@
...
 
-import akka.dispatch.ExecutionContexts
 import akka.annotation.ApiMayChange
-import akka.{Done, NotUsed}
+import akka.dispatch.ExecutionContexts
 import akka.kafka.CommitterSettings
-import akka.kafka.ConsumerMessage.{Committable, CommittableOffset, CommittableOffsetBatch}
+import akka.kafka.ConsumerMessage.{Committable, CommittableOffsetBatch}
 import akka.stream.scaladsl.{Flow, FlowWithContext, Keep, Sink}
+import akka.{Done, NotUsed}
```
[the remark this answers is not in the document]
seglo: Probably got shuffled in a previous commit.  I'll revert.

docs/src/main/paradox/testing-testcontainers.md
seglo: I may have misunderstood what you meant about having it as a separate page.  I thought you wanted it to be highlighted in the ToC as well, otherwise why separate it?

build.sbt
```diff
@@ -243,6 +240,7 @@ lazy val testkit = project
         "EmbeddedKafkaJunit4Test.java"
       } else (Test / unmanagedSources / excludeFilter).value
     },
+    Compile / compile / scalacOptions += "-P:silencer:globalFilters=scala.jdk.CollectionConverters",
```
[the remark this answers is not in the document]
seglo: I assumed `scala.jdk.CollectionConverters` wouldn't have been available in 2.11, but I was wrong.  I'll update the converter usages and remove this silencer exception.

testkit/src/main/scala/akka/kafka/testkit/KafkaTestkitTestcontainersSettings.scala
```diff
@@ -0,0 +1,135 @@
...
+
+package akka.kafka.testkit
+
+import java.util.function.Consumer
+
+import akka.actor.ActorSystem
+import com.typesafe.config.Config
+import org.testcontainers.containers.{GenericContainer, KafkaContainer}
+
+final class KafkaTestkitTestcontainersSettings private (
```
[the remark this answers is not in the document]
seglo: It's nested under the testkit, so I thought it was appropriate.  It's a mouthful though.  I could go either way.

testkit/src/main/scala/akka/kafka/testkit/KafkaTestkitTestcontainersSettings.scala
```diff
@@ -0,0 +1,135 @@
...
+   */
+  def withInternalTopicsReplicationFactor(internalTopicsReplicationFactor: Int): KafkaTestkitTestcontainersSettings =
+    copy(internalTopicsReplicationFactor = internalTopicsReplicationFactor)
+
+  /**
+   * Java Api
+   *
+   * Replaces the default Kafka testcontainers configuration logic
+   */
+  def withConfigureKafkaJava(
```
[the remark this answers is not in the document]
seglo: Ok.  I guess I need to add a Java Api method for `configureZooKeeper` too.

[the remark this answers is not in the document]
seglo: Implemented `configureZooKeeperConsumer`

docs/src/main/paradox/testing-testcontainers.md
seglo: Resolved offline.

testkit/src/main/scala/akka/kafka/testkit/KafkaTestkitTestcontainersSettings.scala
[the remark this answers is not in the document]
seglo: I'll keep it as-is for now.  We can update it before release if necessary.

Issue #680, opened 2018-12-07
Remove deprecated API


ennru:
APIs deprecated prior to 1.0 should be removed prior to releasing 1.1 milestones.

Philippus:
does this include deprecations marked since `1.0-RC1` or `1.0-M1` as well?

ennru:
We'll release a 1.1 with just the upgrade to the Kafka client 2.1.1 and no other changes. So we defer removing deprecations to a later version.

ennru:
We keep those around until 2.0 which is not too far away.

ennru:
Fixed with #945. We leave #948 unchanged for 2.x

PR #945, opened 2019-10-22
Remove methods deprecated before 1.0

ennru:
## Purpose

Remove methods that were deprecated before 1.0.

## References

References #680

## Changes

- [ ] Should the `Producer.flow` be removed? (we recommend `flexiFlow` since 0.21)


seglo:
I think we should swap out the default `Producer.flow` impl with the `flexiFlow`, and then deprecate or remove `flexiFlow`.

ennru:
I agree that would be best. There is some value in the `flow`: its type signature is simpler which makes it a bit easier to use, especially from Java.

Now that I look at it again, the Java signature should have been
```
def flexiFlow[K, V, PassThrough, IN <: Envelope[K, V, PassThrough]](
      settings: ProducerSettings[K, V]
  ): Flow[IN, Results[K, V, PassThrough], NotUsed]
```

ennru:
Let's merge this and discuss the `flow` and `flexiFlow` in an issue.

Issue #911, opened 2019-09-26
User-defined seeking for subscriptions


ennru:
### Short description

It might be useful to allow access to the `RestrictedConsumer` when subscribing so that special seeking can be achieved (eg. seek-to-beginning).

### Details

The subscription would be enhanced to send in a function `(Set[TopicPartition], RestrictedConsumer) => Unit` so that the user may pass it for their need to seek to a specific position before any data is received. That function would be applied when the first `onPartitionsAssigned` is called from the Kafka client.

jewertow:
I would like to do it. I could prepare PR within 1-2 weeks.

ennru:
This can now be implemented by the user by leveraging the `PartitionAssignmentHandler` added in #949

Issue #843, opened 2019-07-02
Resource handling via partition assignment handler


ennru:
# Partitioned source: explore the usefulness of partition assignment handler

Alpakka Kafka's partitioned sources are the current tool for use-cases where actions need to be taken per partition when reading from Kafka. A common use-case is to handle resources connected to the assigned partitions. 
As the partitioned sources are quite cumbersome to understand and use, other solutions to their use-cases should be explored.

## Resource handling via partition assignment handler

The `PartitionAssignmentHandler` in [#761](https://github.com/akka/alpakka-kafka/pull/761) could be used to open and close resources connected to assigned partitions. This would work with a flat stream with just a regular source.

### Potential improvements

1. Resources can be closed on revoking and the assignment to a new consumer won't happen before closing is finished.
1. Simpler stream structure
1. Access to `DrainingControl` (as it works with flat streams)

### Potential negative changes

1. Handing out a blocking handler introduces a dangerous API.

PR #949, opened 2019-10-23
Consumer: make partition assignment handler public API

### LINKED ISSUE #539: Rebalance event listeners don't respect ConsumerRebalanceListener contract

ennru:
## Purpose

Allow users to pass a `PartitionAssignmentHandler` which allows for hooking into Kafka rebalances.

Potential use cases may include
* direct control over committing via the Kafka consumer API
* store offsets to an external storage
* seek offsets on assignment in a different way (see #911)

## References

Fixes #539, #843, #911
#761

## Changes

* Let subscriptions pass a `PartitionAssignmentHandler`
* Add `javadsl.PartitionAssignmentHandler`
* Add `seekToBeginning` and `seekToEnd` (usful for #911)
* Test cases in Scala an Java
* Upgrade Hamcrest 2.1 to 2.2

## TODO

- [ ] Documentation

### DIFF
--- a/core/src/main/mima-filters/1.1.0.backwards.excludes/PR949-partition-assignment-handler.excludes
+++ b/core/src/main/mima-filters/1.1.0.backwards.excludes/PR949-partition-assignment-handler.excludes
@@ -1,0 +1,8 @@
+# Internal API
+ProblemFilters.exclude[Problem]("akka.kafka.Subscriptions#TopicSubscription.*")
+ProblemFilters.exclude[Problem]("akka.kafka.Subscriptions#TopicSubscriptionPattern.*")
+ProblemFilters.exclude[MissingTypesProblem]("akka.kafka.Subscriptions$TopicSubscription$")
+ProblemFilters.exclude[MissingTypesProblem]("akka.kafka.Subscriptions$TopicSubscriptionPattern$")
+# Sealed trait
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.AutoSubscription.partitionAssignmentHandler")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.AutoSubscription.withPartitionAssignmentHandler")
--- a/core/src/main/scala/akka/kafka/RestrictedConsumer.scala
+++ b/core/src/main/scala/akka/kafka/RestrictedConsumer.scala
@@ -53,4 +53,14 @@ final class RestrictedConsumer(consumer: Consumer[_, _], duration: java.time.Dur
    * See [[org.apache.kafka.clients.consumer.KafkaConsumer#seek(TopicPartition, Long)]]
    */
   def seek(tp: TopicPartition, offset: Long): Unit = consumer.seek(tp, offset)
+
+  /**
+   * See [[org.apache.kafka.clients.consumer.KafkaConsumer#seekToBeginning(java.util.Collection[TopicPartition])]]
+   */
+  def seekToBeginning(tps: java.util.Collection[TopicPartition]): Unit = consumer.seekToBeginning(tps)
+
+  /**
+   * See [[org.apache.kafka.clients.consumer.KafkaConsumer#seekToEnd(java.util.Collection[TopicPartition])]]
+   */
+  def seekToEnd(tps: java.util.Collection[TopicPartition]): Unit = consumer.seekToEnd(tps)
 }
--- a/core/src/main/scala/akka/kafka/RestrictedConsumer.scala
+++ b/core/src/main/scala/akka/kafka/RestrictedConsumer.scala
@@ -6,7 +6,7 @@
 package akka.kafka
 
 import akka.annotation.ApiMayChange
-import org.apache.kafka.clients.consumer.{Consumer, OffsetAndMetadata}
+import org.apache.kafka.clients.consumer.{Consumer, OffsetAndMetadata, OffsetAndTimestamp}
 import org.apache.kafka.common.TopicPartition
 
 /**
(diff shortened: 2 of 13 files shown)


core/src/main/scala/akka/kafka/RestrictedConsumer.scala
```diff
@@ -53,4 +53,14 @@ final class RestrictedConsumer(consumer: Consumer[_, _], duration: java.time.Dur
...
+  /**
+   * See [[org.apache.kafka.clients.consumer.KafkaConsumer#seekToBeginning(java.util.Collection[TopicPartition])]]
+   */
+  def seekToBeginning(tps: java.util.Collection[TopicPartition]): Unit = consumer.seekToBeginning(tps)
+
+  /**
+   * See [[org.apache.kafka.clients.consumer.KafkaConsumer#seekToEnd(java.util.Collection[TopicPartition])]]
+   */
+  def seekToEnd(tps: java.util.Collection[TopicPartition]): Unit = consumer.seekToEnd(tps)
 }
```
seglo: I suggest we also add `KafkaConsumer.offsetsForTimes` while we're updating the `RestrictedConsumer`.  It's a blocking call.  It's useful for apps running in multi-dc Kafka environments.

core/src/main/scala/akka/kafka/javadsl/PartitionAssignmentHandler.scala
```diff
@@ -0,0 +1,53 @@
...
+package akka.kafka.javadsl
+
+import akka.kafka.RestrictedConsumer
+import akka.annotation.ApiMayChange
+import org.apache.kafka.common.TopicPartition
+
+/**
+ * The API is new and may change in further releases.
+ *
+ * Allows to execute user code when Kafka rebalances partitions between consumers, or an Alpakka Kafka consumer is stopped.
```
seglo:
```suggestion
 * Allows the user to execute user code when Kafka rebalances partitions between consumers, or an Alpakka Kafka consumer is stopped.
```

core/src/main/scala/akka/kafka/javadsl/PartitionAssignmentHandler.scala
```diff
@@ -0,0 +1,53 @@
...
+ * Use with care: These callbacks are called synchronously on the same thread Kafka's `poll()` is called.
+ * A warning will be logged if a callback takes longer than the configured `partition-handler-warning`.
+ *
+ * There is no point in calling `CommittableOffset`'s commit methods as their committing won't be executed as long as any of
+ * the callbacks in this class are called.
```
seglo: Could mention that `RestrictedConsumer.commitSync` is available though.

core/src/main/scala/akka/kafka/javadsl/PartitionAssignmentHandler.scala
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * This complements the methods of Kafka's [[org.apache.kafka.clients.consumer.ConsumerRebalanceListener ConsumerRebalanceListener]] with
+ * an `onStop` callback.
```
seglo: To match the scaladsl comments.

```suggestion
 * an `onStop` callback which is called before `Consumer.close`.
```

core/src/main/scala/akka/kafka/scaladsl/PartitionAssignmentHandler.scala
```diff
@@ -20,15 +20,15 @@ import org.apache.kafka.common.TopicPartition
...
  */
 @ApiMayChange
 trait PartitionAssignmentHandler {
 
   /**
    * See [[org.apache.kafka.clients.consumer.ConsumerRebalanceListener#onPartitionsRevoked]]
    *
-   * @param revokedTps The list of partitions that were assigned to the consumer on the last rebalance
+   * @param revokedTps The list of partitions that were revoked from the consumer
    * @param consumer A restricted version of the internally used [[org.apache.kafka.clients.consumer.Consumer Consumer]]
```
seglo: It's not really a `KafkaConsumer` though in the `is a` sense.  Should we link to `RestrictedConsumer` instead, or in addition, to this?

core/src/main/mima-filters/1.1.0.backwards.excludes/PR949-partition-assignment-handler.excludes
```diff
@@ -0,0 +1,8 @@
+# Internal API
```
seglo: I like the PR-based path convention.  I'll do that in other PRs too (#930)

PR #952, opened 2019-10-24
Pass ProducerSettings to stage

seglo:
## Purpose

Pass the `ProducerSettings` object directly into producer stages so that the stage has full access to the producer factory helper methods.  It also saves on the number of parameters that need to be passed to these stages.

I also took the opportunity to add more producer factory options to `ProducerSettings` so that it's not necessary to maintain overloads to pass external or shared `KafkaProducer` instances to `Producer` flows and sinks.  I deprecated methods that take these overloads.

## References

* Preliminary work for asynchronous producer creation in #930 

## Changes

* Pass ProducerSettings to stage.
* Move producer factories to settings
* Deprecate flow/sink overloads that take producers

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -22,6 +22,10 @@ akka.kafka.producer {
   # Duration to wait for `KafkaProducer.close` to finish.
   close-timeout = 60s
 
+  # Call `KafkaProducer.close` when the stream is shutdown. This is important to override to false
+  # when the producer instance is shared across multiple producer stages.
+  close-on-producer-stop = true
+
   # Fully qualified config path which holds the dispatcher configuration
   # to be used by the producer stages. Some blocking may occur.
   # When this value is empty, the dispatcher configured for the stream
--- a/core/src/main/scala/akka/kafka/ProducerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ProducerSettings.scala
@@ -61,6 +61,7 @@ object ProducerSettings {
       "Value serializer should be defined or declared in configuration"
     )
     val closeTimeout = config.getDuration("close-timeout").asScala
+    val closeOnProducerStop = config.getBoolean("close-on-producer-stop")
     val parallelism = config.getInt("parallelism")
     val dispatcher = config.getString("use-dispatcher")
     val eosCommitInterval = config.getDuration("eos-commit-interval").asScala
@@ -69,11 +70,12 @@ object ProducerSettings {
       keySerializer,
       valueSerializer,
       closeTimeout,
+      closeOnProducerStop,
       parallelism,
       dispatcher,
       eosCommitInterval,
       enrichAsync = None,
-      ProducerSettings.createKafkaProducer
+      producerFactorySync = None
     )
   }
 
@@ -156,7 +158,6 @@ object ProducerSettings {
     new KafkaProducer[K, V](settings.getProperties,
                             settings.keySerializerOpt.orNull,
                             settings.valueSerializerOpt.orNull)
-
 }
 
 /**
@@ -172,12 +173,19 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
     val keySerializerOpt: Option[Serializer[K]],
     val valueSerializerOpt: Option[Serializer[V]],
     val closeTimeout: FiniteDuration,
+    val closeProducerOnStop: Boolean,
     val parallelism: Int,
     val dispatcher: String,
     val eosCommitInterval: FiniteDuration,
     val enrichAsync: Option[ProducerSettings[K, V] => Future[ProducerSettings[K, V]]],
-    val producerFactory: ProducerSettings[K, V] => Producer[K, V]
+    val producerFactorySync: Option[ProducerSettings[K, V] => Producer[K, V]]
 ) {
+
+  @deprecated(
+    "Use createKafkaProducer(), createKafkaProducerAsync(), or createKafkaProducerCompletionStage() to get a new KafkaProducer",
+    "1.1.1"
+  )
+  def producerFactory: ProducerSettings[K, V] => Producer[K, V] = _ => createKafkaProducer()
 
   /**
    * A comma-separated list of host/port pairs to use for establishing the initial connection to the Kafka cluster.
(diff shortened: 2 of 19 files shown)


core/src/main/scala/akka/kafka/internal/DefaultProducerStage.scala
```diff
@@ -28,14 +29,13 @@ import scala.util.{Failure, Success, Try}
  */
 @InternalApi
 private[kafka] class DefaultProducerStage[K, V, P, IN <: Envelope[K, V, P], OUT <: Results[K, V, P]](
-    val closeTimeout: FiniteDuration,
-    val closeProducerOnStop: Boolean,
-    val producerProvider: ExecutionContext => Future[Producer[K, V]]
+    val settings: ProducerSettings[K, V],
+    val closeProducerOnStop: Boolean // TODO: when is this ever `False`?  is it just for testing?
```
ennru: It is false if you share the producer. You may need to add that to the settings if those are to be used to share the producer.

seglo: Thanks.

core/src/main/resources/reference.conf
```diff
@@ -22,6 +22,9 @@ akka.kafka.producer {
   # Duration to wait for `KafkaProducer.close` to finish.
   close-timeout = 60s
 
+  # Call `KafkaProducer.close` when the stream is shutdown.
```
ennru: add "important if the producer instance is shared" or something

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
```diff
@@ -351,39 +339,21 @@ object Producer {
...
-   *
-   * Supports sharing a Kafka Producer instance provided from a future.
-   */
-  def flexiFlow[K, V, PassThrough](
-      settings: ProducerSettings[K, V],
-      producer: Future[org.apache.kafka.clients.producer.Producer[K, V]]
   ): Flow[Envelope[K, V, PassThrough], Results[K, V, PassThrough], NotUsed] = {
+    val settingsWithProducer = settings
+      .withProducerFactory(_ => producer)
+      .withCloseProducerOnStop(false)
```
ennru: This should be `withProducer` now.

core/src/main/scala/akka/kafka/scaladsl/Producer.scala
```diff
@@ -321,12 +308,13 @@ object Producer {
       settings: ProducerSettings[K, V],
       producer: org.apache.kafka.clients.producer.Producer[K, V]
   ): Flow[Message[K, V, PassThrough], Result[K, V, PassThrough], NotUsed] = {
+    val settingsWithProducer = settings
+      .withProducerFactory(_ => producer)
+      .withCloseProducerOnStop(false)
```
ennru: use `withProducer`

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -86,6 +86,8 @@ object Transactional {
       transactionalId: String
   ): Flow[Envelope[K, V, ConsumerMessage.PartitionOffset], Results[K, V, ConsumerMessage.PartitionOffset], NotUsed] = {
     require(transactionalId != null && transactionalId.length > 0, "You must define a Transactional id.")
+    require(settings.producerFactory.isDefined, "You cannot use a shared or external producer factory.")
+    require(settings.producerFactoryAsync.isDefined, "You cannot use a shared or external async producer factory.")
```
ennru: So this won't work with Akka Discovery? Isn't it OK as long as it creates a new instance? Guarding against `withProducer` is the important bit, right?

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -96,10 +98,7 @@ object Transactional {
     val flow = Flow
       .fromGraph(
         new TransactionalProducerStage[K, V, ConsumerMessage.PartitionOffset](
-          txSettings.closeTimeout,
-          closeProducerOnStop = true,
-          producerProvider = (ec: ExecutionContext) => txSettings.createKafkaProducerAsync()(ec),
-          settings.eosCommitInterval
+          settings = txSettings
```
ennru: No need for named parameters anymore.

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
seglo: This should work with Akka Discovery.  The internal `producerFactorySync` and `producerFactoryAsync` values are optional and only set when the user uses a `withProducer` method.  They're not used internally to create a producer by default.  We'll know for sure once the tests aren't blocked by other issues.

ennru: Ah, you're right. I mixed up the `enrichAsync` and `producerFactoryAsync`.
A feature we tried to unlock with the producer factory was the possibility to wrap the producer with the [OpenTracing](https://github.com/opentracing-contrib/java-kafka-client) decorator.

seglo:
Ready for a final review.

core/src/main/scala/akka/kafka/ProducerSettings.scala
```diff
@@ -349,7 +399,11 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
    * (without blocking for `enriched`).
    */
   def createKafkaProducerAsync()(implicit executionContext: ExecutionContext): Future[Producer[K, V]] =
-    enriched.map(producerFactory)
+    (producerFactoryAsync, producerFactorySync) match {
+      case (Some(asyncFactory), _) => asyncFactory(executionContext)
```
ennru: What is the use case for the `asyncFactory`?
Why doesn't it get the enriched settings passed?

seglo: I copied this from the `Transactional.flow` implementation that's been been removed in this PR. 
 The "async" producer factory was passed to `TransactionalProducerStage`: https://github.com/akka/alpakka-kafka/pull/952/files#diff-832423d99d7759061130b358267d97a3L101

I don't know if we need this.  We no longer require it internally.

ennru: It seems irregular to me, remove it.

seglo: I [re-]discovered the documentation example used to point to this overload that takes a `Future[KafkaProducer[K,V]]` that was added with the discovery PR (https://github.com/akka/alpakka-kafka/pull/952/files#diff-832205ad71b0f8522d293275289adc35L60).  It seems like the only use for this is if you were to create a `KafkaProducer` asynchronously externally from Alpakka Kafka, but I'm not sure why we need to support that.

I'll revert the documentation to showing how to share an external `KafkaProducer` synchronously.

seglo: I guess it's because `ProducerSettings` exposes `createKafkaProducerAsync` and `createKafkaProducerCompletionStage`.  I'm using this internally for the transactions PR.  Maybe it would be useful to share to the user if they want to get an asynchronous producer for some other purpose?

seglo: I'll hold off removing this for now.  

I squashed the commits in this branch to make rebasing easier for the transactions PR.

ennru: Ah, thank you for digging. Yes, I added those to allow passing the enriched producer to the factory methods with a shared producer. With this solution for passing the producer factory, we don't need those.

seglo: Ok

Issue #942, opened 2019-10-21
`RestartSource`-wrapped consumers can't batch commits


ennru:
The commit batching expects exactly one consumer actor per group ID as offsets are sent back to that consumer actor for committing:

https://github.com/akka/alpakka-kafka/blob/d93b4b81bcdf68251b7e205181717fd6d8914441/core/src/main/scala/akka/kafka/internal/MessageBuilder.scala#L207-L217

This does not work when the consumer is wrapped in a `RestartSource` and the commits are batched outside of it. Whenever the `RestartSource` fails and creates a new consumer, those offsets will refer to a new consumer actor.

If ordering can be assumed, it might be correct to flush the offsets that arrived for a different actor for the same group ID.

brendan-mcadams-exa:
Roughly , our code that is seeing this issue looks as such:
```scala
val dataSource = RestartSource.withBackoff(
  minBackoff = 3.seconds,
  maxBackoff = 30.seconds,
  randomFactor = 0.2
) { () =>
 val kafkaConsumer = // Our class that sets up and returns a Kafka Consumer Source
 kafkaConsumer mapMaterializedValue { c => consumerControl.set(c) } 
}

val dataFuture = dataSource
  .toMat(Committer.sink(CommitterSettings(system)))(Keep.right)
  .run()
```

We do attach some flow steps into the source but it is still returned as a source.
It would appear intermittently, in circumstances  where the flow restarted, that we get the above listed error when attempting to commit. @ennru's diagnosis pretty accurately explains what we're seeing.

ennru:
I dug into this and found a good solution (and fixed another bug) in #953

PR #953, opened 2019-10-25
CommittableOffsetBatch: Handle multiple committer refs

ennru:
## Purpose

Harden the `CommittableOffsetBatch` so it can aggregate offsets with committer refs from several consumer stages/actors.

## References

Problem reported in #942 

## Changes

(Reviewing per commit might make sense)

* Test various scenarios where different consumer stages/actors feed a commit batch
* Fix `equals` in `KafkaAsyncConsumerCommitterRef`
* Keep committer refs per GroupId-Topic-Partition instead of just per group and send per offsets per that combination to the committing actor
* Store the to-be-committed offsets as tuple lists in the actor

## Background Context

The `CommittableOffsetBatch` guarded against batching offsets for the same group ID that came from different consumer stages/actors. That becomes a problem for unusual flow topologies, and when a `RestartSource` is used.


core/src/main/scala/akka/kafka/internal/CommittableSources.scala
```diff
@@ -117,14 +117,12 @@ private[kafka] final class CommittableSubSource[K, V](
 /**
  * Internal API.
  *
- * [[InternalCommitter]] implementation for committable sources.
- *
- * Sends [[akka.kafka.internal.KafkaConsumerActor.Internal.Commit Commit]] messages to the consumer actor.
- *
- * This should be case class to be comparable based on consumerActor and commitTimeout. This comparison is used in [[CommittableOffsetBatchImpl]].
+ * Sends [[akka.kafka.internal.KafkaConsumerActor.Internal.Commit]] and
+ * [[akka.kafka.internal.KafkaConsumerActor.Internal.CommitSingle]] messages to the consumer actor.
```
seglo: We could add `CommitWithoutReply` too for completeness.

Issue #668, opened 2018-11-30
Draining Control throws exception when used immediately after stream start


2m:
Exception is:

```
java.lang.IllegalStateException: not yet initialized: only setHandler is allowed in GraphStageLogic constructor
```

2m:
https://github.com/akka/alpakka-kafka/pull/665/files#r237763633

seglo:
```
[info] - must access metrics *** FAILED *** (80 milliseconds)
[info]   java.lang.IllegalStateException: not yet initialized: only setHandler is allowed in GraphStageLogic constructor. To access materializer use Source/Flow/Sink.setup factory
[info]   at akka.stream.stage.GraphStageLogic.interpreter(GraphStage.scala:392)
[info]   at akka.stream.stage.GraphStageLogic.materializer(GraphStage.scala:403)
[info]   at akka.kafka.internal.BaseSingleSourceLogic.executionContext(BaseSingleSourceLogic.scala:35)
[info]   at akka.kafka.internal.MetricsControl$class.metrics(ControlImplementations.scala:88)
[info]   at akka.kafka.internal.BaseSingleSourceLogic.metrics(BaseSingleSourceLogic.scala:25)
[info]   at akka.kafka.scaladsl.IntegrationSpec$$anonfun$2$$anonfun$apply$mcV$sp$11$$anonfun$apply$24.apply(IntegrationSpec.scala:400)
[info]   at akka.kafka.scaladsl.IntegrationSpec$$anonfun$2$$anonfun$apply$mcV$sp$11$$anonfun$apply$24.apply(IntegrationSpec.scala:388)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
[info]   at akka.kafka.scaladsl.IntegrationSpec$$anonfun$2$$anonfun$apply$mcV$sp$11.apply(IntegrationSpec.scala:388)
[info]   at akka.kafka.scaladsl.IntegrationSpec$$anonfun$2$$anonfun$apply$mcV$sp$11.apply(IntegrationSpec.scala:388)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[...]

PR #958, opened 2019-10-31
DrainingControl: expose failures on overwritten methods

### LINKED ISSUE #956: Return a Failure in DrainingControl.shutdown when a stream has already failed in
### Short description

The `DrainingControl.shutdown()` will always return a `Success[Done]`.  We should consider returning a `Failure(Throwable)` in situations where the stream has already failed.

ennru:
## Purpose

Expose stream completion errors in the `DrainingControl` methods that override `Control`.

## References

Fixes #956

## Changes

* Update Scaladoc for `Control`
* Combine `shutdown` and `isShutdown` with the stream completion future to make it fail for any errors
* Mark `DrainingControl.shutdown` as deprecated as users should prefer the `drainAndShutdown` solution

## Background Context

`DrainingControl` inherits methods from `Control` and did not add any value over the finished future to those, even though it has knowledge of the stream's completion status.

### DIFF
--- a/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
+++ b/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
@@ -26,20 +26,31 @@ object Consumer {
 
   /**
    * Materialized value of the consumer `Source`.
+   *
+   * See [[https://doc.akka.io/docs/alpakka-kafka/current/consumer.html#controlled-shutdown Controlled shutdown]]
    */
   trait Control {
 
     /**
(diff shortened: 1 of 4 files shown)


seglo:
Looks like a new transient exception in one of the test runs: https://travis-ci.org/akka/alpakka-kafka/jobs/605393167#L704

ennru:
That is actually https://github.com/akka/alpakka-kafka/issues/668. I haven't seen it failing before, did Travis have a very fast run?

PR #959, opened 2019-11-01
Committable: deprecate commitJavadsl and commitScaladsl

ennru:
## Purpose

Deprecate `commitJavadsl` and `commitScaladsl` methods in the `Committable` trait.

## References

References #424 

## Changes

* Update scaladoc of `Committable`
* Introduce `commitInternal` as internal API
* Deprecate `commitJavadsl` and `CommitScaladsl` with all available annotations
* Mark `CommittableOffset` and `PartitionOffset` as `sealed`
* Remove the scala-only example with complex commit aggregation from `consumer.md` 
* Remove `()` from `offsets` and `getOffsets` from `CommittableOffsetBatch`

## Background Context

With the recommended use of `Committer` the direct use of the `commitJavadsl` and `commitScaladsl` can now be deprecated and be up for removal in later versions.

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerMessage.scala
@@ -53,7 +53,7 @@ object ConsumerMessage {
     def commitScaladsl(): Future[Done]
 
     /**
-     * @deprecated use `Committer.flow` or `Committer.sink` instead of direct usage, since 1.1.1
+     * @deprecated use `Committer.flow` or `Committer.sink` instead of direct usage, since 2.0.0
      */
     @java.lang.Deprecated
     @deprecated("use `Committer.flow` or `Committer.sink` instead of direct usage", "2.0.0")
--- a/core/src/main/scala/akka/kafka/ConsumerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerMessage.scala
@@ -42,15 +42,25 @@ object ConsumerMessage {
   )
 
   /**
-   * Commit an offset that is included in a [[CommittableMessage]].
-   * If you need to store offsets in anything other than Kafka, this API
-   * should not be used.
+   * Carries offsets from Kafka for aggregation and committing by the [[scaladsl.Committer]]
+   * or [[javadsl.Committer]].
    *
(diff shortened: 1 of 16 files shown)


core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -42,16 +42,26 @@ object ConsumerMessage {
...
-   * This interface might move into `akka.stream`
+   * `Committable` may be a single offset in [[CommittableOffset]] or [[CommittableOffsetMetadata]],
+   * or a number of offsets aggregated as [[CommittableOffsetBatch]].
    */
   @DoNotInherit trait Committable {
+    @deprecated("use `Committer.flow` or `Committer.sink` instead of direct usage", "2.0.0")
     def commitScaladsl(): Future[Done]
+
+    /**
+     * @deprecated use `Committer.flow` or `Committer.sink` instead of direct usage, since 1.1.1
```
seglo: Should it be deprecated since 2.0.0?

core/src/main/scala/akka/kafka/internal/MessageBuilder.scala
```diff
@@ -138,8 +138,9 @@ private[kafka] trait OffsetContextBuilder[K, V]
 )(
     val committer: KafkaAsyncConsumerCommitterRef
 ) extends CommittableOffsetMetadata {
-  override def commitScaladsl(): Future[Done] = committer.commitSingle(this)
-  override def commitJavadsl(): CompletionStage[Done] = commitScaladsl().toJava
+  override def commitScaladsl(): Future[Done] = commitInternal()
+  override def commitJavadsl(): CompletionStage[Done] = commitInternal().toJava
+  override def commitInternal(): Future[Done] = committer.commitSingle(this)
```
seglo: Maybe just call it `commit` and mark/doc it as internal use.

ennru: As these classes are marked as internal, I did not repeat the annotations.
The name is more of an extra safeguard, keep in mind that `private[kafka]` translates to `public` for Java.

PR #968, opened 2019-11-11
WIP- Akka Cluster Sharding Group Assignor Experiment

seglo:
Experiment creating custom consumer group partition assignor.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -257,6 +257,7 @@ lazy val tests = project
     name := "akka-stream-kafka-tests",
     libraryDependencies ++= Seq(
         "com.typesafe.akka" %% "akka-discovery" % akkaVersion,
+        "com.typesafe.akka" %% "akka-cluster-sharding-typed" % akkaVersion,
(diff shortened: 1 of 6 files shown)


tayvs:
@seglo This PoC is very useful. Would it be implemented in next releases?

seglo:
@tayvs Thanks for the interest! It's something we're discussing at our engineering meeting this week. I meant to post this to my own fork for collaboration with some teammates.

ennru:
Our discussions within the core Akka team led to the creation of https://github.com/akka/akka/issues/28175 which intends to cover this use case.

chbatey:
@tayvs i'm working on https://github.com/akka/akka/issues/28175 now so if you can share your usecase it would be very helpful

tayvs:
@chbatey My case is very simple. I have sharding and kafka consumer. I want avoid unnesessery sending messages via network and just send it to a local node. To achieve this goal I would use same key for kafka and sharding, same hash function and use same amount of shards and partitions. So consumer would be consuming just a message for a shards that hosted on local node. BUT I nead some partition assignor that allows me to syncronise shrads and partitions

PR #970, opened 2019-11-14
Cleanup KafkaSpec before termination of embedded server

shobull:
## Purpose

Fix sequence of clean up operations after test with EmbeddedKafka finished. Without this fix embedded Kafka is stopped but producer & adminClient are kept in polling state which produces WARN log messages. `akka.kafka.testkit.scaladsl.TestcontainersKafkaPerClassLike` already has order of operations fixed.

## Changes

- `EmbeddedKafkaLike` switched oprations of cleanup

### DIFF
--- a/testkit/src/main/scala/akka/kafka/testkit/scaladsl/EmbeddedKafkaLike.scala
+++ b/testkit/src/main/scala/akka/kafka/testkit/scaladsl/EmbeddedKafkaLike.scala
@@ -22,7 +22,7 @@ trait EmbeddedKafkaLike extends KafkaSpec {
   }
 
   override def cleanUp(): Unit = {
-    EmbeddedKafka.stop()
     super.cleanUp()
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

seglo:
Makes sense, thanks for the PR @shobull.  It seems there's a mismatch between the user details in your commit and GitHub ID.  Were you able to successfully sign the CLA?

shobull:
> Makes sense, thanks for the PR @shobull. It seems there's a mismatch between the user details in your commit and GitHub ID. Were you able to successfully sign the CLA?

Fixed commit. Sorry for that.

ennru:
Nice catch! Thank you.

PR #991, opened 2019-11-28
Stage Logic and Actor logging with ID

seglo:
## Purpose

Prefix each Akka Stream graph stage logic and `KafkaConsumerActor` log statement with a 5 character UID to make it easier to troubleshoot logging when multiple streams or partitioned sources are being used.

## References

* Issue: #990 

## Changes

* Update all stage logics with `StageIdLogging`
* Update `KafkaConsumerActor` with `ActorIdLogging`

## Background Context

I used something like this to troubleshoot problems with partitioned sources and the transactional partitioned source PRs (#930)

Issue #526, opened 2018-06-28
SubSourceLogic may send Stop twice


ennru:
When using the `ConsumerControl` to stop a consumer the `Stop` message may be sent twice to the `KafkaConsumerActor` resulting in a dead letter.
Both `SubSourceLogic.performShutdown` and `SubSourceLogic.postStop` send `Stop`. 

```
16:51:08.800 DEBUG a.kafka.internal.SingleSourceLogic - performShutdown sending Stop
16:51:08.847 DEBUG a.kafka.internal.KafkaConsumerActor - received handled message Stop from Actor[akka://Spec/deadLetters]
16:51:08.848 DEBUG a.kafka.internal.KafkaConsumerActor - Stopping Actor[akka://Spec/system/kafka-consumer-5#-912300600]
16:51:08.848 DEBUG a.kafka.internal.SingleSourceLogic - postStop sending Stop
16:51:08.852 DEBUG a.kafka.internal.KafkaConsumerActor - received handled message Subscribe(Set(topic7),ListenerCallbacks(akka.kafka.internal.SingleSourceLogic$$Lambda$1139/181292641@17274548,akka.kafka.internal.SingleSourceLogic$$Lambda$1140/1239809534@20a96f3c)) from Actor[akka://Spec/system/StreamSupervisor-0/$$f#1905692510]
16:51:08.858 INFO  akka.actor.RepointableActorRef - Message [akka.kafka.KafkaConsumerActor$Stop$] without sender to Actor[akka://Spec/system/kafka-consumer-5#-912300600] was not delivered. [5] dead letters encountered. If this is not an expected behavior, then [Actor[akka://Spec/system/kafka-consumer-5#-912300600]] may have terminated unexpectedly, This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
``

AndreasPresthammer:
This may not be the same issue, but I observe similar behavior with the SingleSourceLogic component.

Using akka-stream-kafka v2.0.5.

BaseSingleSourceLogic.scala:114 will cause the function performShutdown() in SingleSourceLogic.scala:48 which will shut down the consumer actor.

After this is done the postStop function in SingleSourceLogic.scala:44 will send a StopFromStage message to the (already killed) consumer actor, which causes a dead letter message to appear in the logs.

Repro pseudocode:
Consumer.plainSource(..)
[...]

PR #995, opened 2019-11-29
Stage ID traceability

seglo:
## Purpose

Clean up log messages to make it easier to trace lifecycle events of stream stages and actors.

## References

* Exploration of options and continuation of work in #991

## Changes

* Log stage id of stage that sent request to stop KafkaConsumerActor
* Allow stage to override `idLogPrefix`
* Include KafkaConsumerActor number in `idLogPrefix` of sub sources
* Upgrade Starting and Shutdown logging to `INFO` and make more consistent

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
@@ -59,6 +59,7 @@ private class SubSourceLogic[K, V, Msg](
 
   private val consumerPromise = Promise[ActorRef]
   final val actorNumber = KafkaConsumerActor.Internal.nextNumber()
+  override val idLogPrefix: String = idLogPrefixWithActorId(super.idLogPrefix, actorNumber)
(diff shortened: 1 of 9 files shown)


core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -336,6 +342,10 @@ private object SubSourceLogic {
         actorNumber: Int
     ): SubSourceStageLogic[K, V, Msg]
   }
+
+  @InternalApi
+  private[kafka] def idLogPrefixWithActorId(idLogPrefix: String, actorNumber: Int): String =
+    idLogPrefix + s"KafkaConsumerActor [#$actorNumber] "
```
ennru: This should be more compact. eg. `[37b86#45]`

ennru:
I wonder if this may solve https://github.com/akka/alpakka-kafka/issues/526 ?

seglo:
> I wonder if this may solve #526 ?

Yes, I had that in mind.

Issue #978, opened 2019-11-22
Consumer drops records after rebalance


altomch:
### Versions used 

Akka version: 2.5.23
Alpakka-Kafka version: 1.1.0
Scala version: 2.12.10

### Expected Behavior

Committable Consumer shouldn't loose any record consumed from Kafka 

### Actual Behavior

We observed that after rebalance some records were dropped out from processing, exactly the same behavior that was described by @vlovgr in [#336](https://github.com/akka/alpakka-kafka/issues/336#issuecomment-410738395), we've lost 500 records from topic, it was very clear because we've consumed the data from the beginning of topic and we didn't observe any result from those records (no logs, no state changes).

### Reproducible Test Case

After deep diving into code I've end up with opening bug for  [Scala - #11807](https://github.com/scala/bug/issues/11807).

This bug causes misbehavior when records are consumed from Kafka but concatenated to broken buffer [`BaseSingleSourceLogic`](https://github.com/akka/alpakka-kafka/blob/master/core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala#L60).

ennru:
Thank you for diving into this and reporting it to the Scala issue tracker. We'll try to find a work-around for the upcoming release.

seglo:
Nice find :+1:

altomch:
Seems SubSourceLogic doesn't suffer from this issue.
SubSourceStage requests only one TopicPartition, so KafkaConsumeActor will never concatenate messages.

ennru:
We plan to release Alpakka Kafka 2.0.0-M1 #988 now and can't wait for new Scala version of 2.12 and 2.13.
To avoid hitting the `ConcatIterator` bug I created #997.

ennru:
Fixed for now with #997.
The fix targetting Scala 2.13.2 was merged, as well.

Issue #957, opened 2019-10-31
[Failure] CommittingSpec: Multiple consumers to one committer...


ennru:
...must merge multiple consumers` offsets for different partitions (#942) 

```
[info] Multiple consumers to one committer
[info] - must merge multiple consumers` offsets for different partitions (#942) *** FAILED *** (19 seconds, 660 milliseconds)
[info]   The future returned an exception of type: akka.kafka.CommitTimeoutException, with message: Kafka commit took longer than: 15 seconds. (CommittingSpec.scala:459)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.concurrent.Futures$FutureConcept.tryTryAgain$1(Futures.scala:531)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValueImpl$(Futures.scala:479)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.Futures$FutureConcept.futureValue$(Futures.scala:475)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at akka.kafka.scaladsl.CommittingSpec.$anonfun$new$44(CommittingSpec.scala:459)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
```

nikkatalnikov:
we got this error with `committablePartitionedSource`.
any progress on it?

seglo:
https://travis-ci.org/akka/alpakka-kafka/jobs/619520795#L556

ennru:
https://travis-ci.org/akka/alpakka-kafka/jobs/625389283#L983 in #1005 
https://travis-ci.org/akka/alpakka-kafka/jobs/625389285#L986 in #1005

ennru:
https://travis-ci.org/akka/alpakka-kafka/jobs/625607991#L978 in #926

ennru:
I'm trying to understand how this can be a 15s timeout https://travis-ci.org/akka/alpakka-kafka/jobs/625654499#L979-L982

[...]

Issue #895, opened 2019-09-05
[Failed] CommittingWithMockSpec should collect commits to be sent to commitAsync


ennru:
```
[info] CommittingWithMockSpec:

[info] - should commit metadata in message (469 milliseconds)

[info] - should call commitAsync for commit message and then complete future (228 milliseconds)

[info] - should fail future in case of commit fail (220 milliseconds)

[info] - should collect commits to be sent to commitAsync *** FAILED *** (10 seconds, 101 milliseconds)

[info]   java.util.concurrent.TimeoutException: Futures timed out after [10 seconds]

[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:259)

[info]   at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:263)

[info]   at scala.concurrent.Await$.$anonfun$result$1(package.scala:220)

[info]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)

[info]   at scala.concurrent.Await$.result(package.scala:146)

[info]   at akka.kafka.internal.CommittingWithMockSpec.$anonfun$new$13(CommittingWithMockSpec.scala:229)
```

https://travis-ci.org/akka/alpakka-kafka/jobs/581091376#L827

seglo:
https://travis-ci.org/akka/alpakka-kafka/jobs/592616580#L832

seglo:
https://travis-ci.org/akka/alpakka-kafka/jobs/595824299#L591

seglo:
https://travis-ci.org/akka/alpakka-kafka/jobs/640314577#L650

ennru:
https://travis-ci.org/akka/alpakka-kafka/jobs/640635808#L649 in #1026

ennru:
https://travis-ci.org/akka/alpakka-kafka/jobs/648052162#L586 in master

ignasi35:
I think the problem is on :

https://github.com/akka/alpakka-kafka/blob/4f7760f2ec10a3abd7ca049667dd85b5c0b408e3/tests/src/test/scala/akka/kafka/internal/CommittingWithMockSpec.scala#L216-L222

[...]

PR #997, opened 2019-12-02
Temporary fix for #978: avoid ConcatIterator

ennru:
## Purpose

Use a `Vector` to collect `ConsumerRecords` requested by a stage so that the concatenation in the source logic can't hit the [bug in `ConcatIterator`](https://github.com/scala/bug/issues/11807). 

## References

Reported in https://github.com/akka/alpakka-kafka/issues/978
and https://github.com/scala/bug/issues/11807 (Scala 2.12.10 and 2.13.1).

## Changes

* Use a `Vector.newBuilder` and iterative adding instead of `foldLeft` over `Iterator`s.
* Retire the work-around for a historic bug in Scala iterators


ennru:
Both Failures are #895

ennru:
Failure was #957 
https://travis-ci.org/akka/alpakka-kafka/jobs/619520795#L556

Issue #999, opened 2019-12-03
Fail build for Java formatting changes


ennru:
### Short description

Fail the CI build when the Java code formatter changes code.

### Details

The CI check phase detects Scala code which requires reformatting by Scalafmt, but Java code formatting changes do not stop the build.
This interacts in bad ways with sbt-dynver when builds are created, as any changes to the sources will be reflected in the derived version number.

ennru:
Fixed with #1003

PR #1003, opened 2019-12-04
Build: fail for unformatted Java sources

ennru:
## Purpose

Make sure the build doesn't progress if Java sources get reformatted.

## References

#999


ennru:
As more long term improvement I've added commands to the java formatter plugin in https://github.com/sbt/sbt-java-formatter/pull/53

PR #1007, opened 2019-12-16
CommittingSpec: add logging

ennru:
* Log the ask timeout's message when failing a commit
* Introduce timings in `CommittingSpec` to keep it from failing

Intends to solve #957

### DIFF
--- a/tests/src/test/scala/akka/kafka/scaladsl/CommittingSpec.scala
+++ b/tests/src/test/scala/akka/kafka/scaladsl/CommittingSpec.scala
@@ -452,7 +452,9 @@ class CommittingSpec extends SpecBase with TestcontainersKafkaLike with Inside {
           .map(_.committableOffset)
           .groupedWithin(20, 5.seconds)
           .map(CommittableOffsetBatch.apply)
+          .log("sending offset batch")
           .via(Committer.batchFlow(committerDefaults.withMaxBatch(1L)))
+          .log("offset batch done")
           .runWith(Sink.head)
 
       val batch = result.mapTo[CommittableOffsetBatchImpl].futureValue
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/akka/kafka/scaladsl/CommittingSpec.scala
```diff
@@ -435,24 +435,32 @@ class CommittingSpec extends SpecBase with TestcontainersKafkaLike with Inside {
...
         Consumer
-          .committableSource(consumerSettings, Subscriptions.assignment(new TopicPartition(topic, partition0)))
+          .committableSource(
+            consumerSettings
+              .withStopTimeout(100.millis), // this consumer actor needs to stay around to receive the commit
+            Subscriptions.assignment(new TopicPartition(topic, partition0))
+          )
           .take(10)
+          // triggers commit timeout as the actor is terminated
+          .delay(50.millis)
```
seglo: It's good that this resolves the issue, but it looks like it could still arise on systems with slow clocks?  We should try repeating this test (i.e. `sbt test thistest -- -DtimesToRepeat=100`) to see if it fails occasionally.

More long term we should find a way around the stop timeout for happy-path shutdown scenarios.

ennru: This PR fixes the test, not the issue (and added more information to the commit timeout error).
I'm not sure there is a way to fix this, for how long after emitting data should a source live to receive commits? Should the actor stay around, but the stage complete?

seglo:
Unrelated build error

https://github.com/akka/alpakka-kafka/issues/1009

Issue #826, opened 2019-06-12
[Failure] TransactionsSpec: must complete with messages filtered out and transient failure causing an abort with restartable source


ennru:
```
[info] TransactionsSpec:
[info] A consume-transform-produce cycle
[info] - must complete with messages filtered out and transient failure causing an abort with restartable source *** FAILED *** (14 seconds, 419 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (10 seconds) during expectMsg while waiting for OnNext(51)
[info]   at scala.Predef$.assert(Predef.scala:223)
[info]   at akka.testkit.TestKitBase.expectMsg_internal(TestKit.scala:418)
[info]   at akka.testkit.TestKitBase.expectMsg(TestKit.scala:395)
[info]   at akka.testkit.TestKitBase.expectMsg$(TestKit.scala:395)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:923)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.$anonfun$expectNextN$1(StreamTestKit.scala:411)
[info]   at scala.collection.Iterator.foreach(Iterator.scala:941)
[info]   at scala.collection.Iterator.foreach$(Iterator.scala:941)
[info]   at scala.collection.AbstractIterator.foreach(Iterator.scala:1429)
[info]   at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[info]   at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[info]   at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNextN(StreamTestKit.scala:411)
[info]   at akka.kafka.scaladsl.TransactionsSpec.$anonfun$new$17(TransactionsSpec.scala:199)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
[info]   at akka.kafka.scaladsl.TransactionsSpec.$anonfun$new$16(TransactionsSpec.scala:148)
```

https://travis-ci.org/akka/alpakka-kafka/jobs/544581914#L1823

seglo:
https://travis-ci.org/akka/alpakka-kafka/jobs/601873942#L455

seglo:
https://travis-ci.org/akka/alpakka-kafka/jobs/615297999#L468

[...]

PR #1025, opened 2020-01-21
Producer: avoid race between failure and upstream finish

ennru:
Decrementing `awaitingConfirmation` before the `closeAndFailStageCb` gets handled may lead to successful stage completion if upstream completes.

The new test cases fails without this fix.


seglo:
Known issues in build

* https://github.com/akka/alpakka-kafka/issues/895 x 2

seglo:
https://github.com/akka/alpakka-kafka/issues/826

Issue #1027, opened 2020-01-23
plainPartitionedManualOffsetSource hangs if a rebalance occurs while loading offsets


gabrielreid:
### Versions used 

Akka version:  2.0.0 and 2.0.1

### Expected and Actual Behavior 

It appears that partitions can be completely omitted from the output of `Consumer.plainPartitionedManualOffsetSource` if the loading of offsets is still in progress during a rebalance.

The scenario in question is as follows, in a setup with a topic 2 partitions:
1. Consumer A starts up, and is assigned partitions 0 and 1, and starts asynchronously loading the offsets for these partitions
2. Consumer B starts up
3. Consumer A gets a partition revoke for partitions 0 and 1
4. Consumer A gets a partition assignment for partition 0, and starts asynchronously loading the offsets
5. Consumer B gets a partition assignment for partition 1, and starts asynchronously loading the offsets
6. The loading of offsets for Consumer A from step 1 completes
7. The loading of offsets for Consumer B from step 5 completes
8. The loading of offsets for Consumer A from step 4 completes

After this sequence of events, it is _expected_ that Consumer A has a single running source that provides elements from partition 0,
and Consumer B has a single running source that provides elements from partition 1.

What currently happens is that Consumer A does not receive a single source, while Consumer B does receive a single source for partition 1. This means that no elements are processed for partition 0. However, it is still internally assigned to Consumer A (in a paused state), so it also doesn't get rebalanced.

The end result is also the same if the two offset-loading Futures for Consumer A complete in the reverse order.

This is a significant problem in our situation, where we're loading a snapshotted partition state that also contains the offset when a partition is assigned. We're also starting up multiple consumers at (nearly) the same time, which results in exactly the behavior above. Even if 
offset loading is quite quick, there's still always a chance of running into this situation.

I've got some test cases to reproduce this and a proposed fix which I'll put in a PR.

seglo:
Thanks for reporting this issue.  After looking at your PR I can see how the scenario you document could arise.  I think the easiest way to resolve this issue is to create a new `AsyncCallback` `onOffsetsFromExternalResponse`, that's invoked with the response from `getOffsetsFromExternal(assigned)`.  In that callback, we can filter partitions out of the offset map returned with what is currently assigned.  We'll know what's currently assigned because `partitionAssignedCB` would have been called concurrently while waiting for the external offset `Future`.

seglo:
[...]

PR #1028, opened 2020-01-23
Fix slow-loading offsets during rebalance (#1027)

gabrielreid:
Correctly loading of out-of-date offsets for
plainPartitionedManualOffsetSource when a rebalance occurs while
the offsets are still in the process of loading.

Also deal with multiple offset requests returning in a different
order than they were requested.

## Purpose

Resolves the incorrect behavior (i.e. stopping of processing) when external offset loading coincides with a consumer rebalance

## References

References #1027

## Changes

A couple of tests that fail without the fix have also been added. 

Note that the dummy consumer in `PartitionedSourceSpec` has also been changed to throw an exception if a seek is performed on an unassigned partition (as per the API of the Kafka Consumer interface).

This exception is currently only logged (and not otherwise handled) in `akka.kafka.internal.SubSourceLogic#seekAndEmitSubSources`. This seems questionable to me, but failing the whole stage due to this exception resulted in some other test failures, so I've just logged it instead.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
@@ -5,6 +5,8 @@
 
 package akka.kafka.internal
 
+import java.util.concurrent.atomic.AtomicLong
+
 import akka.NotUsed
 import akka.actor.Status
 import akka.actor.{ActorRef, ExtendedActorSystem, Terminated}
@@ -75,6 +77,9 @@ private class SubSourceLogic[K, V, Msg](
   /** Kafka has signalled these partitions are revoked, but some may be re-assigned just after revoking. */
   private var partitionsToRevoke: Set[TopicPartition] = Set.empty
 
+  /** Generate fencing tokens for async offset loading so that out-of-date loaded offsets are ignored */
+  private val offsetLoadingFencingGen = new AtomicLong()
+
   override def preStart(): Unit = {
     super.preStart()
     log.info("Starting")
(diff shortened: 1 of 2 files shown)


core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -75,6 +77,9 @@ private class SubSourceLogic[K, V, Msg](
   /** Kafka has signalled these partitions are revoked, but some may be re-assigned just after revoking. */
   private var partitionsToRevoke: Set[TopicPartition] = Set.empty
 
+  /** Generate fencing tokens for async offset loading so that out-of loaded offsets are ignored */
```
francisdb: typo: out of order?

gabrielreid: Oops, indeed. Fixed now.

core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
```diff
@@ -128,7 +136,15 @@ private class SubSourceLogic[K, V, Msg](
                 )
               )
             case Success(offsets) =>
-              seekAndEmitSubSources(formerlyUnknown, offsets)
+              if (fencingToken == offsetLoadingFencingGen.get()) {
```
seglo: I think the easiest way to resolve this issue is to create a new `AsyncCallback` `onOffsetsFromExternalResponse`, that's invoked with the response from `getOffsetsFromExternal(assigned)`. In that callback, we can filter partitions out of the offset map returned with what is currently assigned. We'll know what's currently assigned because partitionAssignedCB would have been called concurrently while waiting for the external offset `Future`.

This way we wouldn't require the fencing token and atomic reference.

seglo:
Fixed in #1029

PR #1029, opened 2020-01-24
Fix slow-loading offsets during rebalance

gabrielreid:
Correctly loading of out-of-date offsets for
plainPartitionedManualOffsetSource when a rebalance occurs while
the offsets are still in the process of loading.

Also deal with multiple offset requests returning in a different
order than they were requested.

## Purpose

Resolves the incorrect behavior (i.e. stopping of processing) when external offset loading coincides with a consumer rebalance

## References

References #1027

## Changes

A couple of tests that fail without the fix have also been added. 

Note that the dummy consumer in `PartitionedSourceSpec` has also been changed to throw an exception if a seek is performed on an unassigned partition (as per the API of the Kafka Consumer interface).

This exception is currently only logged (and not otherwise handled) in `akka.kafka.internal.SubSourceLogic#seekAndEmitSubSources`. This seems questionable to me, but failing the whole stage due to this exception resulted in some other test failures, so I've just logged it instead.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/SubSourceLogic.scala
@@ -102,6 +102,13 @@ private class SubSourceLogic[K, V, Msg](
     failStage(ex)
   }
 
+  private val onOffsetsFromExternalResponseCB = getAsyncCallback[(Set[TopicPartition], Map[TopicPartition, Long])] {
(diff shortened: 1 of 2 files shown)


gabrielreid:
Sorry about that failing compilation -- I totally neglected to try building against 2.13. I've updated the PR and it's building cleanly now.

seglo:
@gabrielreid Thanks for the bugfix!

Issue #1020, opened 2020-01-16
Java API `createCommittableOffsetBatch` should accept `Committable`


ennru:
The Java API `createCommittableOffsetBatch` requires a list of `CommittableOffset` where it should allow `Committable` (as the Scala API does in `apply`)
https://github.com/akka/alpakka-kafka/blob/086c6056dc3b3cdafd90a8ce4ed906fe298e6eb7/core/src/main/scala/akka/kafka/ConsumerMessage.scala#L184

https://github.com/akka/alpakka-kafka/blob/086c6056dc3b3cdafd90a8ce4ed906fe298e6eb7/core/src/main/scala/akka/kafka/ConsumerMessage.scala#L166

Changing this is almost binary-compatible as type parameter for `List` is not part of the message signature, and `Committable` is wider than `CommittableOffset`.

jewertow:
I will fix it.

seglo:
Fixed in #1033

PR #1033, opened 2020-01-31
Java API createCommittableOffsetBatch accepts Committable

jewertow:
## Purpose

The Java API `createCommittableOffsetBatch` requires a list of `CommittableOffset` where it should allow `Committable` (as the Scala API does in apply).

## References

#1020

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerMessage.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerMessage.scala
@@ -181,7 +181,7 @@ object ConsumerMessage {
    * Java API:
    * Create an offset batch out of a list of offsets.
    */
-  def createCommittableOffsetBatch(offsets: java.util.List[CommittableOffset]): CommittableOffsetBatch = {
+  def createCommittableOffsetBatch[T <: Committable](offsets: java.util.List[T]): CommittableOffsetBatch = {
     import scala.jdk.CollectionConverters._
     CommittableOffsetBatch(offsets.asScala.toList)
   }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -181,7 +181,7 @@ object ConsumerMessage {
    * Java API:
    * Create an offset batch out of a list of offsets.
    */
-  def createCommittableOffsetBatch(offsets: java.util.List[CommittableOffset]): CommittableOffsetBatch = {
+  def createCommittableOffsetBatch[T <: Committable](offsets: java.util.List[T]): CommittableOffsetBatch = {
```
jewertow: The test I added proves that the type of this list has to be `T <: Committable` istead of `Committable`, but both types `Committable` and `T <: Committable` are binary-incompatible.

tests/src/test/java/docs/javadsl/AtLeastOnceTest.java
```diff
@@ -200,4 +202,44 @@ public void consumerOneProduceConditional() throws Exception {
...
+        Consumer.sourceWithOffsetContext(consumerSettings, Subscriptions.topics(topic1))
+            .grouped(5)
+            .map(
+                records -> {
+                  String key = records.iterator().next().key();
+                  String value =
+                      records.stream().map(ConsumerRecord::value).collect(Collectors.joining(","));
+                  return ProducerMessage.single(new ProducerRecord<>(topic2, key, value));
+                })
+            .mapContext(ConsumerMessage::createCommittableOffsetBatch)
```
jewertow: In case of `def createCommittableOffsetBatch(offsets: java.util.List[Committable])` it wouldn't compile.

core/src/main/scala/akka/kafka/ConsumerMessage.scala
seglo: Agreed. This seems fine to me. It's binary-incompatible, but since it's only widening permitted types it shouldn't break anything that was using the previous method.

tests/src/test/java/docs/javadsl/AtLeastOnceTest.java
seglo: :+1:

PR #1034, opened 2020-02-04
Upgrade to CP 5.4.0

seglo:
Upgrade to Confluent Platform 5.4.0:

* Embedded Kafka
* Embedded Kafka with Schema Registry
* Confluent Platform `cp-kafka` Docker image used by testcontainers tests


seglo:
It looks like a 2.13 version of embedded-kafka-schema-registry wasn't released for some reason.  Will follow up upstream.

seglo:
Confluent Platform dependencies don't support 2.13 yet.

https://github.com/embeddedkafka/embedded-kafka-schema-registry/issues/78

seglo:
Java 11 builds are failing because a transient dependency is bringing in a dep that no longer exists `org.openjfx:javafx.base:11.0.0-SNAPSHOT`.

@eed3si9n identified the [root cause](https://hibernate.atlassian.net/browse/HV-1644).  I'm working around it by adding an exclusion rule for `javafx.base`, but I'll follow up with confluent to let them know about the issue (if they don't know already).

Details: https://stackoverflow.com/a/59872126/895309

seglo:
I logged an issue upstream. https://github.com/confluentinc/rest-utils/issues/170

seglo:
The `javafx.base` artifact is published to the Sonatype public repo (i.e. adding `Resolver.sonatypeRepo("public")` to your list of SBT repo resolvers).  I think it would be better to use the exclude rule in our build so that downstream users of our testkit don't also have this problem.  According to maven docs:

> If you deploy Project-A to a repository, and Project-X declares a normal dependency on Project-A, will Project-D still be excluded from the classpath?
>
> ```Project-X -> Project-A```
>
> The answer is Yes. Project-A has declared that it doesn't need Project-D to run, so it won't be brought in as a transitive dependency of Project-A.

http://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

PR #1035, opened 2020-02-05
Fix logging an error with cause

Sebruck:
Swapping the message and cause in the logger made it using the [error method which takes a template](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/event/Logging.scala#L1261) and therefore tries to parse the cause into the message.

This results in a confusing error message and a loss of information. For example the log in the [DeferredProducer](https://github.com/akka/alpakka-kafka/blob/master/core/src/main/scala/akka/kafka/internal/DeferredProducer.scala#L70) results in 

> producer creation failed WARNING arguments left: 1

By putting the arguments back into the initial order (first cause, then message) it will treat the cause properly as a Throwable again in this function https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/event/Logging.scala#L1211


ennru:
Thank you for fixing this!

Sebruck:
Thanks for reviewing and merging!

Issue #1039, opened 2020-02-05
Drain partial offset commit batches when an upstream failure occurs


seglo:
### Short description

Flush partially batched offset commits when an upstream failure occurs.

### Details

The current `Committer.flow` implementation uses a `groupWeightedWithin` to batch records using criteria specified in `CommitterSettings`. If an upstream failure occurs we would want to commit the inflight batch to the `KafkaConsumerActor` so that it can be committed during the shutdown process.

Assuming the upstream error isn't Kafka connection related and the stream has the time to shutdown we can ensure that as much of the batch as possible is committed to Kafka so that we don't need to replay those messages when the stream starts back up.

See the discussion in #1032 for the motivation behind this feature request.

ennru:
Thank you @sdudzin for bringing up the discussion in #1032.
This sounds like a good idea for a generic solution. I guess this would require a new internal `Commit` message which triggers an immediate `poll`.

sdudzin:
Since there are 2 messages that are sent (`Commit` and `CommitWithoutReply`) would it make sense to add a flag to both (like `immediateCommit` or `flush`) to them to avid having maintain 2 more message types?

Another thing to consider do we want to make this default behaviour or enable behind a feature flag?

I could come up with a draft implementation of the flow part if that's ok.

seglo:
> Since there are 2 messages that are sent (`Commit` and `CommitWithoutReply`) would it make sense to add a flag to both (like `immediateCommit` or `flush`) to them to avid having maintain 2 more message types?

Yes, `flush` sounds good to me.

> 
> Another thing to consider do we want to make this default behaviour or enable behind a feature flag?

I think a feature flag makes sense.  I also think making it the default case makes sense.  The only concern I have is that secondary exceptions from the Kafka Consumer might get swallowed up, but I suppose this is an existing issue since the `KafkaConsumerActor` will attempt to commit on shutdown anyway.

> 
> I guess we could still use `groupWeightedWithin` but not for batch definition but to be able to cutoff failures from proper flow elements..
> 

I'm not sure what you mean.

> I could come up with a draft implementation of the flow part if that's ok.

That would be excellent.  Thanks @sdudzin!

ennru:
For the specific implementation in `Producer.committableSink` this shouldn't be too hard to implement. I'll look into it.

sdudzin:
Well since CommittingProducerSinkStage seems to be doing it's own batching I mostly kept backwards compatibility on api level.

[...]

PR #1058, opened 2020-02-13
Drain partial offset commit batches on upstream failure

sdudzin:
Addresses https://github.com/akka/alpakka-kafka/issues/1039


sdudzin:
The first commit contains mostly the flow part + some refactoring to keep the code compiling.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
```diff
@@ -26,23 +30,60 @@ object Committer {
...
+      }
+      .mapConcat { committables =>
+        val successfulCommittables = committables
+          .takeWhile(_.isSuccess)
+          .collect { case Success(committable) => committable }
+        val failure = committables.collectFirst { case Failure(ex) => Failure(ex) }
+        val hasFailures = failure.isDefined
+
+        Success(FlushableOffsetBatch(CommittableOffsetBatch(successfulCommittables), flush = hasFailures)) :: failure.toList
+      }
```
ennru: This becomes a scenario where I'd recommend a specific stage implementation, similar to what `CommittingProducerSinkStage` does.

seglo: I actually like the use of the operator DSL here.  It seems simple and concise.

Are we swallowing the exception though?

core/src/main/resources/reference.conf
```diff
@@ -166,6 +166,10 @@ akka.kafka.committer {
   # WaitForAck: Expect replies for commits, and backpressure the stream if replies do not arrive.
   # SendAndForget: Send off commits to the internal actor without expecting replies (experimental feature since 1.1)
   delivery = WaitForAck
+
+  # API may change
+  # Flush partial batches on stream cancellation so that it can be committed durign shutdown process
+  flush-partial-batches = true
```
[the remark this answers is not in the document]
sdudzin: I probably misunderstood earlier comment about default setting. Agree with the remarks.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
sdudzin: No it's not supposed to be swallowed. I have not covered this with tests yet though...

sdudzin:
> The most relevant part in the `KafkaConsumerActor` is missing so far.

I had a look at KafkaConsumerActor it looks like it already has some functionality to 'flush'(execute poll). The question is do we want to re-use it or add different message to separate the semantics (and put more code).

seglo:
> I had a look at KafkaConsumerActor it looks like it already has some functionality to 'flush'(execute poll). The question is do we want to re-use it or add different message to separate the semantics (and put more code).

@ennru and I discussed this today. It would be nice to keep things DRY if you can.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
seglo: One concern here is all the extra object instances we're creating by wrapping each record in a `Try`. I don't see a way around this as long as we're using the operator DSL, but a `CustomGraphStage` like @ennru suggested might actually be more efficient.

sdudzin: Indeed there is some overhead in this.

sdudzin: Custom stage stage would indeed be more efficient

sdudzin: @ennru I have noticed that the `CommittingProducerSinkStage` enforces `CommitDelivery.WaitForAck`. It seems we would need the same restriction if we are to implement custom stage for the flow?

ennru: I believe that could be relaxed if the code is made aware of it (eg. it should not keep track of commit acknowledgements).

seglo:
@sdudzin I'm just curious where you're at with this.  We want to implement flushing support in the `ProducerCommittingSinkStage` as well and it would require the same machinery in the `KafkaConsumerActor` to handle flushed commits.  I could work on this and have it merged for use in this PR, but I don't want to duplicate work if you've already started the work in the `KafkaConsumerActor`.

sdudzin:
@seglo I am quite far in implementation of the custom graph stage (`CommittingFlowStage`). I haven't touched `KafkaConsumerActor` much besides what's already in this branch. I think as far as `KafkaConsumerActor` it's relatively safe to edit it.

seglo:
@sdudzin That's great.  In that case I'll park the other work and wait to see what you come up with.  To be clear, feel free to re-factor the batching in `KafkaConsumerActor` to fit the use case of the `CommittingFlowStage`.  Ping me or @ennru when you're ready for a review.

core/src/main/resources/reference.conf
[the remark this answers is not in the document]
sdudzin: fxed

sdudzin:
@seglo , @ennru please have a look. For the last commit no build was triggered for some reason.

seglo:
@sdudzin I think the build check didn't run because of the merge conflict.  Can you rebase on master?

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  override protected def onTimer(timerKey: Any): Unit = timerKey match {
+    case BatchingFlowStage.CommitNow => pushDownStream(Interval, flush = false)(push)
+  }
+
+  private def pushDownStream(triggeredBy: TriggerdBy, flush: Boolean)(
+      emission: (Outlet[Try[FlushableOffsetBatch]], Try[FlushableOffsetBatch]) => Unit
+  ): Unit = {
+    if (activeBatchInProgress) {
+      log.debug("pushDownStream triggered by {} (flush={}, offsetBatch.size={})",
```
seglo: We should do the commit here instead of propagating the batch downstream. If the user wants to handle upstream stream failures they can handle them on their own.

tests/src/test/scala/akka/kafka/scaladsl/CommittingSpec.scala
```diff
@@ -51,7 +51,7 @@ class CommittingSpec extends SpecBase with TestcontainersKafkaLike with Inside {
       val (control, probe1) = Consumer
         .committableSource(consumerSettings, Subscriptions.topics(topic1))
         .mapAsync(10) { elem =>
-          elem.committableOffset.commitInternal().map { _ =>
+          elem.committableOffset.commitInternal(flush = false).map { _ =>
```
seglo: With the default param already `false` you can drop most of these existing test updates.

core/src/main/mima-filters/2.0.1.backwards.excludes
```diff
@@ -0,0 +1,6 @@
+# Internal API
```
seglo: We released 2.0.2 so please update this filename accordingly.

benchmarks/src/main/scala/akka/kafka/benchmarks/ReactiveKafkaConsumerBenchmarks.scala
```diff
@@ -166,7 +166,7 @@ object ReactiveKafkaConsumerBenchmarks extends LazyLogging with InflightMetrics
     val control = fixture.source
       .mapAsync(1) { m =>
         meter.mark()
-        m.committableOffset.commitInternal().map(_ => m)(ExecutionContexts.sameThreadExecutionContext)
+        m.committableOffset.commitInternal(flush = false).map(_ => m)(ExecutionContexts.sameThreadExecutionContext)
```
seglo: Revert for new default param.  There are a bunch of these callsites that can be reverted.

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
```diff
@@ -0,0 +1,143 @@
...
+import akka.kafka.internal.BatchingFlowStage.FlushableOffsetBatch
+import akka.stream._
+import akka.stream.stage._
+
+import scala.util.{Failure, Success, Try}
+
+/**
+ * INTERNAL API.
+ *
+ * Combined stage for producing, batching commits and committing.
```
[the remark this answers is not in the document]
sdudzin: Fixed

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
```diff
@@ -0,0 +1,143 @@
...
+ */
+@InternalApi
+private[kafka] final class BatchingFlowStage(val committerSettings: CommitterSettings)
+    extends GraphStage[FlowShape[Committable, Try[FlushableOffsetBatch]]] {
```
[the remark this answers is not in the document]
sdudzin: Yes , this is a long story. 
On public API level nothing has changed (no Try is exposed) so my expectation/assumption is that it's ok to use `Try`.
I propagate `Try` because otherwise the behaviour was non-deterministic in case of upstream failure. 
Basically sometimes successful batch that comes before the failure also gets 'cancelled'. This triggered random failures in `BatchingFlowStageSpec."batch commit all buffered elements if upstream has suddenly failed"`.  I've identified that this happens due to async step in the `Committer` (replacing asyncMap with map fixes it but we loose parallelism). I have not found a way around this and `Try` seems rather simple solution. Also note that only batch is wrapped in `Try` (not every offset as in the very first version).

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
sdudzin: Actually this was my first version of the GraphStage, unfortunately If we do the commit we loose possibility for parallelism.  And implementing parallelism manually in a `GraphStage` seems like a lot more complex than combining this stage with  simple follow-up steps in the `Committer`. 

Since `Committer` is the public API and not this Graph I assumed this as acceptable solution.

tests/src/test/scala/akka/kafka/scaladsl/CommittingSpec.scala
sdudzin: Removed redundant default arguments (also inother places)

core/src/main/mima-filters/2.0.1.backwards.excludes
sdudzin: fixed

benchmarks/src/main/scala/akka/kafka/benchmarks/ReactiveKafkaConsumerBenchmarks.scala
sdudzin: fixed, I walked through other occurrences as well

sdudzin:
> Looking great 👍. The main thing I noticed is you've chosen to propagate `Try`. I'm not sure this is necessary, but I'm interested in why you think it is.

@seglo please see my comments above. For the context I had 3 implementations:
№1 was implementation that did commits in the graphstage but the parallelism feature was lost.
№2 was flushing batches without commits and `Try`'s but it lead to nondeterministic behaviour in case of upstream failures (see my comments about mapasync)
№3 (current) is a tradeoff between complexity (no async stuff in the graph stage) and overhead (`Try` but only for batches).

seglo:
I see now that the `Try` is not exposed to users and it's only per batch as you mentioned, so that's good.

> №1 was implementation that did commits in the graphstage but the parallelism feature was lost.

Yes, `mapAsync|mapAsyncUnordered` are useful for backpressuring on offset commits under normal operatoring conditions, so it would be ideal to keep that in place.  I agree that reimplementing that property in the stage would not be worth the effort.  

Have you tried committing the batch in the stage only when `onUpstreamFailure` is called and then propagate `FlushableOffsetBatch` (or simply `CommittableOffsetBatch`) otherwise?  In failure scenarios where the batch also reaches `mapAsyncUnordered` before the upstream failure it will only result in the same batch being committed twice, which will get filtered out in the `KafkaConsumerActor` anyway.

sdudzin:
@sebady I just came up with a test case when my implementation is stil non-deterministic :(

> Have you tried committing the batch in the stage only when `onUpstreamFailure` is called and then propagate `FlushableOffsetBatch` (or simply `CommittableOffsetBatch`) otherwise? In failure scenarios where the batch also reaches `mapAsyncUnordered` before the upstream failure it will only result in the same batch being committed twice, which will get filtered out in the `KafkaConsumerActor` anyway.

This is actually very interesting idea. Are we ok with the following?
* Doing double commits (messages to the actor)
* Relying on the `KafkaConsumerActor`'s behaviour that it wont change thus breaking the feature?

Tbh I don't see much other choice...

sdudzin:
I will come up with the implementation and adjusted test case

seglo:
> * Doing double commits (messages to the actor)

I think this is ok in some circumstances. It would be strange to do it all the time though. Kafka will ignore duplicate offset commits as well, so I don't think we would have a problem.  The latest offset commit always wins.

sdudzin:
> > * Doing double commits (messages to the actor)
> 
> I think this is ok in some circumstances. It would be strange to do it all the time though. Kafka will ignore duplicate offset commits as well, so I don't think we would have a problem. The latest offset commit always wins.

I also checked the implementation of the consumer actor it looks safe especially if the commits come one after another (since they are kept in a map).

@seglo I've pushed the new implementation. Can you have a look at it? I've also adjusted a test case that reflects more real-life scenario.

seglo:
Just looked at this again.  Is it even necessary to emit a `Try[CommittableOffsetBatch]` now that the committing on failure is handled by the stage?

sdudzin:
> @sdudzin The tests are very well structured and easy to read, nice work. I ran the suite 500x without issue. You mentioned earlier you experienced some flakiness, has that been resolved now with the commit in the stage on failure?

The last commit (committing batch on failure) resolved the flakiness - commit of partial batch is now guaranteed.

> Can you add some notes about this feature to the [committing](https://doc.akka.io/docs/alpakka-kafka/current/consumer.html#offset-storage-in-kafka-committing) section of the Consumer docs?

Wil do.

> I think at least one integration test would be worthwhile. Can you add a test to `CommittingSpec` that tests flushing commits in a failure scenario and asserts Kafka received the offsets?
Sure, I will let you know when its done.

sdudzin:
> Just looked at this again. Is it even necessary to emit a `Try[CommittableOffsetBatch]` now that the committing on failure is handled by the stage? Could the exception be thrown in the stage?

Well I encountered this decision as well. Basically there are at least 2 possible implementations with their pro's and con's:

1. Current implementation: Commit happens on failure and then the partial batch is forwarded to the `Committer`. 
_Pro's:_ 
Partial batch has a chance of reaching downstream allowing whatever needed processing to take place.
_Con's:_ 
Partial batch reaching downstream is not deterministic due to the nature of the async stage that is used n the `Committer`. 
There is some reliance on `KafkaConsumerActor` being able to deal with double `Commit` messages.
2. Alternative implementation: Commit **still** happens on failure but it also fails the stage. No `Try` is needed.
_Pro's:_ 
No double commit commands are sent to the `KafkaConsumerActor`. Deterministic behaviour - partial batch will never appear in downstream (this has it's negative side, see below).
_Con's:_ 
Although the behaviour is deterministic partial batch will never reach downstream reducing usefulness of this exercise.

I opted for the 1st version because I saw more value in partial batch reaching (albeit non-deterministically) downstream rather than never having a chance of reaching it. I also didn't consider instantiating a `Try` per batch as a significant overhead comparing to all the object creations that happen in the stage and `Committer`.

That being said I do not have strong preference though. @seglo , let me know what you think?

seglo:
> That being said I do not have strong preference though. @seglo , let me know what you think?

Let me consult on this internally.  I'll get back to you by tomorrow.

sdudzin:
> > That being said I do not have strong preference though. @seglo , let me know what you think?
> 
> Let me consult on this internally. I'll get back to you by tomorrow.

@seglo sure. I could try to solve the non-determinism in `option 1` in the `Committer` if replacing `mapAsyncUnordered` with `mapAsync` is acceptable.

seglo:
@sdudzin I think we should proceed with option 2 and not emit the `Try` downstream.  Errors/exceptions in streams do not actually propagate along the same channel as elements.  For instance, they don't respect back-pressure.  So even when using stream stages that enforce ordering there's no guarantees that the elements emitted from upstream before an error occurred will arrive downstream before the error.  If users want to enforce ordering of logical/functional errors in the event stream then they can handle them accordingly, but I think the `BatchingFlowStage` should just flush the batch ASAP when an error is encountered and rethrow an exception.

sdudzin:
@seglo Try is removed, integration test added and documentation changed. Please have a look.

core/src/main/scala/akka/kafka/internal/CommittingProducerSinkStage.scala
```diff
@@ -160,25 +161,25 @@ private final class CommittingProducerSinkStageLogic[K, V, IN <: Envelope[K, V,
...
+  private def commit(triggeredBy: TriggerdBy, flush: Boolean): Unit = {
     if (offsetBatch.batchSize != 0) {
       log.debug("commit triggered by {} (awaitingProduceResult={} awaitingCommitResult={})",
                 triggeredBy,
                 awaitingProduceResult,
                 awaitingCommitResult)
       val batchSize = offsetBatch.batchSize
       offsetBatch
-        .commitInternal()
+        .commitInternal(flush)
```
ennru: Maybe use the trigger instead to be more clear on when it is flushing:
```suggestion
        .commitInternal(flush = TriggeredBy == UpstreamFailure)
```

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -254,16 +254,22 @@ import scala.util.control.NonFatal
   override val receive: Receive = regularReceive
 
   def regularReceive: Receive = LoggingReceive {
-    case Commit(tp, offset) =>
+    case Commit(tp, offset, flush) =>
       // prepending, as later received offsets most likely are higher
       commitMaps = tp -> offset :: commitMaps
       commitSenders = commitSenders :+ sender()
+      if (flush) {
+        requestDelayedPoll()
```
ennru: As this is used for emergencies, we should consider calling `receivePoll` or `poll` immediately. That would improve chances the commit gets across to the broker before the consumer is shut down by stream error propagation.
`emergency` might be more accurate than `flush`.

docs/src/main/paradox/consumer.md
```diff
@@ -147,7 +147,7 @@ Scala
 Java
 : @@ snip [snip](/tests/src/test/java/docs/javadsl/ConsumerExampleTest.java) { #atLeastOnce }
 
-Committing the offset for each message (`withMaxBatch(1)`) as illustrated above is rather slow. It is recommended to batch the commits for better throughput, with the trade-off that more messages may be re-delivered in case of failures.
+Committing the offset for each message (`withMaxBatch(1)`) as illustrated above is rather slow. It is recommended to batch the commits for better throughput, in such cases when upstream fails partial batches are committed thus significantly reducing number of re-deliveries in case of failures.
```
ennru:
```suggestion
Committing the offset for each message (`withMaxBatch(1)`) as illustrated above is rather slow. It is recommended to batch the commits for better throughput, in cases when upstream fails the `Committer` will try to commit the offsets collected before the error.
```

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
```diff
@@ -26,20 +27,20 @@ object Committer {
...
         offsetBatches
-          .mapAsyncUnordered(settings.parallelism) { b =>
-            b.commitInternal().map(_ => b)(ExecutionContexts.sameThreadExecutionContext)
+          .mapAsyncUnordered(settings.parallelism) { batch =>
+            batch.commitInternal().map(_ => batch)(ExecutionContexts.sameThreadExecutionContext)
           }
       case SendAndForget =>
-        offsetBatches
-          .map(_.tellCommit())
+        offsetBatches.map(_.tellCommit())
```
ennru: This is built-in now and should not be needed here anymore.
But the `CommitDelivery` distinction is missing in the stage implementation.

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
```diff
@@ -0,0 +1,151 @@
...
+import akka.stream.stage._
+
+/**
+ * INTERNAL API.
+ *
+ * Combined stage for committing incoming offsets in batches. Capable of emitting dynamic (reduced) size batch in case of
+ * upstream failures. Support flushing on failure (for downstreams).
+ */
+@InternalApi
+private[kafka] final class BatchingFlowStage(val committerSettings: CommitterSettings)
```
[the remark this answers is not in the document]
sdudzin: `CommitCollectorStage` sounds good!

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
```diff
@@ -0,0 +1,151 @@
...
+      failStage(ex)
+    }
+  }
+
+  override def postStop(): Unit = {
+    log.debug("BatchingFlowStage stopped")
+    super.postStop()
+  }
+
+  private def noActiveBatchInProgress: Boolean = offsetBatch.batchSize == 0
```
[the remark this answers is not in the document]
sdudzin: I think this means adding a method to a public API. Is this ok?

core/src/main/scala/akka/kafka/internal/CommittingProducerSinkStage.scala
sdudzin: sounds good

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
sdudzin: > This is built-in now and should not be needed here anymore.

Committing is still used for normal scenario's, stage only commits on failures. Can you elaborate what is not needed?

> But the `CommitDelivery` distinction is missing in the stage implementation.

Indeed, adding!

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
sdudzin: ok, this makes sense, I will change that

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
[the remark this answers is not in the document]
seglo: I think it would be ok as part of the public API.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
seglo: > This is built-in now and should not be needed here anymore.

@ennru We discussed that the normal committing logic could remain as a separate `mapAsyncUnordered` operator step so similar logic wouldn't need to be implemented in the custom graph stage.  WDYT?

core/src/main/scala/akka/kafka/ConsumerMessage.scala
```diff
@@ -60,7 +60,7 @@ object ConsumerMessage {
     def commitJavadsl(): CompletionStage[Done]
 
     @InternalApi
-    private[kafka] def commitInternal(): Future[Done]
+    private[kafka] def commitInternal(flush: Boolean = false): Future[Done]
```
ennru: Instead of a the parameter here, try and add a method `private[kafka] def commitEmergency(): Future[Done]` to `CommittableOffsetBatch`. It makes this less intrusive.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -254,16 +254,22 @@ import scala.util.control.NonFatal
...
-    case CommitWithoutReply(tp, offset) =>
+    case CommitWithoutReply(tp, offset, flush) =>
       // prepending, as later received offsets most likely are higher
       commitMaps = tp -> offset :: commitMaps
+      if (flush) {
+        requestDelayedPoll()
+      }
 
-    case CommitSingle(tp, offset) =>
+    case CommitSingle(tp, offset, _) =>
```
ennru: The `flush` parameter is not needed for `CommitSingle`.

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
ennru: Yes, makes sense. And we do not need the distinction with tell commit I guess. As everything will fail just after anyway, so just using tell commits would be the right thing to do?

sdudzin: What about failed commits? Do we get stream failures with `tellCommit`?

ennru: No, but if the commit is triggered by a stream failure, we shouldn't override that failure anyway. Just try our best to get the offset to be committed over to the Kafka broker, no matter what happens after that.

core/src/main/scala/akka/kafka/internal/BatchingFlowStage.scala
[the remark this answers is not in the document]
sdudzin: added API

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
sdudzin: removed

core/src/main/scala/akka/kafka/ConsumerMessage.scala
sdudzin: I actually like the semantics of `flush` more (also makes the usage compact and descriptive) and with default arguments it's still not as intrusive. 
I've added the method, this required some minor refactoring. Please have a look if this is ok.

sdudzin: Will tellCommit() would also require en `tellCommitEmergency()` method?

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
```diff
@@ -0,0 +1,157 @@
...
+    new OutHandler {
+      def onPull(): Unit = if (!hasBeenPulled(stage.in)) {
+        tryPull(stage.in)
+      }
+    }
+  )
+
+  private def commitAndPushWithFailure(ex: Throwable): Unit = {
+    setKeepGoing(true)
+    if (offsetBatch.batchSize != 0) {
```
ennru:
```suggestion
    if (!offsetBatch.isEmpty) {
```
and possibly turn the else clause around.

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+
+  override def postStop(): Unit = {
+    log.debug("CommitCollectorStage stopped")
+    super.postStop()
+  }
+
+  private def noActiveBatchInProgress: Boolean = offsetBatch.isEmpty
+  private def activeBatchInProgress: Boolean = !noActiveBatchInProgress
```
ennru: I believe using `offsetBatch.isEmpty` directly instead of these methods makes it clearer.

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
```diff
@@ -0,0 +1,157 @@
...
+        case CommitDelivery.WaitForAck =>
+          offsetBatch
+            .commitEmergency()
+            .onComplete { _ =>
+              commitResultOnFailureCallback.invoke(ex)
+            }(materializer.executionContext)
+        case CommitDelivery.SendAndForget =>
+          offsetBatch.tellCommit(flush = true)
+          commitResultOnFailureCallback.invoke(ex)
+      }
```
ennru: Hm, I wonder if it makes sense to do anything but (emergency) tell commit here. Does it really make sense to `keepGoing`?

core/src/main/scala/akka/kafka/ConsumerMessage.scala
sdudzin: @ennru please also see my other remark. Maybe we should just have one `tellCommitEmergency()` (or `commitEmergency()` for simplicity sake)?

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
sdudzin: Indeed, using only tellCommit makes more sense

core/src/main/scala/akka/kafka/scaladsl/Committer.scala
sdudzin: > No, but if the commit is triggered by a stream failure, we shouldn't override that failure anyway. Just try our best to get the offset to be committed over to the Kafka broker, no matter what happens after that.

That sounds logical. 

May I propose to use only `tellCommit` in the `CommitCollectorStage` (thus ignore the `CommitDelivery` setting. And keep it as it is for the regular flow (i.e. respecting the setting in the `Committer`)? 
We might still want to get feedback on commit failures in normal scenario's (i.e. when no upstream failure occurred). With `tellCommit` no stream restart possible in case of kafka commit failures.

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
sdudzin: done

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
sdudzin: fixed

core/src/main/scala/akka/kafka/ConsumerMessage.scala
sdudzin: > @ennru please also see my other remark. Maybe we should just have one `tellCommitEmergency()` (or `commitEmergency()` for simplicity sake)?

On the other hand, `akka.kafka.internal.CommittingProducerSinkStageLogic#commit` uses callbacks, waiting for your feedback on this. For now leaving as it is.

sdudzin:
@seglo , @ennru I've made the following changes:
- renamed `flush` to `emergency` based on feedback
- emergency commit is used only in 'tell' mode
- added specific api call for emergency commits, `commit` and `tellCommit` now have their original signatures, this reduced the number of possible message api calls from 4 to 3.
- use `tellCommitEmergency` in CommittingProducerSinkStage and CommitCollectorStage
- `Commit` message no longer has the 'emergency' flag (only CommitWithoutReply has it)

Please have a look

tests/src/test/scala/akka/kafka/internal/CommitCollectorStageSpec.scala
```diff
@@ -0,0 +1,288 @@
...
+      Consumer.Control,
+      TestSubscriber.Probe[CommittableOffsetBatch],
+      TestOffsetFactory) = {
+    val (source, control, sink) = streamProbes(committerSettings)
+    val factory = TestOffsetFactory(new TestBatchCommitter(committerSettings))
+    (source, control, sink, factory)
+  }
+}
+
+object TestCommittableOffset {
```
ennru: Please move the test helpers into `CommitCollectorStageSpec`.

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
```diff
@@ -0,0 +1,141 @@
...
+    setKeepGoing(true)
+    if (offsetBatch.isEmpty) {
+      failStage(ex)
+    } else {
+      log.debug("committing batch in flight on failure {}", offsetBatch)
+      offsetBatch.tellCommitEmergency()
+      offsetBatch = CommittableOffsetBatch.empty
+      failStage(ex)
+    }
+  }
```
ennru: Move into into `onUpstreamFailure`? The `setKeepGoing` can be removed.

sdudzin: fixed

tests/src/test/scala/akka/kafka/internal/CommitCollectorStageSpec.scala
sdudzin: fixed

core/src/main/mima-filters/2.0.2.backwards.excludes
```diff
@@ -0,0 +1,9 @@
+# Internal API
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.tellCommit")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.tellCommit")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.commitInternal")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.ConsumerMessage#Committable.commitInternal")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#Committable.commitInternal")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.isEmpty")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#Committable.commitEmergency")
+ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.kafka.ConsumerMessage#CommittableOffsetBatch.tellCommitEmergency")
```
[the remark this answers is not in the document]
sdudzin: fixed

sdudzin:
@ennru I've addressed your remarks.

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
```diff
@@ -0,0 +1,138 @@
...
+
+      override def onUpstreamFailure(ex: Throwable): Unit = {
+        log.debug("Received onUpstreamFailure with exception {}", ex)
+        if (noActiveBatchInProgress) {
+          failStage(ex)
+        } else {
+          {
+            if (offsetBatch.isEmpty) {
+              failStage(ex)
+            } else {
```
ennru: This inlining wasn't perfect.

sdudzin: my bad, I've fixed it

ennru: Sorry, I wasn't clear enough: I disliked to duplicated `if` on the same condition. 
I pushed a fix to that now.

ennru:
Thank you @sdudzin!
This was quite a long way from your original need to this solution. Thank you for taking it all the way!

sdudzin:
@ennru thank you for letting me work on this!

seglo:
Great work @sdudzin!

Issue #1062, opened 2020-02-24
Default PartitionAssignmentHandler behaviour on revoke+assing in transactional


szymonm:
### Short description

`SingleSourceLogic` clears the buffer of accumulated records of revoked partitions only after new partitions are assigned. I think that this causes blocking `onRevoke` callback to wait for all the records in the buffer to be drained. While this approach may work in certain cases, it is not universal and I propose to add a configuration option to allow dropping records just after receiving `onRevoke`.

### Details

Looking into the following code from `SingleSourceLogic`:
```
override protected def addToPartitionAssignmentHandler(
      handler: PartitionAssignmentHandler
  ): PartitionAssignmentHandler = {
    val flushMessagesOfRevokedPartitions: PartitionAssignmentHandler = new PartitionAssignmentHandler {
      private var lastRevoked = Set.empty[TopicPartition]

      override def onRevoke(revokedTps: Set[TopicPartition], consumer: RestrictedConsumer): Unit =
        lastRevoked = revokedTps

      override def onAssign(assignedTps: Set[TopicPartition], consumer: RestrictedConsumer): Unit =
        filterRevokedPartitionsCB.invoke(lastRevoked -- assignedTps)


    }
    new PartitionAssignmentHelpers.Chain(handler, flushMessagesOfRevokedPartitions)
  }
```
I can see that we call `filterRevokedPartitionsCB` only after the new partitions are assigned. I would like to understand what are the pros and cons of this approach vs dropping the messages in the `buffer` just after receiving `onRevoke` from the consumer.

While it seems reasonable not to drop records that have already been read from Kafka it leads to the following inefficiency:
1. `onRevoke` is called in the thread that calls `consumer.poll`. After this callback, we start draining the transactional stream. This is done by blocking the `KafkaConsumerActor`.
2.  We don't have control over stream processing, so the stream is working as before, except it will not receive new messages from the `KafkaConsumerActor`.
3. The source is keeping track of record offsets emitted and will unblock the `KafkaConsumerActor` only after all emitted messages are committed. It is possible that we will wait for all messages accumulated in `buffer` to be committed.
4. `onRevoke` callback is blocking all consumers in the consumer group for the given topic. Eventually, all consumers for this consumer group and topic are waiting for the last stream buffer to be drained.

In our setting, where we have hundreds of KTPs, it means that all consumers are blocked waiting for the last one to finish draining the whole buffer. This slows down processing for the whole system.
[...]

PR #1063, opened 2020-02-25
Filtering of messages in transactional source buffer on revoke (#1062)

szymonm:
References #1062

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
+++ b/core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
@@ -163,13 +163,15 @@ private[internal] abstract class TransactionalSourceLogic[K, V, Msg](shape: Sour
       override def onAssign(assignedTps: Set[TopicPartition], consumer: RestrictedConsumer): Unit = ()
 
       // This is invoked in the KafkaConsumerActor thread when doing poll.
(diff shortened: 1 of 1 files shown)


seglo:
To carry on what I mentioned in #1062, I'm going to run our integration tests locally for a little while to be safe.  What I'm doing is doing repeated loops of the `TransactionsSpec` in the tests integration test project.

```
tests/testOnly *.TransactionsSpec -- -DtimesToRepeat=100
```

I'm also going to bump up the number of test elements in the `provide consistency when multiple transactional streams are being restarted` test a few orders of magnitude.  I'll start with 10 million.

seglo:
Coincidentally, the command I mentioned above will also run the `TransactionsSpec` in the normal test scope, which can't hurt!

seglo:
On second thought.  I'll add a temp task to the travis build to do this instead.

szymonm:
Can you run the same configuration on master?

seglo:
> Can you run the same configuration on master?

Yes, I'll use the #1065 to do both.

patriknw:
@szymonm Seems like this has been stale for a long while. Shall we close as obsolete?

szymonm:
Yes, I'm not convinced of this change anymore.

PR #1080, opened 2020-03-17
Cluster-sharding: single listener per entity type key + Java DSL

chbatey:
That way if it is loaded multiple times only one actor is created.

Also remove the method that takes in an actor system. We shouldn't allow
an extension from one system to use another system

Adds a Java API + docs

### DIFF
--- a/cluster-sharding/src/main/scala/akka/kafka/cluster/sharding/KafkaClusterSharding.scala
+++ b/cluster-sharding/src/main/scala/akka/kafka/cluster/sharding/KafkaClusterSharding.scala
@@ -221,6 +221,7 @@ final class KafkaClusterSharding(system: ExtendedActorSystem) extends Extension 
    * val listenerClassicActorRef: akka.actor.ActorRef = listenerTypedActorRef.toClassic
    * }}}
    */
+  @ApiMayChange(issue = "https://github.com/akka/alpakka-kafka/issues/1074")
   def rebalanceListener(typeKey: EntityTypeKey[_]): akka.actor.typed.ActorRef[ConsumerRebalanceEvent] = {
     rebalanceListeners.computeIfAbsent(typeKey, _ => {
       system.toTyped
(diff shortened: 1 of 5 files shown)


cluster-sharding/src/main/scala/akka/kafka/cluster/sharding/KafkaClusterSharding.scala
```diff
@@ -143,11 +221,16 @@ final class KafkaClusterSharding(system: ExtendedActorSystem) extends Extension
    * val listenerClassicActorRef: akka.actor.ActorRef = listenerTypedActorRef.toClassic
    * }}}
    */
-  @ApiMayChange(issue = "https://github.com/akka/alpakka-kafka/issues/1074")
-  def rebalanceListener(typeKey: EntityTypeKey[_]): akka.actor.typed.ActorRef[ConsumerRebalanceEvent] =
-    rebalanceListener(system.toTyped, typeKey)
+  def rebalanceListener(typeKey: EntityTypeKey[_]): akka.actor.typed.ActorRef[ConsumerRebalanceEvent] = {
```
patriknw: lost an `@ApiMayChange`?

chbatey: Doh. Adding

cluster-sharding/src/main/scala/akka/kafka/cluster/sharding/KafkaClusterSharding.scala
```diff
@@ -5,12 +5,13 @@
...
 
+import java.util.concurrent.{CompletionStage, ConcurrentHashMap}
 import java.util.concurrent.atomic.AtomicInteger
 
-import akka.actor.typed.{ActorRef, ActorSystem, Behavior}
+import akka.actor.typed.Behavior
 import akka.actor.typed.scaladsl.Behaviors
 import akka.actor.typed.scaladsl.adapter._
-import akka.actor.{ExtendedActorSystem, Extension, ExtensionId}
+import akka.actor.{ActorSystem, ClassicActorSystemProvider, ExtendedActorSystem, Extension, ExtensionId}
```
ennru: Use the FCQN for `ActorSystem` instead of importing it to make it more clear.

ennru:
Travis wants you to `javafmtAll`.

chbatey:
Then I shall do what travis wants

Issue #1055, opened 2020-02-12
[Failure] IntegrationSpec, Kafka connector must not produce any records after send-failure if stage is stopped


seglo:
https://travis-ci.org/akka/alpakka-kafka/jobs/649448357#L871

```
[info] - must not produce any records after send-failure if stage is stopped *** FAILED *** (1 second, 11 milliseconds)
[info]   java.lang.AssertionError: assertion failed: received unexpected message RealMessage(OnNext(3),Actor[akka://Spec/deadLetters]) after 949 millis
[info]   at akka.testkit.TestKitBase.expectNoMsg_internal(TestKit.scala:726)
[info]   at akka.testkit.TestKitBase.expectNoMessage(TestKit.scala:690)
[info]   at akka.testkit.TestKitBase.expectNoMessage$(TestKit.scala:689)
[info]   at akka.testkit.TestKit.expectNoMessage(TestKit.scala:928)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNoMessage(StreamTestKit.scala:644)
[info]   at akka.kafka.scaladsl.IntegrationSpec.$anonfun$new$20(IntegrationSpec.scala:200)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
[info]   at akka.kafka.scaladsl.IntegrationSpec.$anonfun$new$19(IntegrationSpec.scala:187)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$3.apply(WordSpecLike.scala:1075)
[info]   at akka.kafka.tests.scaladsl.LogCapturing.withFixture(LogCapturing.scala:57)
[info]   at akka.kafka.tests.scaladsl.LogCapturing.withFixture$(LogCapturing.scala:55)
[info]   at akka.kafka.scaladsl.SpecBase.withFixture(SpecBase.scala:15)
[...]

PR #1083, opened 2020-03-24
Remove redundant test "connector must not produce any records after send-failure if stage is stopped"

seglo:
Update test `Kafka connector must not produce any records after send-failure if stage is stopped` in `IntegrationSpec` to reliably fail after a certain number of messages, and ensure that no messages are emitted downstream in the producer stream after a failure occurs.  Assert that messages produced before failure were sent to broker.  I've run this test several 100 times locally successfully.

See this comment for more details on troubleshooting cause https://github.com/akka/alpakka-kafka/issues/1055#issuecomment-602899525

References: #1055

### DIFF
--- a/tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
+++ b/tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
@@ -183,53 +183,6 @@ class IntegrationSpec extends SpecBase with TestcontainersKafkaLike with Inside 
       }
     }
 
-    "not produce or emit records downstream after send-failure and producer stage is stopped" in {
-      assertAllStagesStopped {
-        val numMessagesBeforeFailure = 5
-        final class FailingStringSerializer extends org.apache.kafka.common.serialization.StringSerializer {
-          override def serialize(topic: String, data: String): Array[Byte] = {
-            if (data.toInt > numMessagesBeforeFailure)
-              throw new org.apache.kafka.common.errors.SerializationException()
-            else
-              super.serialize(topic, data)
-          }
-        }
-
-        val topic1 = createTopic(1)
-        val group1 = createGroupId(1)
-
-        val failFirstMessagesProducerSettings =
-          ProducerSettings(system,
-                           new org.apache.kafka.common.serialization.StringSerializer,
-                           new FailingStringSerializer)
-            .withBootstrapServers(bootstrapServers)
-            .withProperty(ProducerConfig.BATCH_SIZE_CONFIG, numMessagesBeforeFailure.toString)
-            .withProperty(ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION, 1.toString)
-            .withProperty(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, "true")
-
-        val producerProbe = Source(1 to 100)
-          .map(n => ProducerMessage.single(new ProducerRecord(topic1, partition0, DefaultKey, n.toString)))
-          .via(Producer.flexiFlow(failFirstMessagesProducerSettings))
-          .toMat(TestSink.probe)(Keep.right)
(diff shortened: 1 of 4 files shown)


tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -183,25 +183,48 @@ class IntegrationSpec extends SpecBase with TestcontainersKafkaLike with Inside
...
+          .via(Producer.flexiFlow(failFirstMessagesProducerSettings))
+          .toMat(TestSink.probe)(Keep.right)
+          .run()
+        producerProbe.request(100)
+        // expect failure to kill stage before any successfully produced messages have callbacks executed
+        // this runs locally after 100x test runs, but it may be possible for a callback to emit an element before
+        // the stage shutdown.  in my tests the callbacks for the first `numMessagesBeforeFailure` messages were executed
+        // about 100ms after the stage shutdown
+        producerProbe.expectError() shouldBe a[org.apache.kafka.common.errors.SerializationException]
+        producerProbe.expectNoMessage(1.second)
```
ennru: How could there be a message after the error?

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -183,25 +183,48 @@ class IntegrationSpec extends SpecBase with TestcontainersKafkaLike with Inside
...
-        val failFirstMessagesProducerSettings = producerDefaults.withProperty(ProducerConfig.MAX_BLOCK_MS_CONFIG, "1")
-
-        val producer = produce(topic1, 1 to 100, failFirstMessagesProducerSettings)
-        // assure the producer fails as expected
-        producer.failed.futureValue shouldBe a[org.apache.kafka.common.errors.TimeoutException]
-
-        val (control, probe) = createProbe(consumerDefaults.withGroupId(group1), topic1)
 
-        probe
+        val failFirstMessagesProducerSettings =
```
raboof: This would fail the 6th message, not the first message, right?

seglo:
> I'm not sure I understand what we're testing here.
> 
> The scenario seems to be that if one message is successfully sent to kafka, but the next message fails, the stage fails without passing the 'result' corresponding to the first message downstream.
> 
> This is probably usually true when the messages arrive together, but when there is a delay between the first and the second message you'd expect the result from the first message to be passed downstream before the next message fails, right?

Yes, that would be reasonable to expect.  At some point the original must have failed for all produced values.  No messages would ever actually be sent to Kafka because the metadata could not be retrieved in time (that's my theory anyway).  The test asserted that the error occurred on the first message and the stage failed and no subsequent messages were produced.

The test now is failing occasionally. The `Producer.send` method never blocks so we keep sending messages, and eventually the metadata is cached and messages actually get sent to the broker before the stage shuts down.

> 
> Do we ever rely on the fact that the result corresponding to the first message doesn't make it downstream? Or am I misunderstanding the scenario entirely?

I don't think so.  The Kafka Producer is supposed to maintain order of callbacks, so even if later messages succeed to be sent to the broker, the previous failing message callbacks should be called and shutdown the stream before successful callbacks emit records downstream.

The best this test does is demonstrate some properties of the Kafka Producer and Akka streams.

1. That messages are still produced to Kafka even if we never receive a callback.
2. No messages are emitted downstream after the streams stage failure.

@ennru, @raboof Both of these things are by design, so there's not a lot of value.  I'd be happy to simply remove this test.

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
seglo: Yes, this was a leftover from the previous iteration of the test.  I'll update it if we decide to keep it.

ennru:
We can drop this test.

PR #1084, opened 2020-03-25
DrainingControl: creation in toMat

ennru:
All examples create the `DrainingControl` from the tuple of `Control` and the stream completion, this introduces methods to create it immediately in `toMat` with a new combinator.

### DIFF
--- a/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
+++ b/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
@@ -143,7 +143,7 @@ object Consumer {
      *
      * For use in the `toMat` combination of materialized values.
      */
-    def form[T]: (Control, Future[T]) => DrainingControl[T] = new DrainingControl[T](_, _)
+    def apply[T]: (Control, Future[T]) => DrainingControl[T] = new DrainingControl[T](_, _)
   }
 
   /**
--- a/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
+++ b/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
@@ -11,7 +11,7 @@ import akka.dispatch.ExecutionContexts
 import akka.kafka.ConsumerMessage.{CommittableMessage, CommittableOffset}
 import akka.kafka._
 import akka.kafka.internal._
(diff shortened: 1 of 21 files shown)


core/src/main/scala/akka/kafka/scaladsl/Consumer.scala
```diff
@@ -128,11 +128,22 @@ object Consumer {
...
     def apply[T](tuple: (Control, Future[T])) = new DrainingControl[T](tuple._1, tuple._2)
+
+    /**
+     * Combine the consumer control and a stream completion signal materialized values into
+     * one, so that the stream can be stopped in a controlled way without losing
+     * commits.
+     *
+     * For use in the `toMat` combination of materialized values.
+     */
+    def form[T]: (Control, Future[T]) => DrainingControl[T] = new DrainingControl[T](_, _)
```
seglo, line 146: We could just overload the constructor for the scaladsl and the factory method for javadsl, instead of coming up with new names.  It seems to work locally.

ennru: You're right, I tried different ways and the compiler was unhappy, but now it works.

ennru:
We can spare deprecating until 2.1.0.

docs/src/main/paradox/consumer.md
```diff
@@ -313,7 +313,7 @@ When you are using offset storage in Kafka, the shutdown process involves severa
 ### Draining control
 
 To manage this shutdown process, use the @apidoc[Consumer.DrainingControl]
-by combining the `Consumer.Control` with the sink's materialized completion future in `mapMaterializedValue`. That control offers the method `drainAndShutdown` which implements the process descibed above.
+by combining the @apidoc[Consumer.Control] with the sink's materialized completion future in `toMat` with @scala[`DrainingControl.form`]@java[`Consumer::formDrainingControl`], or from a tuple in `mapMaterializedValue` with @scala[`DrainingControl.apply`]@java[`Consumer::createDrainingControl`]. That control offers the method `drainAndShutdown` which implements the process described above.
```
seglo, line 316: This should be rephrased to account for the overload.

Issue #500, opened 2018-06-12
Publish API for single messages


patriknw:
Sometimes you don't need a stream, but just want to publish to Kafka. We could provide something like this:

```
import scala.concurrent.Future
import scala.concurrent.Promise

import akka.Done
import akka.kafka.ProducerMessage.Message
import org.apache.kafka.clients.producer.Callback
import org.apache.kafka.clients.producer.Producer
import org.apache.kafka.clients.producer.RecordMetadata

class PublishOneExample[K, V, P](producer: Producer[K, V]) {

  def publish(msg: Message[K, V, P]): Future[Done] = {
    val result = Promise[Done]
    producer.send(msg.record, new Callback {
      override def onCompletion(metadata: RecordMetadata, exception: Exception): Unit = {
        if (exception == null)
          result.success(Done)
        else
          result.failure(exception)
      }
    })
    result.future
  }

}
```

Reeebuuk:
And vice versa, we want to consume only one message and if message isn't there in certain timeout end the stream. 

To specify from which offset to start from and how many messages to read. 

I that's possible now please shout :)

WayneWang12:
Any update for this issue?

ennru:
Implemented with #1085

PR #1085, opened 2020-03-25
Non-streaming producer wrapper

ennru:
This adds a utility class for producing to Kafka in non-Akka Streams scenarios with a `Future`/`CompletionStage`-based API.

It uses the `KafkaProducer` construction in `ProducerSettings` so it is compatible with Akka Discovery and wrapped producers.

It offers methods for Kafka `ProducerRecord`, Alpakka `ProducerMessage.Message` and Alpakka `ProducerMessage.Envelope`.

I'm not so happy with the name `ElementProducer`. Suggestions?

References #500


core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
```diff
@@ -0,0 +1,29 @@
...
+import akka.kafka.ProducerMessage.Message
+import akka.kafka.ProducerSettings
+import org.apache.kafka.clients.producer.{Callback, RecordMetadata}
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+
+class ElementProducer[K, V](val settings: ProducerSettings[K, V])(implicit executionContext: ExecutionContext) {
+  val producerFuture = settings.createKafkaProducerAsync()(executionContext)
+
+  def producer[PT](msg: Message[K, V, PT]): Future[PT] = {
```
patriknw: s/producer/send/

ennru:
This is now ready for a review.

core/src/main/scala/akka/kafka/javadsl/ElementProducer.scala
```diff
@@ -0,0 +1,72 @@
...
+   * @param settings producer settings used to create or access the [[org.apache.kafka.clients.producer.Producer]]
+   */
+  def this(settings: ProducerSettings[K, V], ec: Executor) =
+    this(scaladsl.ElementProducer(settings)(ExecutionContext.fromExecutor(ec)))
+
+  /**
+   * Utility class for producing to Kafka without using Akka Streams.
+   * @param settings producer settings used to create or access the [[org.apache.kafka.clients.producer.Producer]]
+   */
+  def this(settings: ProducerSettings[K, V], ec: ExecutionContext) =
```
patriknw: a dispatcher is:
```
trait ExecutionContextExecutor extends ExecutionContext with Executor
```

which of the constructors will be choosen for that? e.g. `system.getDispatcher()`

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * Close the underlying producer (depending on the "close producer on stop" setting).
+   */
+  def close(): Unit = {
+    if (settings.closeProducerOnStop) producerFuture.foreach { producer =>
+      // we do not have to check if producer was already closed in send-callback as `flush()` and `close()` are effectively no-ops in this case
+      producer.flush()
+      producer.close(settings.closeTimeout.asJava)
```
patriknw: is that blocking? if so, and there is no non-blocking alternative we should have a warning in docs

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
```diff
@@ -0,0 +1,108 @@
...
+      }
+    )
+    result.future
+  }
+
+  /**
+   * Close the underlying producer (depending on the "close producer on stop" setting).
+   */
+  def close(): Unit = {
+    if (settings.closeProducerOnStop) producerFuture.foreach { producer =>
```
patriknw: does `settings.createKafkaProducerAsync()` always create a new Producer? if that is the case, shouldn't that producer always be closed here?

patriknw: any answer here?

core/src/main/scala/akka/kafka/javadsl/ElementProducer.scala
ennru: Thanks, that did not work as javac couldn't decide. I'll remove the `ExecutionContext` variant.

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
ennru: `createKafkaProducerAsync` normally creates a new Producer, but the Future evaluates only once as it is a `val`.

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
ennru: Yes, it is blocking (nowadays at least with a timeout) and there is no alternative in the Kafka API.

ennru: I added an async alternative which wraps it in a future.

ennru:
Would `AsyncProducer` be an acceptable name for this?

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
```diff
@@ -0,0 +1,107 @@
...
+   */
+  override def close(): Unit = {
+    if (settings.closeProducerOnStop) producerFuture.foreach(closeInternal)
+  }
+
+  /**
+   * Close the underlying producer (depending on the "close producer on stop" setting).
+   * The future completes once closed.
+   */
+  def closeAsync(): Future[Done] = Future(close()).map(_ => Done)
```
patriknw, line 93: I think we should only provide `close(): Future[Done]`. I understand that means that it can't be AutoClosable, but so be it then.

Important that this is Future is on the dedicated dispatcher from ProducerSettings. I don't trust users to pass in right ExecutionContext, see other comment about that.

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.kafka.clients.producer.{Callback, Producer, ProducerRecord, RecordMetadata}
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import akka.util.JavaDurationConverters._
+
+/**
+ * Utility class for producing to Kafka without using Akka Streams.
+ * @param settings producer settings used to create or access the [[org.apache.kafka.clients.producer.Producer]]
+ */
+final class ElementProducer[K, V] private (val settings: ProducerSettings[K, V])(implicit ec: ExecutionContext)
```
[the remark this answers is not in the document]
ennru: Ok, makes sense. That requires access to the actor system instead, but alright.

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
ennru: Ok, now I understand your concern. If users want to create the producer themselves, they set it via the `ProducerSettings.withProducer` which flips the `closeProducerOnStop` to false and returns it from `createKafkaProducerAsync`.

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
ennru: I removed the non-async `close` now and rely on the execution context from the producer dispatcher instead.

core/src/main/scala/akka/kafka/scaladsl/ElementProducer.scala
patriknw: ah, then it makes sense

seglo:
Only thing I can think of is sticking it in a separate package, maybe along with the consumer bits in #1090 and calling them the Producer and Consumer.

ennru:
> sticking it in a separate package and calling them the Producer and Consumer.

There are too many of those classes already.

I tried the name `SendProducer` now. WDYT?

patriknw:
SendProducer works for me

PR #1090, opened 2020-04-02
[WIP] Non-streaming consumer helpers

ennru:
Helpers for the simple cases that users can 
* consume elements with a `Future`
* convert incoming elements for producing in a `Future`

### DIFF
--- a/core/src/main/scala/akka/kafka/scaladsl/Processor.scala
+++ b/core/src/main/scala/akka/kafka/scaladsl/Processor.scala
@@ -5,11 +5,11 @@
 
 package akka.kafka.scaladsl
 
+import akka.NotUsed
 import akka.dispatch.ExecutionContexts
 import akka.kafka.ProducerMessage.Envelope
 import akka.kafka._
 import akka.stream.scaladsl.{Keep, Source}
-import akka.{Done, NotUsed}
 import org.apache.kafka.clients.consumer.ConsumerRecord
 
 import scala.concurrent.Future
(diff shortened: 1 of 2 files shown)


core/src/main/scala/akka/kafka/scaladsl/Processor.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.stream.Materializer
+import akka.stream.scaladsl.Keep
+import akka.{Done, NotUsed}
+import org.apache.kafka.clients.consumer.ConsumerRecord
+
+import scala.concurrent.Future
+
+object Processor {
+
+  def atLeastOnce[K, V](
```
renatocaval, line 15: What about adding this method to `Consumer` instead? 

I find it odd that we have `Consumer.atMostOnce`, but the `atLeastOnce` is on another object.

core/src/main/scala/akka/kafka/scaladsl/Processor.scala
```diff
@@ -0,0 +1,44 @@
...
+      consumerSettings: ConsumerSettings[K, V],
+      subscription: Subscription,
+      committerSettings: CommitterSettings
+  )(process: ConsumerRecord[K, V] => Future[Done])(implicit materializer: Materializer): DrainingControl[Done] = {
+    Consumer
+      .sourceWithOffsetContext(consumerSettings, subscription)
+      .mapAsync(1)(process)
+      .toMat(Committer.sinkWithOffsetContext(committerSettings))(Keep.both)
+      .mapMaterializedValue(DrainingControl.apply)
+      .run()
```
renatocaval, line 25: And if we return a `Source` instead of running it?

I find it less flexible to immediately run it. In Akka Projections, we may want to wrap the  Source on a `RestartSource` for instance.

renatocaval: oh, I understand it now. It's called `Processor` because it's effectively running it.

core/src/main/scala/akka/kafka/scaladsl/Processor.scala
ennru: `Consumer` is very overloaded already. But it might end up there anyway.

core/src/main/scala/akka/kafka/scaladsl/Processor.scala
ennru: I'm not happy with the naming. I leave that to when we know what we need.

I made it a sink and `run` as you said you just needed the control in https://github.com/akka/akka-projection/pull/15/files#r401689690

core/src/main/scala/akka/kafka/scaladsl/Processor.scala
```diff
@@ -0,0 +1,88 @@
...
+   * not be processed in parallel.
+   */
+  def atLeastOncePerPartitionSource[K, V](
+      consumerSettings: ConsumerSettings[K, V],
+      subscription: Subscription,
+      maxPartitions: Int,
+      committerSettings: CommitterSettings
+  )(
+      process: ConsumerRecord[K, V] => Future[_]
+  ): Source[ConsumerMessage.CommittableOffsetBatch, Consumer.Control] = {
```
ennru, line 54: I experimented a bit with parallelising the user-provided function. Does that make sense to anybody?

seglo:
I think that the patterns in this PR are useful as ideas for supporting at least once committing semantics using Kafka managed offsets in projections.  However I think the implementation should be in projections itself and not Alpakka Kafka.  This PR implements a more simplified form of projections, but it has evolved considerably since then and I'm not sure it has enough value as a standalone API in Alpakka Kafka.  

I recommend that we close this for now and use it as reference in the next phase of projections Kafka integration.

PR #1096, opened 2020-04-10
Testkit: create ProducerMessage.PassThroughResult

sebarys:
During writing test for solution described here https://discuss.lightbend.com/t/kafka-ability-to-react-on-message-publication-result/6208 I noticed that it is not possible in tests provide instance of ProducerMessage.PassThroughResult that is required to fully cover my implementation.

Thanks for reviewing this tiny PR and merging!


ennru:
Thank you!

PR #1099, opened 2020-04-19
Testkit: Generify test producer consumer

claudio-scandura:
Extends `KafkaTestKit` to cater for non string records serdes. This is useful when testing Kafka integrating code that relies on other types of serialisation (i.e. Avro, Protobuf).

The PR also adds methods to produce messages to Kafka in `BaseKafkaTest`.

This change does not break any existing API.


ennru:
Thank you for more improvements to the testkit!

PR #1100, opened 2020-04-20
Synchronously seek to external offset during rebalance of manual offset sources

### LINKED ISSUE #1086: committablePartitionedManualOffsetSource does not use getOffsetsOnAssign on rebalance
### Versions used 

```scala
private val akkaVersion          = "2.6.4"
private val alpakkaKafkaVersion  = "2.0.2"
```

### Expected Behavior

We do manual offset management where we start from the last snapshot offset and `replay` without side-effects to the committed offset, then we enable side-effects.
We expect the stream to have a monotonic progress where offsets increment one by one. In case the offset goes back we ignore that. If the offset takes a step of > 1 we throw an error as this indicates we are skipping messages.

phases:

1. load snapshot with it's offset x(getOffsetsOnAssign)
2. load last committed offset y (getOffsetsOnAssign)
3. alpakka-kafka seeks to offset x and starts the stream
4. we process messages without side-effects until we reach y
5. we process messages with side-effects enabled

### Actual Behavior

On a rebalance during the replay phase (4) and the partition is revoked/reassigned we see that alpakka-kafka keeps the stream open and seeks to the last committed offset (y) which skips a lot of messages and messes up the replay.

This is a common case on redeploy where we have some successive rebalances as nodes are replaced.

### Relevant logs

> [Consumer clientId=xxx-3aaf87e5-707e-4a78-8d1e-b67280ae98b1@xxx-869655c6-97pq2, groupId=mygroup] Setting offset for partition mypartition-2 to the committed offset FetchPosition{offset=10675950, offsetEpoch=Optional.empty, currentLeader=LeaderAndEpoch{leader=zzz.europe-west1.gcp.confluent.cloud:9092 (id: 11 rack: 2), epoch=130}}

### How to solve this?

Does the above behavior make sense for manual offset handling?
Is this configurable?

Without optimizations we could just go for the simple solution of not keeping the stream open on quick revoke/assign.

Any other suggestions?

seglo:
The `Consumer.committablePartitionedManualOffsetSource` allows the user to control offsets to consume from (by retrieving offsets that have been stored externally) as well as commit offsets to Kafka. There are several use cases where this might be useful including:

* As a backup to externally stored offsets
* To enable Kafka ecosystem software that relies on Kafka committed offsets

When a rebalance occurs and the same partition is assigned back to an existing consumer group member it's possible that the Kafka Consumer will internally seek to the last Kafka committed offset after we seek to our last externally committed offset.

This PR uses the blocking partition rebalance handler to seek to external offsets returned by the `getOffsetsOnAssign` user-defined function during the rebalance process when a manual offset source used.  A new timeout `ConsumerSetting` `get-offsets-on-assign-timeout` is used during the execution of the `getOffsetsOnAssign` UDF.

Fixes #1086


seglo:
Thanks @ennru. This works and could potentially solve other issues, but it doesn't address the problem reported in #1086.  I'd like to explore the issue more and put this on the shelf for now.

seglo:
User reported that this did not resolved their issue.  I'm closing for now.

Issue #1012, opened 2019-12-19
Drop Scala 2.11 support


ennru:
### Short description

When is a good time for Alpakka Kafka to go on and stop supporting Scala 2.11?

### Details

The last release of Scala 2.11 was 2017-11-09.

Many Alpakka connectors support Scala 2.11, 2.12 and 2.13. New connectors are built for Scala 2.12 and 2.13 only.
Akka 2.6 is not available for Scala 2.11.
Apache Kafka 2.5 will drop Scala 2.11.

https://github.com/akka/alpakka/issues/2055

ennru:
Now removed with #1102

PR #1102, opened 2020-04-22
Kafka 2.6.0 client upgrade and drop Scala 2.11 support

### LINKED ISSUE #1098: Upgrade to Kafka 2.6.0 client
Now that [Kafka 2.5.0 is released](https://kafka.apache.org/downloads#2.5.0) Alpakka Kafka should check if we there is anything in particular to adapt to.

Kafka dropped Scala 2.11 which doesn't affect the client library, but the Embedded Kafka will not be able to work with Scala 2.11 anymore.

See https://github.com/akka/alpakka-kafka/issues/1012

seglo:
Fixes #1098

Upgrade to Apache Kafka 2.6.0.

* Kafka 2.5.0 drops support for Scala 2.11.  We'll stop publishing 2.11 builds of Alpakka Kafka as well to avoid the conflict of Scala 2.11 issues using Embedded Kafka with the testkit. #1012.
* Drop remaining embedded kafka tests, except those required for documentation.


ennru:
Let's update this for Kafka 2.6.0.

seglo:
There are two remaining embedded-kafka tests that are consistently failing for some reason in the 2.13 build.  The broker of is logging the following exception:

```
[info] 08:54:28.624 ERROR [ka-request-handler-6] kafka.server.KafkaApis                [KafkaApi-0] Error when handling request: clientId=adminclient-1, correlationId=101, api=METADATA, version=9, body={topics=[],allow_auto_topic_creation=true,include_cluster_authorized_operations=false,include_topic_authorized_operations=false,_tagged_fields={}}
[info] java.lang.NoSuchMethodError: 'org.apache.kafka.common.requests.MetadataResponse org.apache.kafka.common.requests.MetadataResponse.prepareResponse(int, java.util.Collection, java.lang.String, int, java.util.List, int)'
[info]  at kafka.server.KafkaApis.handleTopicMetadataRequest(KafkaApis.scala:1177)
[info]  at kafka.server.KafkaApis.handle(KafkaApis.scala:131)
[info]  at kafka.server.KafkaRequestHandler.run(KafkaRequestHandler.scala:70)
[info]  at java.base/java.lang.Thread.run(Thread.java:834)
```

It's possible this is due to some version mismatch of embedded-kafka and Kafka client libraries, but it's unclear why it's only an issue with Scala 2.13.  At the time of this comment there is not 2.6.0 version of embedded-kafka.

Options:

1) Wait until embedded-kafka 2.6.0 is released
2) Build our own embedded-kafka 2.6.0
3) Don't run the remaining embedded-kafka tests (at least for Scala 2.13) since we're going to replace them before the next release anyway.

project/project-info.conf
```diff
@@ -1,6 +1,6 @@
 project-info {
   version: "current"
-  scala-versions: ["2.11", "2.12"]
+  scala-versions: ["2.12", "2.13"]
```
ennru, line 3: I believe we can leave this out altogether, it should be picked up from cross scala versions.

seglo:
embedded-kafka 2.6.0 has been released.

docs/src/main/paradox/testing.md
```diff
@@ -89,7 +89,7 @@ Add the Embedded Kafka to your test dependencies:
 
 As Kafka uses Scala internally, only the Scala versions supported by Kafka can be used together with Embedded Kafka. To be independent of Kafka's supported Scala versions, run @ref:[Kafka in a Docker container](testing-testcontainers.md).
 
-The helpers for running Embedded Kafka are available for **Scala 2.11 and 2.12**.
+The helpers for running Embedded Kafka are available for **Scala 2.12**.
```
ennru, line 92: And Scala 2.13?

project/project-info.conf
seglo: Leave out the whole `scala-versions` property?

docs/src/main/paradox/testing.md
seglo: I'll update it, but we're going to rearrange this whole section when (if?) we drop embedded-kafka.

project/project-info.conf
seglo: Seems to be when it's dropped :+1:

ennru:
Please rebase and create an initial version of `release-notes/2.1.x.md`.

ennru (approved):
LGTM.

Issue #1108, opened 2020-05-11
Failed: TransactionsFailureRetryExample timeout on complete


seglo:
```
[info] TransactionsFailureRetryExample
[info] - should work *** FAILED *** (25 seconds, 70 milliseconds)
[info]   A timeout occurred waiting for a future to complete. Queried 101 times, sleeping 150 milliseconds between each query. (TransactionsExample.scala:127)
[info]   org.scalatest.concurrent.Futures$FutureConcept$$anon$1:
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.tryTryAgain$1(Futures.scala:538)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValueImpl(Futures.scala:550)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValueImpl(ScalaFutures.scala:275)
[info]   at org.scalatest.concurrent.Futures$FutureConcept$class.futureValue(Futures.scala:476)
[info]   at org.scalatest.concurrent.ScalaFutures$$anon$1.futureValue(ScalaFutures.scala:275)
[info]   at docs.scaladsl.TransactionsExample$$anonfun$5$$anonfun$apply$4.apply(TransactionsExample.scala:127)
[info]   at docs.scaladsl.TransactionsExample$$anonfun$5$$anonfun$apply$4.apply(TransactionsExample.scala:88)
[info]   at akka.stream.testkit.scaladsl.StreamTestKit$.assertAllStagesStopped(StreamTestKit.scala:32)
[info]   at docs.scaladsl.TransactionsExample$$anonfun$5.apply(TransactionsExample.scala:88)
[info]   at docs.scaladsl.TransactionsExample$$anonfun$5.apply(TransactionsExample.scala:88)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.FlatSpecLike$$anon$1.apply(FlatSpecLike.scala:1682)
[info]   at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
[info]   at docs.scaladsl.DocsSpecBase.withFixture(DocsSpecBase.scala:15)
[...]

PR #1109, opened 2020-05-11
Promise based transaction draining logic

seglo:
The draining control for Transactional sources will occasionally get into a situation where the consumer source can no longer be signalled by the producer that a transaction is aborted, in order to flush in flight messages used to drain messages on shutdown.

Should fix https://github.com/akka/alpakka-kafka/issues/1108

Other improvements:

* Add first observed offset to transactional batch for troubleshooting with transaction commit/abort logging
* Prettyprint transactional batch in logging for transaction commit and abort
* Cancel commit timer when `onCompletionFailure` is fired (not sure if this is necessary since we don't `setKeepGoing(true)`, but it's also done on `onCompletionSuccess` and it can't hurt.
* ~~Add `onFirstMessageReceived` callback.  If a transactional source is shutdown while messages are inflight, but the first message hasn't been produced, then we don't need to drain inflight messages.~~
  * This doesn't work.  In cases messages don't reach the producer before the consumer shuts down (due to failure or a stream completion signal) then in-flight messages will eventually be produced and committed by `TransactionalProducerStage` anyway.
* Add maximum restart condition to integration tests to avoid transient live lock scenarios where we continuously abort transactions and can't finish test.


seglo:
I haven't been able to reproduce this flakiness on master.

seglo:
Nevermind, I just did, though it seems to happen less frequently.  The symptom is that the transactional source gets stuck in a loop checking for inflight messages to be committed or aborted, but neither ever occurs because the messages from the producer are never received.

ennru:
This doesn't really make it simpler, does it?
But you seem to be on to something here...

seglo:
Looks like I broke exactly once consistency checks somehow.  I'll take a look at some point today, but this might not be ready for 2.0.3.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -38,31 +40,40 @@ private[kafka] final class TransactionalProducerStage[K, V, P](
...
     // There is no guarantee that offsets adding callbacks will be called in any particular order.
     // Decreasing an offset stored for the KTP would mean possible data duplication.
     // Since `awaitingConfirmation` counter guarantees that all writes finished, we can safely assume
     // that all all data up to maximal offsets has been wrote to Kafka.
     private val previousHighest = tail.getOrElse(head.key, -1L)
     private[internal] val offsets = tail + (head.key -> head.offset.max(previousHighest))
+    private[internal] lazy val firstOffsetMap = first.get(head.key) match {
+      case None => first + (head.key -> head.offset)
+      case _ => first
+    }
```
seglo, line 76: I could probably merge this into a single map of first and end offsets.

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
```diff
@@ -379,6 +414,18 @@ private final class TransactionalSubSourceStageLogic[K, V](
...
+    onTransactionAborted.future.onComplete {
+      case Success(_) => onTransactionAbortedCb.invoke(())
+      case TFailure(ex) => log.error(ex, "An error occurred while during the transaction abort process")
+    }(materializer.executionContext)
+    onFirstMessageReceived.future.onComplete {
+      case Success(_) => onFirstMessageReceivedCb.invoke(())
+      case TFailure(ex) => log.error(ex, "The first message was not received")
+    }(materializer.executionContext)
+  }
+
```
seglo, line 428: Would be nice to consolidate this with the partitioned sources.

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
```diff
@@ -413,12 +460,8 @@ private final class TransactionalSubSourceStageLogic[K, V](
     case (sender, Committed(offsets)) =>
       inFlightRecords.committed(offsets.view.mapValues(_.offset() - 1).toMap)
       sender ! Done
-    case (sender, CommittingFailure) => {
-      log.info("Committing failed, resetting in flight offsets")
-      inFlightRecords.reset()
-    }
     case (sender, Drain(partitions, ack, msg)) =>
-      if (inFlightRecords.empty(partitions)) {
+      if (!firstMessageReceived || inFlightRecords.empty(partitions)) {
```
seglo, line 464: If a message is never received downstream then we don't need to drain.

tests/src/it/scala/akka/kafka/TransactionsPartitionedSourceSpec.scala
```diff
@@ -33,6 +33,12 @@ class TransactionsPartitionedSourceSpec extends SpecBase
   with Repeated {
 
   val replicationFactor = 2
+  // It's possible to get into a livelock situation where the `restartAfter` interval causes transactions to abort
+  // over and over.  This can happen when there are a few partitions left to process and they can never be fully
+  // processed because we always restart the stream before the transaction can be completed successfully.
+  // The `maxRestarts` provides an upper bound for the maximum number of times we restart the stream so if we get
+  // into a livelock it can eventually be resolved by not restarting any more.
+  val maxRestarts = new AtomicInteger(1000)
```
seglo, line 41: This situation was occurring before and after this fix.  This should make the tests _always_ complete, it just places an upper bound on the number of restarts.

tests/src/it/scala/akka/kafka/TransactionsSourceSpec.scala
```diff
@@ -55,13 +63,16 @@ class TransactionsSourceSpec extends SpecBase
...
+        (0 until sourcePartitions).map { part =>
+          val rangeStart = ((part * partitionSize) + 1)
+          val rangeEnd = (partitionSize * (part + 1))
+          log.info(s"Producing [$rangeStart to $rangeEnd] to partition $part")
+          produce(sourceTopic, rangeStart to rangeEnd, part)
+        }
 
       Await.result(Future.sequence(producers), 1.minute)
 
-      val consumerSettings = consumerDefaults.withGroupId(group)
```
seglo, line 64: This change should be rolled back.

tests/src/test/scala/akka/kafka/TransactionsOps.scala
```diff
@@ -162,11 +182,10 @@ trait TransactionsOps extends TestSuite with Matchers {
       .plainSource(settings, Subscriptions.topics(topic))
       .map(r => (r.partition(), r.offset(), r.value()))
       .take(elementsToTake)
-      .idleTimeout(30.seconds)
       .alsoTo(
         Flow[(Int, Long, String)]
           .scan(0) { case (count, _) => count + 1 }
-          .filter(_ % 100 == 0)
+          .filter(_ % 10 == 0)
```
seglo, line 188: rollback

tests/src/it/scala/akka/kafka/TransactionsSourceSpec.scala
```diff
@@ -101,7 +119,6 @@ class TransactionsSourceSpec extends SpecBase
 
       val consumer = offsetValueSource(probeConsumerSettings(probeConsumerGroup), sinkTopic)
         .take(elements.toLong)
-        .idleTimeout(30.seconds)
```
seglo, line 104: This is a magic number that can sometimes cause the test to abort too early before the transactional stream had a chance to complete.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -37,47 +40,61 @@ private[kafka] final class TransactionalProducerStage[K, V, P](
...
-    // Since `awaitingConfirmation` counter guarantees that all writes finished, we can safely assume
-    // that all all data up to maximal offsets has been wrote to Kafka.
-    private val previousHighest = tail.getOrElse(head.key, -1L)
-    private[internal] val offsets = tail + (head.key -> head.offset.max(previousHighest))
+
+    private val withHighest = tail.get(head.key) match {
+      // There is no guarantee that offsets adding callbacks will be called in any particular order.
+      // Decreasing an offset stored for the KTP would mean possible data duplication.
+      // Since `awaitingConfirmation` counter guarantees that all writes finished, we can safely assume
+      // that all all data up to maximal offsets has been written to Kafka.
```
ennru, line 74:
```suggestion
      // that all data up to maximal offsets has been written to Kafka.
```

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -88,8 +105,16 @@ private object TransactionalProducerStage {
...
     }
-  }
 
+    override def toString: String = {
+      val header = tpFirstEndOffsetFormat.format("TopicPartition", "First Observed Offset", "End Offset")
+      val rows = offsets.map {
+        case (gtp, FirstAndHighestOffsets(first, highest)) =>
+          tpFirstEndOffsetFormat.format(gtp.topicPartition.toString, first, highest)
+      }
+      "\n" + header + rows.mkString
```
ennru, line 115: Does this need to be multiline?

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -228,25 +259,36 @@ private final class TransactionalProducerStageLogic[K, V, P](
...
-    batchOffsets.committingFailed()
-    super.onCompletionFailure(ex)
+
+    batchOffsets
+      .transactionAborted()
+      .onComplete {
+        case Success(_) =>
+          log.debug("Successfully sent source stage failure")
+          onCommitFailedAckCb.invoke(ex)
+        case Failure(commitEx) if NonFatal(commitEx) =>
```
ennru, line 271: No need to check for `NonFatal`, futures wouldn't pass fatal exceptions.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -228,25 +259,36 @@ private final class TransactionalProducerStageLogic[K, V, P](
...
+      }(materializer.executionContext)
   }
 
   private def commitTransaction(batch: NonemptyTransactionBatch, beginNewTransaction: Boolean): Unit = {
     val group = batch.group
     log.debug("Committing transaction for transactional id '{}' consumer group '{}' with offsets: {}",
               transactionalId,
               group,
-              batch.offsets)
+              batch.toString)
```
ennru, line 282:
```suggestion
              batch)
```

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -228,25 +259,36 @@ private final class TransactionalProducerStageLogic[K, V, P](
...
     val offsetMap = batch.offsetMap()
     producer.sendOffsetsToTransaction(offsetMap.asJava, group)
     producer.commitTransaction()
     log.debug("Committed transaction for transactional id '{}' consumer group '{}' with offsets: {}",
               transactionalId,
               group,
-              batch.offsets)
+              batch.toString)
```
ennru, line 289:
```suggestion
              batch)
```

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -258,6 +300,12 @@ private final class TransactionalProducerStageLogic[K, V, P](
     }
   }
 
+  private val onCommitFailedAckCb: AsyncCallback[Throwable] = {
+    getAsyncCallback[Throwable](
+      t => super.onCompletionFailure(t)
```
ennru, line 305: The `super` is hideous, should it be just `failStage(t)`?

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -275,7 +323,7 @@ private final class TransactionalProducerStageLogic[K, V, P](
   }
 
   private def abortTransaction(reason: String): Unit = {
-    log.debug("Aborting transaction: {}", reason)
+    log.debug("Aborting transaction with reason: {}, offsets: {}", reason, batchOffsets.toString)
```
ennru, line 326:
```suggestion
    log.debug("Aborting transaction with reason: {}, offsets: {}", reason, batchOffsets)
```

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
```diff
@@ -139,22 +144,36 @@ private[internal] abstract class TransactionalSourceLogic[K, V, Msg](shape: Sour
 
   override val groupId: String = consumerSettings.properties(ConsumerConfig.GROUP_ID_CONFIG)
 
+  private val onTransactionAbortedCb = getAsyncCallback[Unit] { _ =>
+    log.debug(s"Committing failed, draining: $draining, resetting inFlightRecords: $inFlightRecords")
```
ennru, line 148:
```suggestion
    log.debug("Committing failed, draining: {}, resetting inFlightRecords: {}", draining, inFlightRecords")
```

Is this the only use of `draining`?

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
seglo: Yes, it produces a table.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
seglo: :+1:

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
seglo: We could, but if we ever implement anything in `DefaultProducerStage.onCompleteFailure` we may want it to execute that.

core/src/main/scala/akka/kafka/internal/TransactionalSources.scala
seglo: Yes.  It was for troubleshooting.  I don't think it's very useful anymore.  I'll remove it.

seglo:
New failure: https://travis-ci.org/github/akka/alpakka-kafka/jobs/691735521#L386

hepin1989:
@seglo Friendly ping, Any update?

ennru:
This turned out harder than it should be, unless someone in the community digs into it a lot, we won't be able to finish this PR.

patriknw:
I think we have to close this attempt

Issue #955, opened 2019-10-30
Log (and maybe fail) exceptions returned by offset commit callback


seglo:
### Short description

When exceptions are passed to the offset commit callback they are passed back to the original committer as an ask response which will fail the stream.  We should consider at least logging these exceptions, or shutting down the stream at the source instead.

### Details

We should consider logging that these exceptions occur in the `KafkaConsumerActor` and consider failing the stream as well.  The `processErrors` method of `KafkaConsumerActor` does both, perhaps it could be reused?  In some cases we may receive `KafkaExceptions` that are recoverable (retriable).  We could consider not propagating these exceptions and let the `KafkaConsumer` recover internally.

PR #1111, opened 2020-05-13
Committing: retry commits marked as retriable

ennru:
## Changes

* Test the commit callback exception for `RetriableCommitFailedException`
	* Re-add the commits and reply addresses to for aggregation with the next commit
	* Request an immediate poll
* Log other commit failures in the `KafkaConsumerActor` (as suggested in #955)
* Log "Kafka commit took longer than..." for successful commits only

## Background context

The Kafka client may report an exception to the commit async callback which marks the commit as not executed, but worth to be retried.
Before this change Alpakka Kafka would fail the stream for any kind of exception reported to the commit callback.

The "Kafka commit took longer than..." logging showed even for failed commit requests which gave the impression commits became slower and finally failed.

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -566,9 +566,11 @@ import scala.util.control.NonFatal
   private def commitAggregatedOffsets(): Unit = if (commitMaps.nonEmpty && !rebalanceInProgress) {
     val aggregatedOffsets = aggregateOffsets(commitMaps)
     commitRefreshing.add(aggregatedOffsets)
-    commit(aggregatedOffsets, commitSenders)
+    val replyTo = commitSenders
+    // flush the data before calling `consumer.commitAsync` which might call the callback synchronously
     commitMaps = List.empty
     commitSenders = Vector.empty
+    commit(aggregatedOffsets, replyTo)
   }
 
   private def commit(commitMap: Map[TopicPartition, OffsetAndMetadata], replyTo: Vector[ActorRef]): Unit = {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -582,18 +584,33 @@ import scala.util.control.NonFatal
...
+              log.warning("Kafka commit is to be retried, after={} ms, commitsInProgress={}, cause={}",
+                          e.getCause,
+                          duration / 1000000L,
+                          commitsInProgress)
+              commitMaps = commitMap.toList ++ commitMaps
+              commitSenders = commitSenders ++ replyTo
+              requestDelayedPoll()
+
+            case commitException =>
+              log.warning("Kafka commit failed after={} ms, commitsInProgress={}, exception={}",
```
seglo, line 608: Will returning this exception to the asker fail the stream?  Maybe it should be logged as an error if it's not already done later.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -566,9 +566,11 @@ import scala.util.control.NonFatal
   private def commitAggregatedOffsets(): Unit = if (commitMaps.nonEmpty && !rebalanceInProgress) {
     val aggregatedOffsets = aggregateOffsets(commitMaps)
     commitRefreshing.add(aggregatedOffsets)
-    commit(aggregatedOffsets, commitSenders)
+    val replyTo = commitSenders
+    // flush the data before calling `consumer.commitAsync` which might call the callback synchronously
```
seglo, line 570: Really?  That's counter intuitive based on the method name.  Good to know.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
ennru: Yes, that's the exception that fails the stream. But if the user doesn't properly listen to stream completion it will get lost.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
ennru: I didn't expect it to. But in one of my manual tests it appeared as if it did and when I moved these it worked...

ennru: It should really be the thread calling `poll` which executes the callback.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
ennru: I changed it to log as error.

ennru:
With #1584 commit retrying will happen for more types of retriable exceptions.

PR #1131, opened 2020-06-16
Add produce consume check to KafkaContainerCluster

seglo:
Add a message produce & consume check that exercises all brokers in the testcontainer cluster. This occurs during startup and will hopefully mitigate timeouts on the first assert run by tests that rely on the cluster.

### DIFF
--- a/testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
+++ b/testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
@@ -12,10 +12,12 @@ import org.rnorth.ducttape.unreliables.Unreliables;
 import org.testcontainers.DockerClientFactory;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
+import org.testcontainers.images.builder.Transferable;
 import org.testcontainers.lifecycle.Startable;
 import org.testcontainers.lifecycle.Startables;
 
 import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
```diff
@@ -101,24 +112,40 @@ public String getBootstrapServers() {
   }
 
   private Stream<GenericContainer> allContainers() {
-    Stream<GenericContainer> genericBrokers = this.brokers.stream().map(b -> (GenericContainer) b);
+    Stream<GenericContainer> genericBrokers = this.brokers.stream().map(b -> b);
```
ennru, line 115:
```suggestion
    Stream<GenericContainer> genericBrokers = this.brokers.stream();
```

testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
```diff
@@ -101,24 +112,40 @@ public String getBootstrapServers() {
...
     Stream<GenericContainer> zookeeper = Stream.of(this.zookeeper);
     return Stream.concat(genericBrokers, zookeeper);
   }
 
   @Override
   public void start() {
     try {
-      Stream<Startable> startables = this.brokers.stream().map(b -> (Startable) b);
+      Stream<Startable> startables = this.brokers.stream().map(b -> b);
```
ennru, line 123:
```suggestion
      Stream<Startable> startables = this.brokers.stream();
```

testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
seglo: This and the other example you highlighted are actually required to cast the inner type to the desired type.  I'm not sure why an identity function is required, but it's essentially the same as:

```
Stream<GenericContainer> genericBrokers = this.brokers.stream().map(b -> (GenericContainer) b);
```

I made the cast explicit to avoid confusion.

seglo: On second thought, the cast isn't actually necessary.  I've removed it.

PR #1132, opened 2020-06-26
Docs: show use of Protocol Buffers

ennru:


docs/src/main/paradox/serialization.md
```diff
@@ -3,6 +3,30 @@
 The general recommendation for de-/serialization of messages is to use byte arrays (or Strings) as value and do the de-/serialization in a `map` operation in the Akka Stream instead of implementing it directly in Kafka de-/serializers. When deserialization is handled explicitly within the Akka Stream, it is easier to implement the desired error handling strategy as the examples below show.
 
 
+## Protocol buffers
+
+[Protocol Buffers](https://developers.google.com/protocol-buffers) offer a language-neutral, platform-neutral, extensible mechanism for serializing structured data and allow consumers and producers to rely on the message format.
+
+The easiest way to use Protocol Buffers with Alpakka Kafka is to serialize and deserialize the Kafka message payload as byte array and call the Protocol Buffers serialization and deserialization in a regular `map` operator. To serialize the Protobuf-defined type `Order` into a byte array use the `.toByteArray()` method which gets generated by the Protobuf compiler.
```
seglo, line 10:
```suggestion
The easiest way to use Protocol Buffers with Alpakka Kafka is to serialize and deserialize the Kafka message payload as a byte array and call the Protocol Buffers serialization and deserialization in a regular `map` operator. To serialize the Protobuf-defined type `Order` into a byte array use the `.toByteArray()` method which gets generated by the Protobuf compiler.
```

tests/src/test/java/docs/javadsl/SerializationTest.java
```diff
@@ -24,6 +24,12 @@
 import com.fasterxml.jackson.databind.ObjectWriter;
```
seglo, line 24: Surprised these codegen'd files don't end up in `target`.

ignasi35: I'm guessing @ennru copy/pasted them so we don't need a protobuf plugin on the `docs/` build?

seglo: Good point

ennru:
To sort out the build of `SerializationSpec` and `SerializationTest` I moved out the tests using the schema registry to a separate class and make the remaining rely on Testcontainers instead.

Issue #1114, opened 2020-05-13
Drop testkit support for Embedded Kafka


seglo:
### Short description

Bundling embedded Kafka in the testkit can make it challenging to support new versions of Kafka in Alpakka Kafka as the release cycle lags Apache Kafka.  When we included Embedded Kafka on the classpath it can conflict with the core Kafka client dependency and its dependency tree.  Embedded Kafka with Schema Registry also relies on Confluent Platform images which typically lags the Apache Kafka release by 1-2 months and doesn't always support the same Scala versions.

### Details

* Update any remaining tests to use testcontainers instead of Embedded Kafka
* Re-factor `SerializationSpec` to use testcontainers.

seglo:
@ennru All that remains for this work is reorganizing docs. #1209 replaces the last of the tests that depend on embedded-kafka (except for docs snippets).

seglo:
If we decide not to drop it, then I think we should at least isolate the docs snippet tests in another project so they don't interfere with the rest of the tests.

ennru:
Kafka core inlines the Scala library which hinders to use Embedded Kafka with different Scala patch versions than the ones Kafka is compiled against.

This definitely decides that we drop Embedded Kafka support from the Alpakka Kafka testkit.

See https://github.com/akka/alpakka-kafka/pull/1212 and https://github.com/embeddedkafka/embedded-kafka/issues/202

PR #1135, opened 2020-06-29
Replace embedded kafka with schema registry with testcontainers

seglo:
References #1114


testkit/src/main/scala/akka/kafka/testkit/internal/TestcontainersKafka.scala
```diff
@@ -50,13 +51,35 @@ object TestcontainersKafka {
...
+        .map(_.getSchemaRegistryUrl)
+        .getOrElse(
+          throw new RuntimeException("Did you enable schema registry in your KafkaTestkitTestcontainersSettings?")
+        )
+
+    /**
+     * Deprecated: Starts Kafka, but does not override with provided settings.
+     */
+    @deprecated("Use startKafkaWithSettings instead", "2.0.3")
     def startKafka(settings: KafkaTestkitTestcontainersSettings): String = {
```
ennru, line 67: The next version is 2.0.4.
Why was the new `startKafkaWithSettings` needed?

seglo: This method implied that the provided settings would be used, but it ignored the `settings` param and applied settings from the overloaded `testcontainersSettings` property.

```
def startKafka(settings: KafkaTestkitTestcontainersSettings): String = {
  val settings = testcontainersSettings
```

I'm not sure why I did this originally, I think it was just a bug that slipped under the radar.

`startWithKafkaSettings` uses the provided settings.  I deprecated this method for backwards compatibility, and added an overload `startKafka()` which uses the overloaded `testcontainersSettings`.  Alternatively, instead of deprecating I could just fix the bug.

testkit/src/main/java/akka/kafka/testkit/javadsl/TestcontainersKafkaJunit4Test.java
```diff
@@ -7,11 +7,14 @@
 
 import akka.actor.ActorSystem;
 import akka.kafka.testkit.KafkaTestkitTestcontainersSettings;
+import akka.kafka.testkit.internal.SchemaRegistryContainer;
 import akka.kafka.testkit.internal.TestcontainersKafka;
 import akka.stream.Materializer;
 import org.junit.After;
 import org.junit.Before;
 
+import scala.compat.java8.OptionConverters;
```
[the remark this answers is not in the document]
seglo: Nope :+1:

testkit/src/main/scala/akka/kafka/testkit/internal/TestcontainersKafka.scala
ennru: Just fix that bug. It will do what the method promised in the first place and we do not promise strict bin-compatibility for testkits.

seglo: Ah, ok.

seglo:
> The test exclusion in build.sbt can go now, right?

Yup.

Issue #1070, opened 2020-03-02
[Failure] At-Least-Once One To Conditional should support batching of offsets `withOffsetContext`


ennru:
First seen in #1069 

```
AtLeastOnce:
At-Least-Once One To Conditional
- should support batching of offsets `withOffsetContext` *** FAILED *** (10 seconds, 431 milliseconds)
  Vector(ConsumerRecord(topic = topic-2-96, partition = 0, leaderEpoch = 0, offset = 0, CreateTime = 1583141715456, serialized key size = 3, serialized value size = 9, headers = RecordHeaders(headers = [], isReadOnly = false), key = key, value = 1,2,3,4,5), ConsumerRecord(topic = topic-2-96, partition = 0, leaderEpoch = 0, offset = 1, CreateTime = 1583141715456, serialized key size = 3, serialized value size = 10, headers = RecordHeaders(headers = [], isReadOnly = false), key = key, value = 6,7,8,9,10), ConsumerRecord(topic = topic-2-96, partition = 0, leaderEpoch = 0, offset = 0, CreateTime = 1583141715456, serialized key size = 3, serialized value size = 9, headers = RecordHeaders(headers = [], isReadOnly = false), key = key, value = 1,2,3,4,5), ConsumerRecord(topic = topic-2-96, partition = 0, leaderEpoch = 0, offset = 1, CreateTime = 1583141715456, serialized key size = 3, serialized value size = 10, headers = RecordHeaders(headers = [], isReadOnly = false), key = key, value = 6,7,8,9,10)) 
  had size 4 instead of expected size 2 (AtLeastOnce.scala:198)

  org.scalatest.exceptions.TestFailedException:
  at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:343)
  at org.scalatest.Matchers$ResultOfHaveWordForExtent.size(Matchers.scala:2707)
  at docs.scaladsl.AtLeastOnce.$anonfun$new$16(AtLeastOnce.scala:198)
```

https://travis-ci.org/akka/alpakka-kafka/jobs/657205159#L490

seglo:
https://travis-ci.org/github/akka/alpakka-kafka/jobs/699377074#L495

seglo:
https://travis-ci.org/github/akka/alpakka-kafka/jobs/705758567#L504

seglo:
[...]

PR #1139, opened 2020-07-07
Use bulk update API for shard allocation

chbatey:
On top of #1138


build.sbt
```diff
@@ -9,7 +9,7 @@ val Nightly = sys.env.get("TRAVIS_EVENT_TYPE").contains("cron")
 val Scala211 = "2.11.12"
 val Scala212 = "2.12.10"
 val Scala213 = "2.13.1"
-val akkaVersion26 = "2.6.3"
+val akkaVersion26 = "2.6.6"
```
ennru, line 12: Is this upgrade is required? If not we'd keep 2.6.3.

chbatey:
The new API is not in 2.6.3 but I could check what the min upgrade could be
if that helps? But if we’re upgrading maybe we should just go to 2.6.6

ennru:
Yes, when we update we go all the way.

seglo:
Unrelated build failure: https://github.com/akka/alpakka-kafka/issues/1070

Issue #1159, opened 2020-07-14
Mysterious consumer stop in 2.0.3 from draining offset batches on failure


jyates:
# Versions used

Scala 2.11.12
Akka: 2.5.[29, 30]
Alpakka-kafka: 2.0.3 and 2.0.3+[2-7].
kafka: 5.4.2-ccs

### Expected Behavior

We are running 2.0.2 in production without issue. 

### Actual Behavior

With 2.0.3 we are seeing the source logging
```
 a.k.i.SingleSourceLogic:92 - [34665] Completing
```

And the consumer stopping. This happens only on a couple of our consumers, but for no obvious reason. No stack traces, nothing untoward in the logs.

 It is bad enough that the consumers stop making tangible progress due to the frequency of rebalances. We see this much more in some consumer groups than others, but it is reproducible in the same consumer, i.e. it seems somehow related to processing patterns.

When turning on DEBUG logging, the issue still occurs and no obvious stack traces or other issues are logged. However, when turning on actor message debugging, ie.
```
akka.actor.debug {
      receive = true
      autoreceive = true
      lifecycle = true
      unhandled = true
}
```
the issue goes away for much longer (10s of minutes rather than within seconds), but still appears. `autorecieve=true` appears to be the main determinant of time until the issue appears.

### Relevant logs

Redacted, DEBUG (akka.kafka, but not actor message handling) logs
https://gist.github.com/jyates/e9b761f14c05a1fff78e08e9262203fa

### Reproducible Test Case

Unfortunately, no obvious test case as its rare even in prod.

jyates:
After walking the commit history, through production (only reliable place to reproduce), it appears the issue came from #1058 . Still have to dig a big more to figure out root cause, but any hints would be helpful if folks get a chance.

jyates:
I'm convinced now that there is a bug in the `CommitCollectorStage`. Commented out and running in prod for much longer than any other time taken to see the issue (often just minutes) has proven to cause no problems, ie. the change
```diff
@@ -28,8 +28,9 @@ object Committer {
    */
   def batchFlow(settings: CommitterSettings): Flow[Committable, CommittableOffsetBatch, NotUsed] = {
     val offsetBatches: Flow[Committable, CommittableOffsetBatch, NotUsed] =
-      Flow
-        .fromGraph(new CommitCollectorStage(settings))
+      Flow[Committable]
[...]

PR #1160, opened 2020-07-16
Emit offset batches on timer

jyates:
Rather than push, as downstream may not yet have pulled
References #1159


jyates:
cc @seglo

seglo:
Thanks for the fix, @jyates!  This will be included in the next Alpakka Kafka release I'm pushing this week, 2.0.4.

PR #1162, opened 2020-07-21
Create metadata client actor as system actor

### LINKED ISSUE #1115: MetadataClient should create the ConsumerActor with system.systemActorOf
https://github.com/akka/alpakka-kafka/blob/eb24136474e8c7a113ff353ed50ffd2c3ba08f60/core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala#L117

Otherwise that can't be used with a Typed ActorSystem (real Typed with user guardian Behavior)

seglo:
Fixes #1115

### DIFF
--- a/core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
+++ b/core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
@@ -5,24 +5,9 @@
 
 package akka.kafka.scaladsl
 
-import java.util.Collections
-
-import akka.actor.{ActorRef, ActorSystem}
+import akka.actor.{ActorRef, ActorSystem, ExtendedActorSystem}
 import akka.dispatch.ExecutionContexts
-import akka.kafka.Metadata.{
-  BeginningOffsets,
-  CommittedOffset,
-  CommittedOffsets,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/akka/kafka/scaladsl/MetadataClient.scala
```diff
@@ -114,7 +99,9 @@ object MetadataClient {
       consumerSettings: ConsumerSettings[K, V],
       timeout: Timeout
   )(implicit system: ActorSystem, ec: ExecutionContext): MetadataClient = {
-    val consumerActor = system.actorOf(KafkaConsumerActor.props(consumerSettings))
+    val consumerActor = system
+      .asInstanceOf[ExtendedActorSystem]
+      .systemActorOf(KafkaConsumerActor.props(consumerSettings), "alpakka-kafka-metadata-client")
```
johanandren, line 104: A second call would fail because name taken. So it either needs an atomiclong or such to give it a unique id, or use an extension to keep it one shared actor per system

seglo: Ah, right.  I'll use an atomic long.

PR #1177, opened 2020-08-10
Cleanup build warnings

seglo:
A mixed bag of changes to address warnings and linting errors in the build.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -276,7 +276,7 @@ lazy val clusterSharding = project
   .in(file("./cluster-sharding"))
   .dependsOn(core)
   .enablePlugins(AutomateHeaderPlugin)
-  .disablePlugins(MimaPlugin, SitePlugin) // TODO: re-enable MiMa plugin after first release
+  .disablePlugins(MimaPlugin, SitePlugin)
   .settings(commonSettings)
   .settings(
     name := "akka-stream-kafka-cluster-sharding",
--- a/build.sbt
+++ b/build.sbt
@@ -276,7 +276,7 @@ lazy val clusterSharding = project
   .in(file("./cluster-sharding"))
   .dependsOn(core)
   .enablePlugins(AutomateHeaderPlugin)
-  .disablePlugins(MimaPlugin, SitePlugin)
+  .disablePlugins(SitePlugin)
   .settings(commonSettings)
   .settings(
     name := "akka-stream-kafka-cluster-sharding",
--- a/build.sbt
+++ b/build.sbt
@@ -96,7 +96,8 @@ val commonSettings = Def.settings(
   scalaVersion := Scala212,
   crossVersion := CrossVersion.binary,
   javacOptions ++= Seq(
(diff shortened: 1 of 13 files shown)


core/src/main/scala/akka/kafka/RestrictedConsumer.scala
```diff
@@ -36,6 +37,7 @@ final class RestrictedConsumer(consumer: Consumer[_, _], duration: java.time.Dur
   /**
    * See [[org.apache.kafka.clients.consumer.KafkaConsumer#committed(TopicPartition,java.time.Duration)]]
    */
+  @silent
   def committed(tp: TopicPartition): OffsetAndMetadata = consumer.committed(tp, duration)
```
seglo, line 41: Perhaps this should be deprecated as well.

testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
```diff
@@ -214,11 +204,16 @@ private String readinessCheckScript() {
...
+            + connect
+            + " --describe --topic "
+            + READINESS_CHECK_TOPIC
+            + " | wc -l) > 1 ]] && "
+            + "kafka-topics "
             + connect
             + " --delete --topic "
             + READINESS_CHECK_TOPIC
-            + " || echo \"topic does not exist\" \n";
+            + " \n";
```
seglo, line 216: Gracefully check that the topic exists before trying to delete it. Previous version polluted stderr when readiness check ran successfully.

seglo:
I didn't look at evicted dependencies, but maybe we could clean that up as well.

build.sbt
```diff
@@ -275,7 +276,7 @@ lazy val clusterSharding = project
   .in(file("./cluster-sharding"))
   .dependsOn(core)
   .enablePlugins(AutomateHeaderPlugin)
-  .disablePlugins(MimaPlugin, SitePlugin) // TODO: re-enable MiMa plugin after first release
+  .disablePlugins(MimaPlugin, SitePlugin)
```
ennru, line 279: I guess we should enable Mima?

core/src/main/scala/akka/kafka/RestrictedConsumer.scala
ennru: If `consumer.committed` is deprecated, we should deprecate as well.

seglo: I'll make a follow up PR that adds the replacement and deprecates this instance.

build.sbt
seglo: :+1:

core/src/main/scala/akka/kafka/RestrictedConsumer.scala
seglo: https://github.com/akka/alpakka-kafka/pull/1178

Issue #1128, opened 2020-06-11
Failed: AssignmentSpec: subscription with partition assignment must consume from the specified topic pattern


seglo:
From an unrelated PR build

https://travis-ci.org/github/akka/alpakka-kafka/jobs/696760319#L559

```
Failed(org.scalatest.exceptions.TestFailedException: 1 was not equal to 200)
```

seglo:
Attempt to fix in #1181

PR #1181, opened 2020-08-17
Simplify consuming stream for AssignmentSpec "consume from the specified topic pattern" test

seglo:
Addresses #1128 

I couldn't reproduce the failure in #1128, but I thought the test was a little strange so I simplified it.  With the old approach I suppose it's possible an early cancellation signal could shutdown the stream before the primary branch returns all the elements.


seglo:
Unrelated build failures: https://github.com/akka/alpakka-kafka/issues/1185

PR #1182, opened 2020-08-17
Reduce bloat of log when testsuite log reported in failed travis build

seglo:
I've noticed that logs from failing Travis builds have been quite large for awhile. If you look at any recent build failure you'll see after all the logs for each individual test the repetition of this logging from the `CommitCollectorStage`:

```
...
2020-08-17 15:38:29,197 DEBUG [default-dispatcher-8] [a.k.i.CommitCollectorStageLogic     ]  [310bf] Consuming offset CommittableOffsetImpl(PartitionOffset(key=GroupTopicPartition(group1,topic,0),offset=100),metadata)
2020-08-17 15:38:29,197 DEBUG [default-dispatcher-8] [a.k.i.CommitCollectorStageLogic     ]  [310bf] Consuming offset CommittableOffsetImpl(PartitionOffset(key=GroupTopicPartition(group1,topic,0),offset=101),metadata 2)
...
```

i.e. from build https://travis-ci.org/github/akka/alpakka-kafka/jobs/718663608

These messages are generated in `TestkitSampleSpec` where an iterator is used to demonstrate generating mock messages using the `CommitResultFactory`.  A `Source.cycle` will repeat the iterator for the period of 1 second in a tight loop.  When run locally this produces a ~30MB log.  On travis it will exceed the maximum log length of the job and terminate it with the message `The job exceeded the maximum log length, and has been terminated.`.

For the purpose of demonstrating this testkit utility I think it's fine to just generate 10 messages and let the stream shutdown naturally.

We may also want to reconsider having this log line in `CommitCollectorStage` in general since it's probably more noise than you want to see even while in `DEBUG`.

### DIFF
--- a/tests/src/test/scala/docs/scaladsl/TestkitSamplesSpec.scala
+++ b/tests/src/test/scala/docs/scaladsl/TestkitSamplesSpec.scala
@@ -19,8 +19,6 @@ import org.apache.kafka.clients.producer.ProducerRecord
 import org.scalatest.concurrent.{IntegrationPatience, ScalaFutures}
 import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
 
-import scala.collection.immutable
-
 class TestkitSamplesSpec
     extends TestKit(ActorSystem("example"))
     with FlatSpecLike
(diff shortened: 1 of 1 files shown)


seglo:
Unrelated build failure: https://github.com/akka/alpakka-kafka/issues/895

Issue #1187, opened 2020-08-20
Allow `ClassicActorSystemProvider` in `DiscoverySupport`


ennru:
## Short description

The current `DiscoverySupport` implementations expect a classic actor system to be available.
Instead, they should use `ClassicActorSystemProvider` which allows to pass both classic or typed actor systems.

ennru:
Fixed with #1192

PR #1192, opened 2020-08-31
Allow typed Actor Systems in SendProducer and DiscoverySupport

ennru:
* upgrade to Akka 2.5.31 for better `ActorSystemProvider` integration
* duplicate construction methods for `SendProducer` to support the new actors API (typed) directly
* duplicate construction methods in `DiscoverySupport` to allow direct access with the new actors API #1187


core/src/main/scala/akka/kafka/scaladsl/SendProducer.scala
```diff
@@ -96,6 +96,10 @@ final class SendProducer[K, V] private (val settings: ProducerSettings[K, V])(im
 }
 
 object SendProducer {
-  def apply[K, V](settings: ProducerSettings[K, V])(implicit system: ActorSystem): SendProducer[K, V] =
-    new SendProducer(settings)
+  def apply[K, V](settings: ProducerSettings[K, V])(implicit system: ClassicActorSystemProvider): SendProducer[K, V] =
```
seglo, line 99: It's interesting that this remains binary compatible (according to MiMa).  I did a little experimentation with overloaded methods that contain different implicit parameter types and discovered the magnet pattern solution in [this SO answer](https://stackoverflow.com/a/17366915/895309). I tried implementing it in a branch and it breaks binary compatibility, but seems code compatibility.

https://github.com/akka/alpakka-kafka/compare/master...seglo:seglo/allow-classic-provider#diff-02cd8e6bc99d9aaa7e2ebdbe16e9d121R98-R114

Since your solution appears to satisfy both it seems like the right approach to me.

core/src/main/scala/akka/kafka/javadsl/DiscoverySupport.scala
```diff
@@ -28,24 +29,43 @@ object DiscoverySupport {
...
+      system: ClassicActorSystemProvider
   ): java.util.function.Function[ConsumerSettings[K, V], CompletionStage[ConsumerSettings[K, V]]] = {
+    implicit val sys: ClassicActorSystemProvider = system
     val function: ConsumerSettings[K, V] => Future[ConsumerSettings[K, V]] =
-      scaladsl.DiscoverySupport.consumerBootstrapServers(config)(system)
+      scaladsl.DiscoverySupport.consumerBootstrapServers(config)
     function.andThen(FutureConverters.toJava).asJava
   }
 
+  @InternalApi // kept for bin-compatibility
```
seglo, line 40: I'm confused why you marked this and the other overloads in `DiscoverySupport` as an internal API.  Shouldn't it be deprecated instead?

ennru: Internal API might be wrong, you're right.
I couldn't decide if we'd go all in and do `@deprecated` + `private[kafka]`.
Looking at it again, there is no way to call this method when recompiling so we can go all in.

ennru: Mima dislikes `private[kafka]` nowadays, so I stayed away from it.

PR #1202, opened 2020-09-17
Prefer current over 2.6 in url segment

ignasi35:
Using URLs with `/2.6` causes the warning "This is not the latest version of docs" to render.

It's probably fine in all cases to use `/current` instead.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -395,12 +395,12 @@ lazy val docs = project
         "javadoc.akka.link_style" -> "direct",
         "extref.akka-management.base_url" -> s"https://doc.akka.io/docs/akka-management/current/%s",
         // Akka 2.6. These can be removed when we drop Akka 2.5 support.
-        "akka.version26" -> akkaVersion26,
-        "extref.akka26.base_url" -> s"https://doc.akka.io/docs/akka/$AkkaBinaryVersion26/%s",
-        "scaladoc.akka.actor.typed.base_url" -> s"https://doc.akka.io/api/akka/$AkkaBinaryVersion26/",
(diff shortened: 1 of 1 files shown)


raboof:
> Using URLs with `/2.6` causes the warning "This is not the latest version of docs" to render.

i think we should fix that, instead.

seglo:
We still release referencing Akka 2.5.x with the exception of the Cluster Sharding module.  That's why there's two sets of docpaths.

It's curious that https://doc.akka.io/docs/akka/2.5/stream/index.html produces the warning, but https://doc.akka.io/docs/akka/2.5.31/stream/index.html does not.  Shouldn't the former be pointing to the "latest" for that major version?

> i think we should fix that, instead.

@raboof Is that the same thing you're referring to?

ignasi35:
> i think we should fix that, instead.

Good point, I'll close this.

raboof:
> We still release referencing Akka 2.5.x with the exception of the Cluster Sharding module. That's why there's two sets of docpaths.
> 
> It's curious that https://doc.akka.io/docs/akka/2.5/stream/index.html produces the warning, but https://doc.akka.io/docs/akka/2.5.31/stream/index.html does not. Shouldn't the former be pointing to the "latest" for that major version?
> 
> > i think we should fix that, instead.
> 
> @raboof Is that the same thing you're referring to?

Yes: indeed https://doc.akka.io/docs/akka/2.5/stream/index.html shows the 2.5.31 docs, but the javascript gets confused and incorrectly shows the warning (same for 2.6)

raboof:
(fixed on the Akka side with https://github.com/akka/akka/pull/29647)

Issue #1186, opened 2020-08-19
Cannot push port twice, or before it being pulled


wernerdegroot:
### Versions used 

`akka-stream-kafka` version 2.0.4
`akka-actor` version 2.5.30

### Expected Behavior

No exceptions

### Actual Behavior

Getting `IllegalArgumentException` once every couple of hours

### Relevant logs

```
Error in stage [akka.kafka.internal.CommitCollectorStage@48c3c78b]: Cannot push port (FlowOut(1890132076)) twice, or before it being pulled
java.lang.IllegalArgumentException: Cannot push port (FlowOut(1890132076)) twice, or before it being pulled
	at akka.stream.stage.GraphStageLogic.push(GraphStage.scala:642)
	at akka.kafka.internal.CommitCollectorStageLogic.$anonfun$consume$1(CommitCollectorStage.scala:60)
	at akka.kafka.internal.CommitCollectorStageLogic.$anonfun$consume$1$adapted(CommitCollectorStage.scala:60)
	at akka.kafka.internal.CommitCollectorStageLogic.akka$kafka$internal$CommitCollectorStageLogic$$pushDownStream(CommitCollectorStage.scala:80)
	at akka.kafka.internal.CommitCollectorStageLogic.akka$kafka$internal$CommitCollectorStageLogic$$consume(CommitCollectorStage.scala:60)
	at akka.kafka.internal.CommitCollectorStageLogic$$anon$1.onPush(CommitCollectorStage.scala:90)
	at akka.stream.impl.fusing.GraphInterpreter.processPush(GraphInterpreter.scala:523)
	at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:480)
	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:376)
	at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:606)
	at akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:485)
	at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:581)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:749)
[...]

PR #1205, opened 2020-09-25
Commit collector: avoid pushing twice

ennru:
The first commit tackles #1186 by explicitly checking if out can be pushed. The new tests did not illustrate the possible unlucky timing which may lead to pushing it twice.

The other commits simplify things in independent steps.

References #1186

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
@@ -46,7 +46,7 @@ private final class CommitCollectorStageLogic(
 
   override protected def logSource: Class[_] = classOf[CommitCollectorStageLogic]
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
```diff
@@ -46,79 +46,77 @@ private final class CommitCollectorStageLogic(
 
   override protected def logSource: Class[_] = classOf[CommitCollectorStageLogic]
 
-  // ---- initialization
+  private var emitOnPull = false
```
seglo, line 49: Naming suggestion: `pushOnNextPull`.  We only emit now on shutdown and the variable is toggled on the next pull.

seglo (approved):
LGTM

PR #1213, opened 2020-10-07
Use the latest JDK 8 and 11 versions in CI

ennru:


ignasi35 (approved):

ennru:
The build failure might be caused by #1214

Issue #1106, opened 2020-05-06
False positive ProducerMessage.Results


yiminyuan:
**Versions**:
- Scala 2.12.11
- Akka 2.5.26
- Alpakka Kafka 2.0.2

I have the message loss problem if the Kafka producer meets `NotLeaderForPartitionException` when partitions got reassigned. Please check https://github.com/akka/alpakka-kafka/issues/438 for symptoms.

Let's say I have a pattern like this:
```scala
source
  .map { _ =>
    ProducerMessage.single(...)
  }
  .via(Producer.flexiFlow(producerSettings))
  .map {
    case ProducerMessage.Result(metadata, ProducerMessage.Message(...)) =>
      s"Successfully produced message for partition ${metadata.partition}, offset ${metadata.offset}: ${record.value}"
  }
  .runWith(...)
```
Then I may have some false positive logs for those messages that have been dropped. Let's say message A is lost and message B has been sent successfully, we could have logs like this:
```
Successfully produced message for partition 0, offset 1: A
Successfully produced message for partition 0, offset 1: B
```
The log for A is false positive. Notice that A and B have the **SAME** metadata.

I saw we have an issue at https://github.com/akka/alpakka-kafka/issues/1101 to handle the publication failure. I wonder if this could be a part of that improvement.

ennru:
Thanks for the report. 
Interesting that the send callback is called as if it was successful.

yiminyuan:
I figured out the root cause of this problem. If a partition reassignment causes leader reelection, then the record could be lost due to the default Kafka producer's setting: `acks=1`.

I drew a diagram for the scenario in this ticket (Assume the topic has 3 replicas A, B, C, and all of them are in-sync. So any of them could be elected as a new leader):
![01](https://user-images.githubusercontent.com/13373154/84212309-59ee3b00-aa72-11ea-8d32-50b00e6721c7.jpg)
![02](https://user-images.githubusercontent.com/13373154/84212316-5c509500-aa72-11ea-9f6e-dabe913692a0.jpg)
![03](https://user-images.githubusercontent.com/13373154/84212319-5eb2ef00-aa72-11ea-97a3-9d790c969930.jpg)
![04](https://user-images.githubusercontent.com/13373154/84212323-607cb280-aa72-11ea-9848-f38a5b08aa39.jpg)
![05](https://user-images.githubusercontent.com/13373154/84212325-62df0c80-aa72-11ea-87fe-4915ea0464bd.jpg)

[...]

Issue #1223, opened 2020-10-08
Benchmarks fail since Kafka 2.6.0 upgrade


ennru:
Something is off with the versions in the Benchmarks.

```
08:32:36.823 INFO  [ntainers-lifecycle-1] 🐳 [confluentinc/cp-kafka:5.4.0]      Container confluentinc/cp-kafka:5.4.0 started in PT11.652S
08:32:52.788 ERROR [-thread | producer-1] o.a.kafka.common.utils.KafkaThread    Uncaught exception in thread 'kafka-producer-network-thread | producer-1':
java.lang.ClassNotFoundException: com.fasterxml.jackson.databind.JsonNode
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at sbt.internal.ManagedClassLoader.findClass(ManagedClassLoader.java:98)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 6 common frames omitted
Wrapped by: java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/JsonNode
	at org.apache.kafka.common.requests.ApiVersionsRequest$Builder.<clinit>(ApiVersionsRequest.java:36)
	at org.apache.kafka.clients.NetworkClient.handleConnections(NetworkClient.java:910)
	at org.apache.kafka.clients.NetworkClient.poll(NetworkClient.java:555)
	at org.apache.kafka.clients.producer.internals.Sender.runOnce(Sender.java:325)
	at org.apache.kafka.clients.producer.internals.Sender.run(Sender.java:240)
	at java.lang.Thread.run(Thread.java:748)
```

https://jenkins.akka.io:8498/job/alpakka-kafka-perf-master/1324/consoleFull

**I've disabled the benchmarks until this has been investigated.**

ennru:
This was due to an inconsistent Confluent Platform version which now is fixed in #1224.

PR #1224, opened 2020-10-09
Align Confluent Platform version to 6.0.0

ennru:
Follow-up to #1106

### DIFF
--- a/testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
+++ b/testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
@@ -9,7 +9,6 @@ import akka.annotation.InternalApi;
 import com.github.dockerjava.api.command.ExecCreateCmdResponse;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.model.ContainerNetwork;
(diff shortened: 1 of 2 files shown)


ennru:
References #1223

testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
```diff
@@ -32,6 +31,7 @@
 
   private static final String STARTER_SCRIPT = "/testcontainers_start.sh";
 
+  // Align this with testkit/src/main/resources/reference.conf
```
ignasi35, line 34: Super off-topic: it'd be great if we had a comment indicating `CP` stands for `Confluent Platform`

ignasi35 (approved):

testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
ignasi35: I'm talking about the name of this constant: `DEFAULT_CP_PLATFORM_VERSION`

seglo: That's an artifact of this type which was from https://github.com/testcontainers/testcontainers-java and modified for our purposes.

Issue #1225, opened 2020-10-09
Adapt `AlpakkaKafkaContainer` to Testcontainers 1.15


ennru:
The upcoming Testcontainers 1.15 version brings important improvements to the framework.
https://github.com/testcontainers/testcontainers-java/releases

As `AlpakkaKafkaContainer` replaces `KafkaContainer` it needs to be updated with the latest changes from Testcontainers. Specifically, https://github.com/testcontainers/testcontainers-java/pull/3293 adapts it to work with the Kafka broker bundled in Confluent Platform 6.0.0.

agemooij:
Any idea when this issue can be fixed and released?

All our Kafka tests started failing on my machine after I innocently said yes to a new stable release (2.5.0.0) of the docker for Mac distro. This triggered [an issue in testcontainers](https://github.com/testcontainers/testcontainers-java/pull/3159) that causes it to fail catastrophically.

I'll try downgrading my docker client to pre-2.4 for now but it would be great if this incompatibility could be fixed ASAP.

ennru:
With #1257 Alpakka Kafka uses Testcontainers 1.15.0.

@agemooij I would expect overriding the Testcontainers version in 3.0.0-M1 works. We'd want to progress a bit more before releasing M2.

agemooij:
OK, thanks! I'll try that. The Docker for Mac downgrade worked for me as well but I assume quite a few people will start running into this as they upgrade their systems so probably good to clearly mention it in the release notes and perhaps even as a known issue for the 2.0 releases.

agemooij:
@ennru I guess you meant the 2.1.0-M1 release? Or are you going to relabel it to 3.0 for M2?

ennru:
My bad, I mixed things up with main Alpakka. 
The upcoming Alpakka Kafka version is 2.1.0.

PR #1226, opened 2020-10-09
Rollback Confluent Platform version; add Jackson Databind

ennru:
The Kafka client depends on Jackson Databind but marks it as provided.
https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/2.6.0

Roll back the Confluent Platform as 6.0.0 requires changes to Testcontainers #1225

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,10 +15,12 @@ val AkkaBinaryVersion26 = "2.6"
 val AkkaBinaryVersion = if (Nightly) AkkaBinaryVersion26 else AkkaBinaryVersion25
 
 val kafkaVersion = "2.6.0"
+// TODO Jackson is now a provided dependency of kafka-clients
+// https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/2.6.0
 val jacksonVersion = "2.10.5"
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -15,9 +15,12 @@ val AkkaBinaryVersion26 = "2.6"
 val AkkaBinaryVersion = if (Nightly) AkkaBinaryVersion26 else AkkaBinaryVersion25
 
 val kafkaVersion = "2.6.0"
+// TODO Jackson is now a provided dependency of kafka-clients
+// https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/2.6.0
+val jacksonVersion = "2.10.5"
```
ignasi35, line 20: Could we move the declaration of the `val` elsewhere? (one line up or a few lines down)

It's now sitting between `kafkaVersion` and `embeddedKafkaVersion`/`kafkaVersionForDocs` which will often stay in sync.

ignasi35 (approved):
LGTM with a really small nitpick

build.sbt
ennru: I'll leave that for the build file rewrite.

PR #1227, opened 2020-10-10
CI: actually compile cluster sharding

ennru:
As it turns out, the build step to compile cluster sharding didn't run anything in sbt as the environment variable wasn't correctly set.


raboof (approved):
Oops!

PR #1229, opened 2020-10-13
Drop embedded-kafka testkit support

### LINKED ISSUE #1114: Drop testkit support for Embedded Kafka

seglo:
Fixes #1114


seglo:
I'm having some trouble getting mima exceptions to work on the testkit.  I'll look into further tomorrow, unless someone knows what's wrong off hand.

seglo:
I had the `mima-filters` dir in the wrong location.

ennru:
The `RetentionSpec` uses Kafka-internals https://travis-ci.org/github/akka/alpakka-kafka/jobs/735859971#L353

tests/src/test/scala/akka/kafka/scaladsl/ReconnectSpec.scala
```diff
@@ -195,13 +135,14 @@ class ReconnectSpec extends SpecBase with TestcontainersKafkaPerClassLike {
...
       Await.ready(control.shutdown(), remainingOrDefault)
+      Await.ready(proxyBinding2.unbind(), remainingOrDefault)
     }
   }
 
   /**
    * Create a proxy so it can be shut down.
    */
-  def createProxy(): (Future[Tcp.ServerBinding], Future[UniqueKillSwitch]) = {
+  def createProxy(): (Tcp.ServerBinding, Future[UniqueKillSwitch]) = {
```
ennru, line 145: I wonder if this really works. The Kafka broker will distribute all the broker addresses to the client anyway, so the client will talk to the non-proxied port once it is connected.

testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
```diff
@@ -32,7 +32,7 @@
   private static final String STARTER_SCRIPT = "/testcontainers_start.sh";
 
   // Align this with testkit/src/main/resources/reference.conf
-  public static final String DEFAULT_CONFLUENT_PLATFORM_VERSION = "5.4.0";
+  public static final String DEFAULT_CONFLUENT_PLATFORM_VERSION = "6.0.0";
```
ennru, line 35: Does our current Testcontainers version work with 6.0.0?
(The compile error makes me believe you haven't tried.)

ennru (commented):
Looks like you didn't notice #1230

testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
seglo: I was premature on the ready to review button.

tests/src/test/scala/akka/kafka/scaladsl/ReconnectSpec.scala
seglo: Good point. We'll need to think of some other way to hangup the connection using the container to preserve these tests. The problem with stopping and starting the container is that state is blown away.

testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
seglo: Fixed

tests/src/test/scala/akka/kafka/scaladsl/ReconnectSpec.scala
seglo: I tried a few things today:

* I'm able to pause and unpause the broker containers in place of embedded-kafka stop and start. This doesn't appear to kill any active connections. When I enable Kafka `NetworkClient` logs I can see with the previous embedded-kafka impl. a disconnect would occur.
* I can disconnect and reconnect the network to broker containers. This will force any active connections to close, but when the network is reattached the host port is different. Testcontainers will always generate a random host port and there doesn't appear to be a way around this. https://www.testcontainers.org/features/networking/

Some other ideas

* Use something like iptables to disable connections in the running container itself, but I can't find any of the usual programs in the image to do this (`iptables`, `route`) or even network interface commands.
* Stop and start the broker process within the running container. This should do the trick, but it will require some work to update the container run command to support it.

ennru: Create an issue about this instead and point to it from the test.

tests/src/test/scala/akka/kafka/scaladsl/RetentionPeriodSpec.scala
```diff
@@ -48,129 +41,130 @@ class RetentionPeriodSpec extends SpecBase with TestcontainersKafkaPerClassLike
...
-      probe1.cancel()
-      probe2.cancel()
-      probe3.cancel()
-
-      control1.isShutdown.futureValue shouldBe Done
-      control2.isShutdown.futureValue shouldBe Done
-      control3.isShutdown.futureValue shouldBe Done
-    }
-  }
+  // TODO: this test needs to manually consume events from __consumer_offsets without the aid of types from Kafka core
```
ennru, line 44: Create an issue about this as well. At least we know it worked before...
Could we just copy in a few classes?

ennru (approved):
LGTM.

tests/src/test/scala/akka/kafka/scaladsl/RetentionPeriodSpec.scala
seglo: Copying code might be a good idea.  I created an issue https://github.com/akka/alpakka-kafka/issues/1234

testkit/src/main/mima-filters/2.0.5.backwards.excludes/PR1229-drop-embedded-kafka.excludes
```diff
@@ -0,0 +1,4 @@
+ProblemFilters.exclude[MissingClassProblem]("akka.kafka.testkit.javadsl.EmbeddedKafkaJunit4Test")
+ProblemFilters.exclude[MissingClassProblem]("akka.kafka.testkit.javadsl.EmbeddedKafkaTest")
+ProblemFilters.exclude[MissingClassProblem]("akka.kafka.testkit.scaladsl.EmbeddedKafkaLike")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.testkit.scaladsl.KafkaSpec.mat")
```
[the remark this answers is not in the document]
seglo: I was toying with the idea of re-adding it and adding a deprecated flag to it so we can remove it later.  WDYT?

.scala-steward.conf
```diff
@@ -5,7 +5,6 @@ updates.ignore = [
 ]
 updates.pin = [
   { groupId = "org.apache.kafka", artifactId="kafka-clients", version="2.4." }
```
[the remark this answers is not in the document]
seglo: :+1:

tests/src/test/scala/akka/kafka/scaladsl/ReconnectSpec.scala
seglo: https://github.com/akka/alpakka-kafka/issues/1233

testkit/src/main/mima-filters/2.0.5.backwards.excludes/PR1229-drop-embedded-kafka.excludes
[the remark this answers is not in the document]
ennru: You're right, a deprecation makes sense.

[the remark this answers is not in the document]
seglo: On second thought, this just creates a lot of unnecessary noise.

PR #1230, opened 2020-10-14
Testkit: enable Mima

ennru:
Even though Akka testkits do not promise binary compatibilty nor source compatibility we should use Mima to detect changes to the API.﻿


seglo (approved):
LGTM.  Didn't get to this yesterday.  I'll rebase and update the Mima changes I made in #1229.

PR #1209, opened 2020-10-01
Drop Akka 2.5 dependency

seglo:
Fixes #1113

PR #1231, opened 2020-10-15
DiscoverySupport: remove deprecation

ennru:
With the deprecation warning in place it became impossible to avoid it in user code without doing effort as the compiler prefers the exact type of the classic actor system.

This removes the deprecation so that a classic actor system can be passed in.﻿

See https://github.com/akka/alpakka-kafka/pull/1209#discussion_r504450434


seglo (approved):
LGTM.

You can also remove the `@SuppressWarnings("deprecation")` attributes I added to `ProducerSettingsTest` and `ConsumerSettingsTest` from #1209.

ennru:

ennru:

PR #1235, opened 2020-10-19
Start & stop Kafka process within running container

### LINKED ISSUE #1233: Simulate downed broker with testcontainers in ReconnectSpec
### Short description

As part of the effort to remove embedded-kafka (#1114) I rediscovered a test that simulates a broker being lost to assure that Alpakka Kafka reconnects gracefully and completes consuming messages. This was fairly trivial to do with Embedded Kafka, but is a challenge to recreate with testcontainers.

### Details

Ideally we would simply stop and start the container again, but `testcontainers-java` will implicitly stop and remove state when stop any containers.  Even if we could stop the container without removing state, bringing it back online will still generate a random host port which won't be known by Alpakka Kafka when it's trying to reconnect.  There's `testcontainers-java` issue for this functionality that discusses some of the challenges https://github.com/testcontainers/testcontainers-java/issues/606.

Revised details from https://github.com/akka/alpakka-kafka/pull/1229#discussion_r505353715 to discuss what I've tried so far:

* I'm able to pause and unpause the broker containers in place of embedded-kafka stop and start. This doesn't appear to kill any active connections. When I enable Kafka `NetworkClient` logs I can see with the previous embedded-kafka impl. a disconnect would occur.

i.e.)

```
      // pause Kafka brokers
      this.brokerContainers.foreach(c => c.getDockerClient.pauseContainerCmd(c.getContainerId).exec())
      // unpause Kafka brokers and produce another round
      this.brokerContainers.foreach(c => c.getDockerClient.unpauseContainerCmd(c.getContainerId).exec())
```

* I can disconnect and reconnect the network to broker containers. This will force any active connections to close, but when the network is reattached the host port is different. Testcontainers will always generate a random host port and there doesn't appear to be a way around this. https://www.testcontainers.org/features/networking/ .

i.e.)

```
	  // detach network
      this.brokerContainers
        .foreach(
          c =>
            c.getDockerClient
              .disconnectFromNetworkCmd()
              .withForce(true)
              .withContainerId(c.getContainerId)
              .withNetworkId(c.getNetwork.getId)
              .exec()
        )

      // re-attach network
      this.brokerContainers
        .foreach(
          c =>
            c.getDockerClient
              .connectToNetworkCmd()
              .withContainerId(c.getContainerId)
              .withNetworkId(c.getNetwork.getId)
              .exec()
        )
```

Some other ideas

* Use something like `iptables` or other networking tools to disable connections in the running container itself, but I can't find any of the usual programs in the image to do this (`iptables`, `route`) or even network interface commands. Since `confluentinc/cp-kafka:6.0.0` the image now defaults to a non-root user `appuser`, so I can't install any tools anyway unless I rebase the image.

* Stop and start the broker process within the running container. This should do the trick, but it will require some work to update the container run command to support it.

seglo:
Fixes #1233

Add support to stop and start the Kafka broker programmatically without having to stop or start the container (see #1233 for issues with managing container lifetime).

### DIFF
--- a/testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
+++ b/testkit/src/main/java/akka/kafka/testkit/internal/AlpakkaKafkaContainer.java
@@ -99,11 +99,8 @@ public class AlpakkaKafkaContainer extends GenericContainer<AlpakkaKafkaContaine
 
   public void stopKafka() {
     try {
-      // System.out.println(execInContainer("sh", "-c", "/usr/bin/kafka-server-stop").getStdout());
       ExecResult execResult = execInContainer("sh", "-c", "touch /tmp/stop");
-      System.out.println(execResult.getStdout());
-      System.out.println(execResult.getStderr());
-      System.out.println(execResult.getExitCode());
+      if (execResult.getExitCode() != 0) throw new Exception(execResult.getStderr());
     } catch (Exception ex) {
       throw new RuntimeException(ex);
     }
@@ -111,11 +108,8 @@ public class AlpakkaKafkaContainer extends GenericContainer<AlpakkaKafkaContaine
 
   public void startKafka() {
     try {
-      // System.out.println(execInContainer("sh", "-c", "'" + STARTER_SCRIPT + " &'").getStdout());
       ExecResult execResult = execInContainer("sh", "-c", "touch /tmp/start");
-      System.out.println(execResult.getStdout());
-      System.out.println(execResult.getStderr());
-      System.out.println(execResult.getExitCode());
+      if (execResult.getExitCode() != 0) throw new Exception(execResult.getStderr());
     } catch (Exception ex) {
       throw new RuntimeException(ex);
     }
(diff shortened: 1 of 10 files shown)


ennru:
This looks very good.
I wonder how Testcontainers 1.15 affects this. #1225

seglo:
Thanks. I'm not certain if I'm reproducing `"pick up again when the Kafka server comes back up"` as was intended. In my local tests the consumer won't usually reset to offset 0, but it will disconnect. On travis it will always reset to 0. I think it's because the broker takes longer to start again, when I manually increase the sleep after the broker stop then I can reproduce the problem locally. The change I made to accommodate both scenarios is to always request the full number of messages produced and assert the tail is correct. In both scenarios the Consumer is reconnecting silently, so I _think_ it satisfies the original intention of the test.

seglo:
> I wonder how Testcontainers 1.15 affects this. #1225

Good question. I was thinking about referencing their 1.15 release, but it's not a final yet. Maybe it will be available before our release.

bsideup:
@seglo JFYI the "rc" status of 1.15.0 is only for the image-related changes (deprecated constructors & more) that we wanted to evaluate before doing a Big Bang Release :)
The "core" API is stable (not to mention that we care deeply about the backward compatibility) and it is safe to use now. Also, I am not planning to touch Kafka's internals any time soon, so I'd say you're good to go with rc2 unless you have a super strict & restrictive policy irt dependency versions.

seglo:
@bsideup Thanks for the info. We'll give the RC a test run before our release to confirm our use case is still working as expected.

ennru (approved):
LGTM.

ennru:

PR #1236, opened 2020-10-20
Lock scalafmt version

ennru:
It's not useful to reformat the code as later scalafmt versions would as it hinders comparisons across multiple versions.

Refs #1221


seglo (approved):
LGTM

There will likely be a day we're forced to upgrade for one reason or another, but it should be planned. Each formatting style change should be considered individually.

PR #1244, opened 2020-10-26
Retry commits that failed during a rebalance

altomch:
References #1243

PR #1247, opened 2020-10-29
Retry commit on RebalanceInProgressException during next poll 

### LINKED ISSUE #1243: Commits should be retried on RebalanceInProgressException
Kafka-client's committing behavior was changed since v2.5.0 ([Release notes](https://kafka.apache.org/documentation.html#upgrade_250_notable)):
> When RebalanceProtocol#COOPERATIVE is used, Consumer#poll can still return data while it is in the middle of a rebalance for those partitions still owned by the consumer; in addition Consumer#commitSync now may throw a non-fatal RebalanceInProgressException to notify users of such an event, in order to distinguish from the fatal CommitFailedException and allow users to complete the ongoing rebalance and then reattempt committing offsets for those still-owned partitions.

So `RebalanceInProgressException` ([OffsetCommitCallback](https://github.com/apache/kafka/blob/2.6/clients/src/main/java/org/apache/kafka/clients/consumer/OffsetCommitCallback.java#L41)) should be retried on committing.

### Versions used 
Alpakka-Kafka version: 2.1.0-M1
Akka version: 2.6.10
Kafka-client version: 2.6.0

### Expected Behavior

Commits which failed with `RebalanceInProgressException` should be retried on next poll attempt after rebalance.

### Actual Behavior

`RebalanceInProgressException` propagated to Committer stage

seglo:
Fixes #1243

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -29,6 +29,7 @@ import akka.kafka._
 import akka.kafka.scaladsl.PartitionAssignmentHandler
 import com.github.ghik.silencer.silent
 import org.apache.kafka.clients.consumer._
+import org.apache.kafka.common.errors.RebalanceInProgressException
 import org.apache.kafka.common.{Metric, MetricName, TopicPartition}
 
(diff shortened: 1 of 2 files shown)


seglo:
I'm not sure how I missed #1244

Issue #1228, opened 2020-10-12
`stop-timeout` delays tear-down of consumer if stream itself has failed


jrudolph:
It seems the `consumer.stop-timeout` can be used to drain a consumer of messages that have already been read. (I cannot comment on whether that is (still?) useful in general.)

However, the `stop-timeout` seems to be also applied if the stream consuming / committing the messages failed. In that case, it doesn't seem to make sense to further delay the shutdown of the consumer because it will not be able to drain any further messages in any case.

This was observed in Lagom (https://github.com/lagom/lagom/issues/1786) which has retry logic to reestablish a consumer after it failed but the new instance also fails directly because the old instance is still alive with the same `clientId` so this exception is thrown:

```
 > javax.management.InstanceAlreadyExistsException: kafka.consumer:type=app-info,id=inventory-1
 >     at java.management/com.sun.jmx.mbeanserver.Repository.addMBean(Repository.java:436)
 >     at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerWithRepository(DefaultMBeanServerInterceptor.java:1855)
 >     at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerDynamicMBean(DefaultMBeanServerInterceptor.java:955)
 >     at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerObject(DefaultMBeanServerInterceptor.java:890)
 >     at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.registerMBean(DefaultMBeanServerInterceptor.java:320)
 >     at java.management/com.sun.jmx.mbeanserver.JmxMBeanServer.registerMBean(JmxMBeanServer.java:522)
 >     at org.apache.kafka.common.utils.AppInfoParser.registerAppInfo(AppInfoParser.java:62)
 >     at org.apache.kafka.clients.consumer.KafkaConsumer.<init>(KafkaConsumer.java:797)
 >     at org.apache.kafka.clients.consumer.KafkaConsumer.<init>(KafkaConsumer.java:617)
 >     at akka.kafka.ConsumerSettings$.createKafkaConsumer(ConsumerSettings.scala:178)
 >     at akka.kafka.ConsumerSettings$.$anonfun$apply$3(ConsumerSettings.scala:94)
 >     at akka.kafka.ConsumerSettings.createKafkaConsumer(ConsumerSettings.scala:573)
[...]

PR #1258, opened 2020-11-09
Keep cancellation cause around

ennru:
* Adapt `onDownstreamFinish` with cause (by Akka 2.6)
* Delay consumer actor stop only for non-failure cancellations

References #1228

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
+++ b/core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
@@ -111,7 +111,7 @@ import scala.concurrent.{ExecutionContext, Future}
   setHandler(shape.out, new OutHandler {
     override def onPull(): Unit = pump()
     override def onDownstreamFinish(cause: Throwable): Unit =
-      performShutdown()
+      performShutdown(cause)
   })
 
   override def postStop(): Unit = {
@@ -119,6 +119,6 @@ import scala.concurrent.{ExecutionContext, Future}
     super.postStop()
   }
 
-  def performShutdown(): Unit =
-    log.info("Completing")
+  def performShutdown(cause: Throwable): Unit =
+    log.info("Completing ({})", cause)
 }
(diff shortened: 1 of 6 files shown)


ennru:
This fails consistently, I'll need to find time to investigate it.

ennru:
> Delay consumer actor stop only for non-failure cancellations

As `CommittingSpec` shows, it is relevant to leave the consumer alive so that emergency commits from the `CommitCollectorStage` have a chance to reach the consumer.

Setting `stop-timeout` to 0 makes the test fail in the same way.

patriknw:
Is this PR changing anything else than logging? If so, could you highlight what that it is?

ennru:
> Is this PR changing anything else than logging? If so, could you highlight what that it is?

The original idea didn't work out so easily. The remaining bit is only passing the shutdown cause around via the Akka 2.6 API and logging it.

core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
```diff
@@ -111,14 +111,14 @@ import scala.concurrent.{ExecutionContext, Future}
   setHandler(shape.out, new OutHandler {
     override def onPull(): Unit = pump()
     override def onDownstreamFinish(cause: Throwable): Unit =
-      performShutdown()
+      performShutdown(cause)
```
patriknw, line 114: We are not calling super.onDownstreamFinish by default?
I don't know if that is important? Could there be problems when just swallowing the signal as we do by default here?

ennru: Good find. That should have been there.

core/src/main/scala/akka/kafka/internal/BaseSingleSourceLogic.scala
```diff
@@ -108,17 +108,22 @@ import scala.concurrent.{ExecutionContext, Future}
...
-    override def onPull(): Unit = pump()
-    override def onDownstreamFinish(cause: Throwable): Unit =
-      performShutdown()
-  })
+  setHandler(
+    shape.out,
+    new OutHandler {
+      override def onPull(): Unit = pump()
+      override def onDownstreamFinish(cause: Throwable): Unit = {
+        super.onDownstreamFinish(cause)
```
patriknw, line 116: This will by default `thisStage.cancelStage`. Will that work well together with subclass implementations such as in `SingleSourceLogic`, which is using `setKeepGoing(true)` and `complete(shape.out)`?

ennru: I'm not sure. Maybe nothing we want to touch without trying it harder.

ennru:
The improvements in this PR are too small for this to be relevant.

Issue #1242, opened 2020-10-25
At lease once guarantees of Producer.committableSink 


PrathikPuthran:
Producer.committableSink seems to drop messages when the consumer is shutting down. It seems the offset commits are out of order.  This could be a problem when the consumer is shutting down and partitions are rebalanced to other consumer because higher offsets are committed to kafka even though the messages corresponding to lower offsets are not successfully published to the target kafka topic. 
I found this while investigating lost messages in the stream and all the lost messages were during the consumer rebalance ( usually triggered by consumer process restart)

ennru:
Thank you for reporting what you experienced.

What versions are you using?
Can you explain your findings more exact?
Only offsets of successfully produced records are added to be committed. Did you inspect records being produced out of order?

PrathikPuthran:
Thanks @ennru for looking into it. 
I was using the below version 
compile group: 'com.typesafe.akka', name: 'akka-stream-kafka_2.12', version: '2.0.1' 

Yes the kafkaproducer can produce records out of order as described in the doc https://kafka.apache.org/documentation/#retries 
We have 10 consumers subscribing to a kafka topic. It does some mutation to the data and sends them to target kafka topic. This system processes around 15-20K events/second. I was investigating missing messages where some messages in source topic was not available in target topic. On further investigation the lost messages were during consumer restarts. After digging into the implementation of Producer.committableSink() it looks like it commits offsets as soon as it gets acks without accounting for the ordering. So after partition re-assignment the new consumer would start consuming from the higher committed offset even though the messages from the lower offsets were not successfully produced to kafka.

I changed the implementation to use combination of Producer.flexiFlow() and Consumer.committableSink. 
Producer.flexiFlow guarantees ordering because it uses mapAsync. I haven't seen missing messages issue since then.

ennru:
Yes, you are right. `committableSink` will allow committing of offsets out of order if the producer callbacks finish in a different order than the `send` was issued.

silles79:
I think i am having the same bug but stream looks like

`Consumer.committableSource(consumerSettings, Subscriptions.topicPattern(config.kafkaConsumer.topicsIn))
      .map(process)
      .via(Producer.flexiFlow(producerSettings))
      .map(_.passThrough)
      .toMat(Committer.sink(CommitterSettings(system).withMaxBatch(20)))(DrainingControl.apply)
      .run()
`

def process  is simple and just passes back the payload.committableOffset

[...]

PR #1265, opened 2020-11-25
Ensure write and order guarantees with Producer.committableSink

seglo:
References #1242 

Enforce `enable.idempotence` when using the `Producer.committableSink`. This will ensure that produced messages are recorded once and in the order they were sent to the broker.  Some caveats with this solution:

* The broker must be version 0.11 or greater
* `max.in.flight.connections` must be `<= 5` (enforced by setting `enable.idempotence=true`)
* `acks` must be set to `all` (enforced by setting `enable.idempotence=true`)

### DIFF
--- a/benchmarks/src/main/scala/akka/kafka/benchmarks/AlpakkaCommittableSinkFixtures.scala
+++ b/benchmarks/src/main/scala/akka/kafka/benchmarks/AlpakkaCommittableSinkFixtures.scala
@@ -18,7 +18,7 @@ import akka.stream.scaladsl.{Keep, Sink, Source}
 import com.codahale.metrics.Meter
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.kafka.clients.consumer.ConsumerConfig
-import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.clients.producer.{ProducerConfig, ProducerRecord}
 import org.apache.kafka.common.serialization.{
   ByteArrayDeserializer,
   ByteArraySerializer,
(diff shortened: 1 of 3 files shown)


seglo:
Initial benchmarks indicate enforcing `enable.idempotence` with `Producer.committableSink` has little to no effect on producer throughput. It could be that other Alpakka Kafka performance penalties outweigh any value provided by this setting, or it could be that our benchmark infrastructure scale is just not sufficient to observe a notable difference.  Either way, if this fixes the dropped/unordered messages it may be an acceptable solution when using `Producer.committableSink`.

seglo:
@PrathikPuthran The intention is that using `enable.idempotence` could resolve the issue you raised in #1242.  Can you try this PR, or just add `enable.idempotence` to your `ProducerSettings` yourself?

seglo:
@PrathikPuthran Hello again. Let me know if you have time to check if this setting resolves the issue for you. If it does then we can make this the default for 2.1.0.

PrathikPuthran:
@seglo Sorry for the delay. Would this work if custom partitioning is used in the kafka producer?

seglo:
> @seglo Sorry for the delay. Would this work if custom partitioning is used in the kafka producer?

@PrathikPuthran Yes, I don't see why not. All it's doing is using sequence numbers (per destination partition for the length of the producer session) to ensure message ordering and for the broker to determine if any messages have been missed or duplicated.  Missed messages will result in a fatal exception, dupes will be de-duped silently on the broker side. Your broker must be at least version `0.11`.

PrathikPuthran:
Yes but if send to one broker partition fails (if one broker is unhealthy) but the send for some other broker partition succeeds with messages of higher offsets you would still be committing offsets out of order I think?

seglo:
> Yes but if send to one broker partition fails (if one broker is unhealthy) but the send for some other broker partition succeeds with messages of higher offsets you would still be committing offsets out of order I think?

I did some digging and I can't find an official statement that callbacks can return out of order with `enable.idempotence=true`, but I think it's implied that this should not happen. If you can reliably reproduce this commit out of order issue I would encourage you to try it anyway. Otherwise it may take awhile for me to verify it myself.

seglo:
This may be better served as a note in the docs on `Producer.committableSink`, instead of enforcing a Producer property.

seglo:
If someone can confirm if `enable.idempotence` resolves the issue I'll refresh this PR with a docs update.  Closing for now.

PR #1267, opened 2020-11-30
Run link validator and release drafter (as GitHub workflows)

ennru:


.github/workflows/link-validator.yml
```diff
@@ -0,0 +1,34 @@
+name: Link Validator
+
+on:
+  pull_request_target:
+    branches: [$default-branch]
+  schedule:
+    - cron:  '0 6 * * 1'
```
seglo, line 7: How come not on a PR?

seglo (commented):
Interesting. I like the approach of slowly transitioning to actions.

.github/workflows/link-validator.yml
ennru: `pull_request_target` means "run on pull requests targeting the listed branches".

seglo: Got it.  Is it also running on a cron?

ennru: Yes, every gardening Monday.

seglo (approved):

PR #1268, opened 2020-11-30
Maximum number of transient failures for Transactions integration tests

seglo:
Add a maximum restart condition to transactions integration tests.

It's possible to get into a livelock situation where the `restartAfter` interval causes transactions to abort continuously until a timeout.  This can happen when there are a few partitions left to process and they can never be fully processed because we always restart the stream before the transaction can be completed successfully. The `maxRestarts` provides an upper bound for the maximum number of times we restart the stream so if we get into a livelock it can eventually be resolved by not restarting any more.

Originally implemented in #1109

### DIFF
--- a/tests/src/test/scala/akka/kafka/TransactionsOps.scala
+++ b/tests/src/test/scala/akka/kafka/TransactionsOps.scala
@@ -4,6 +4,8 @@
  */
 
 package akka.kafka
+import java.util.concurrent.atomic.AtomicInteger
+
 import akka.{Done, NotUsed}
 import akka.actor.ActorSystem
 import akka.kafka.ConsumerMessage.PartitionOffset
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/akka/kafka/TransactionsOps.scala
```diff
@@ -163,7 +175,6 @@ trait TransactionsOps extends TestSuite with Matchers {
       .plainSource(settings, Subscriptions.topics(topic))
       .map(r => (r.partition(), r.offset(), r.value()))
       .take(elementsToTake)
-      .idleTimeout(30.seconds)
```
seglo, line 166: Prevent prematurely timing out the stream when consumer is idle.  Give `maxRestarts` a chance to elapse and then all remaining elements will be consumed.

ennru (approved):
LGTM.

PR #1287, opened 2020-12-14
Update default CP version and make docker image/tag configurable

seglo:
`TestcontainersConfiguration.getInstance().getKafkaImage()` is deprecated and returns an image and tag (instead of just an image). I'm not sure when this changed, but we expect only an image name so we can append an image tag ourself.  The result is that `latest` (the default tag) is always used Kafka testcontainers 1.15.0+.

Confluent is now unpublishing older minor releases of docker images, so we need to update our default image to `6.0.1`.

### DIFF
--- a/testkit/src/main/resources/reference.conf
+++ b/testkit/src/main/resources/reference.conf
@@ -15,12 +15,16 @@ akka.kafka.testkit {
 # // #testkit-testcontainers-settings
 akka.kafka.testkit.testcontainers {
 
-  # define these settings to select a different Kafka/ZooKeeper docker image and version of Confluent Platform
+  # define these settings to select a different Kafka/ZooKeeper docker image
+  # we recommend using Confluent Platform docker images and using the same version across all images
   # Confluent publishes images on DockerHub: https://hub.docker.com/r/confluentinc/cp-kafka/tags
   # Kafka versions in Confluent Platform: https://docs.confluent.io/current/installation/versions-interoperability.html
-  confluent-platform-zookeeper-image = "confluentinc/cp-zookeeper"
-  confluent-platform-kafka-image = "confluentinc/cp-kafka"
-  confluent-platform-schema-registry-image = "confluentinc/cp-schema-registry"
+  zookeeper-image = "confluentinc/cp-zookeeper"
+  zookeeper-image-tag = ${akka.kafka.testkit.testcontainers.confluent-platform-version}
+  kafka-image = "confluentinc/cp-kafka"
+  kafka-image-tag = ${akka.kafka.testkit.testcontainers.confluent-platform-version}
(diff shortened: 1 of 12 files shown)


testkit/src/main/resources/reference.conf
```diff
@@ -18,7 +18,7 @@ akka.kafka.testkit.testcontainers {
   # define this to select a different Kafka version by choosing the desired version of Confluent Platform
   # available Docker images: https://hub.docker.com/r/confluentinc/cp-kafka/tags
   # Kafka versions in Confluent Platform: https://docs.confluent.io/current/installation/versions-interoperability.html
-  confluent-platform-version = "6.0.0"
+  confluent-platform-version = "6.0.1"
```
ennru, line 21: Should we put the image name into config as well?

seglo: Good idea.

testkit/src/main/resources/reference.conf
```diff
@@ -15,10 +15,13 @@ akka.kafka.testkit {
...
-  # define this to select a different Kafka version by choosing the desired version of Confluent Platform
-  # available Docker images: https://hub.docker.com/r/confluentinc/cp-kafka/tags
+  # define these settings to select a different Kafka/ZooKeeper docker image and version of Confluent Platform
+  # Confluent publishes images on DockerHub: https://hub.docker.com/r/confluentinc/cp-kafka/tags
   # Kafka versions in Confluent Platform: https://docs.confluent.io/current/installation/versions-interoperability.html
-  confluent-platform-version = "6.0.0"
+  confluent-platform-zookeeper-image = "confluentinc/cp-zookeeper"
+  confluent-platform-kafka-image = "confluentinc/cp-kafka"
+  confluent-platform-schema-registry-image = "confluentinc/cp-schema-registry"
+  confluent-platform-version = "6.0.1"
```
ennru, line 24:
```suggestion
  zookeeper-image = "confluentinc/cp-zookeeper"
  zookeeper-version = ${akka.kafka.testkit.testcontainers.confluent-platform-version}
  kafka-image = "confluentinc/cp-kafka"
  kafa-version = ${akka.kafka.testkit.testcontainers.confluent-platform-version}
  kafka-with-schema-registry-image = "confluentinc/cp-schema-registry"
  kafka-with-schema-registry-version = ${akka.kafka.testkit.testcontainers.confluent-platform-version}
  confluent-platform-version = "6.0.1"
```

Would it be beneficial to make this less dependent on the Confluent Platform packaging by individual versions?

seglo: Good idea.

ennru (approved):
Nice!

seglo:
Unrelated integration test failure.

ennru:

PR #1288, opened 2020-12-14
Remove parameterless methods

seglo:
I noticed when upgrading from 2.0.5 to 2.1.0-M1 that if you reference `consumerDefaults` or `producerDefaults` with parens you will get a compile time error because they were dropped.  Should we keep the params for parameterless methods for 2.1.0?

### DIFF
--- a/testkit/src/main/scala/akka/kafka/testkit/internal/KafkaTestKit.scala
+++ b/testkit/src/main/scala/akka/kafka/testkit/internal/KafkaTestKit.scala
@@ -34,13 +34,13 @@ trait KafkaTestKit {
   val StringSerializer = new StringSerializer
   val StringDeserializer = new StringDeserializer
 
-  def producerDefaults: ProducerSettings[String, String] = producerDefaults(StringSerializer, StringSerializer)
+  def producerDefaults(): ProducerSettings[String, String] = producerDefaults(StringSerializer, StringSerializer)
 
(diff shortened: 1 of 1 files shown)


ennru (approved):
LGTM.

ennru (changes requested):
Hm, these are effectively `val`s, so they should stay without parens.

seglo:
That's true. My point is it will cause breaking changes for users who reference them as methods (with params). I'm on the fence if we should change it or not. It could be an annoyance for many who rely on the testkit.

ennru:
This problem shows with Scala 2.13.4 which is stricter about nullary methods and parens, right?

If we add the parens, they need to be added in the specs and examples that extend `KafkaTestKit` as well.

ennru:
Tried it on Alpakka Kafka with Scala 2.13.4 now, out own use of these values makes me believe we should not add parens.

`100 warnings found`...

ennru:
The parens appeared on `consumerSettings` in #1099 as part of 2.0.3 
https://github.com/akka/alpakka-kafka/pull/1099/files#diff-58a836c24b91550202dfa15b5a0aa7b66e4a16a503fb7e50039d1d6876d6d80eR43

So it was an oversight to add them there.

seglo:
Got it.  Closing this.

PR #1292, opened 2020-12-21
TestKit: make consumerDefaults parameterless again

ennru:
By mistake the `def consumerDefaults` in `KafkaTestKit` got parenthesis added in #1099 which got released in Alpakka Kafka 2.0.3.
As Scala 2.13.4 started to warn about the inconsistent use of parens, we should go back to not having the parenthesis on this method.

References https://github.com/akka/alpakka-kafka/pull/1288#issuecomment-749015437


johanandren (approved):

raboof (approved):
now I want a blue MCPA hat

PR #1289, opened 2020-12-14
Migrate build to GitHub actions

seglo:
TODO
- [x] Check, Build, Test
- [x] Whitesource (using `release` environment for secrets, can put `publish` secrets there too, see whitesource workflow for reference)
- [ ] Publish
- [x] Dependency cache
- [x] Print log on failure

PR #1298, opened 2020-12-23
Login to dockerhub account in build

seglo:


seglo:
As I went about updating a build I quickly rediscovered that [secrets are not made available to PR builds for branches on forks](https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions) (for good reason).  If you can restrict your build steps that require image pulls to only run on branches of the upstream repository, then you can use a dockerhub account, but for other builds that make extensive use of public Docker images in their standard CI tests used in PRs (such as Alpakka builds), then another solution is required.

After a little more digging I found a [Travis thread](https://travis-ci.community/t/docker-hub-you-have-reached-your-pull-rate-limit/10517) about this same issue, but with no official resolution at this time (though someone suggested using a registry mirror like https://mirror.gcr.io).  However, the OP referenced an issue on the [same subject for GitHub actions](https://github.com/actions/virtual-environments/issues/1445).  [GitHub actions has reached an agreement with dockerhub](https://github.com/actions/virtual-environments/issues/1445#issuecomment-729675520) to allow unlimited pulls for any builds on their infrastructure.  I would expect travis might work on a similar deal, but since we were already working on porting Alpakka builds to GitHub actions (#1289) we will gain this benefit as a side effect.

Issue #1286, opened 2020-12-14
Reset defense for intermittently 'lost' offsets with low-watermark


jyates:
### Short description

There are a number of JIRAs where Kafka Brokers can 'mess up' and claim that an offset is out-of-range, often around when a log segment is rolled (see below). This causes the consumer to reset offsets based on its configuration, either latest, earliest or none. Latest will skip data (but that is true of any use, so out of scope here). Earliest and none however, can have very bad impacts on stream processing, likely waking up folks to fix the issue by force-setting the offsets to the right place

We should detect when there is this offset drop/reset and handle it more gracefully.

### Details

`<pitch>`
This issue seems to crop up quite frequently, with different root causes across versions. Often users do not have as easy of a path to upgrade their brokers (unwilling/able to maintain a fork, externally provided cluster, etc) while upgrading the client is more in their control. This also provides an opportunity for alpakka-kafka to provide a superior, delightful experience when compared to other clients.

An 'earliest' rewind when offsets are dropped can cause days or weeks of data to start being consumed again. This can swamp downstream systems and likely will page as the lag will go from minutes back to that start as the consumer begins committing at that earliest offset.

Similarly, a 'none' rewind approach will hang the consumer, eventually paging someone. This is better for not disturbing systems, but is likely rarely used as makes it harder to deploy new consumers using defaults. While we can't do anything for none (that is handled at a lower level in the kafka client) we can help the 'reset to earliest' case.
`</pitch>`

How might something like this work? We already filter out offsets from partitions that are no longer assigned in the SourceBuffer stage. However, that does not take into account the latest committed offsets. We track the latest committed offsets when doing the periodic offset commits. That maintains a map and doesn't seem to add too much overhead (do not have numbers on hand, but updating a couple maps _seems_ not so bad compared to reading/writing lots of data across network).

It seems like we can combine these two components to identify when there is a rewind and then trigger a seek on the consumer to back to the latest committed location, the low-watermark of the partition. 

The obvious concern here is about when someone intentionally seeks a consumer. In that case, the consumer needs to be part of the latest epoch, so either its running in the current set of consumers or its triggered externally.

For the latter, the current consumer group can no longer be part of the latest epoch, so it will be subject to a rebalance. Thus, we would drop the partition 'low watermark' in favor of the next commit (this works because a rebalance will seek to the latest committed offset anyways, so that is safe; if we have the lost offset problem there, not much we can safely do).

[...]

PR #1299, opened 2020-12-28
Add configurable protection against server-bug induced resets

jyates:
Includes some refactoring to pull out the common commit progress tracking
that is useful for both reset tracking and commit refreshing. Did not
want to keep state twice, so you get this little bit of twisted design
with the progress tracker, but keeps the logic contained.

Continues to keep the same amount of overhead (that is, very little)
for when commit refreshing and reset tracking are not enabled. At the
same time, avoids double work when both are enabled

References #1286


jyates:
can't see where/how my changes [broke the docs](https://travis-ci.com/github/akka/alpakka-kafka/jobs/465633994), but does look like we would need a minor version bump for the change in the `ConsumerSettings` to support the new option.

jyates:
Not sure at all how verifyDocs is passing in master, given my code doesn't touch those lines at all. That said, looks like an occurance of https://github.com/scala/scala-collection-compat/issues/203

jyates:
Not sure what is up with the tests now, but running locally...seems to work fine
```
$ sbt "tests/testOnly *.RebalanceExtSpec"
[info] Loading global plugins from /Users/jyates/.sbt/1.0/plugins
[info] Loading settings for project alpakka-kafka-build from plugins.sbt ...
[info] Loading project definition from /Users/jyates/dev/oss/alpakka-kafka/project
[info] Loading settings for project alpakka-kafka from build.sbt ...
[info] 
[info] ** Welcome to the Alpakka Kafka connector! **
...
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Formatting 2 Scala sources...
[info] Compiling 3 Scala sources to /Users/jyates/dev/oss/alpakka-kafka/core/target/scala-2.13/classes ...
[info] Compiling 1 Java source to /Users/jyates/dev/oss/alpakka-kafka/testkit/target/scala-2.13/classes ...
[info] Compiling 55 Scala sources and 17 Java sources to /Users/jyates/dev/oss/alpakka-kafka/tests/target/scala-2.13/test-classes ...
[info] RebalanceExtSpec:
[info] Fetched records
[info] - must no messages should be lost when two consumers consume from one topic and two partitions and one consumer aborts mid-stream (12 seconds, 367 milliseconds)
[info] ScalaTest
[info] Run completed in 1 minute, 52 seconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```

Same for `$ sbt "++2.12.11 tests/testOnly *.RebalanceExtSpec"` and `sbt "++2.13.2 tests/testOnly *.RebalanceExtSpec"`

Running on jdk8
```
$ java -version
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
```

jyates:
Looks like the tests are getting stuck just spinning up the KafkaContainerCluster, not sure how to go about managing that. Maybe busy build infra at the time?

jyates:
Any thoughts @ennru?

ennru:
Looks like something in TestContainers didn't work correctly. I need to look closer...

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
```diff
@@ -0,0 +1,103 @@
...
+                  // requested and committed are assumed to be kept in-sync, so this _should_ be safe. Fails
+                  // catastrophically if this is not the case
+                  val committed = progress.committedOffsets(tp)
+                  log.warning(
+                    s"Dropping offsets for partition $tp - received an offset which is less than "
+                    + s"allowed $threshold from the last requested offset "
+                    + s"(threshold: $allowedMaxDeltaFromRequest). "
+                    + s"Seeking to the latest known safe (committed or assigned) offset: $committed"
+                  )
+                  consumer ! Seek(Map(tp -> committed.offset()))
```
ennru, line 66: This is the central piece. I think it deserves a method.

Would it be better to pass the actor reference to `protect` instead of the instance value? (Comes a bit as a surprise that it sends messages back to the actor that called it.)

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -96,61 +95,39 @@ import scala.util.control.NonFatal
...
-                            positionTimeout: java.time.Duration): Unit = {}
+    private final class Noop() extends CommitRefreshing {
+      override def updateRefreshDeadlines(tps: Set[TopicPartition]): Unit = {}
+
+      override val refreshOffsets: Map[TopicPartition, OffsetAndMetadata] = Map()
     }
+    private final class Impl(commitRefreshInterval: FiniteDuration, progress: ConsumerProgressTracking)
+        extends CommitRefreshing
+        with ConsumerProgressTracking {
+      progress.addProgressTrackingCallback(this)
```
ennru, line 117: This delegation gives me a hard time.

core/src/main/scala/akka/kafka/internal/ConsumerProgressTracking.scala
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Track the current state of the consumer: what offsets its has requested and committed, filtering by the current
+ * assignments to the consumer. When a partition is assigned the consumer for the first time, its assigned offset
+ * is the current position of the partition (uses underlying Kafka Consumer to leverage the configured offset-reset
+ * policy).
+ */
+trait ConsumerProgressTracking {
+  var committedOffsets: Map[TopicPartition, OffsetAndMetadata] = _
+  var requestedOffsets: Map[TopicPartition, OffsetAndMetadata] = _
```
ennru, line 20: Should these be `def`s and only the implementation have access to change the values?

ennru (commented):
Thank you for this intricate work. I wonder how we can document it so that users that might need it can find and understand it.

The hierarchy of `ConsumerProgressTracking` is a bit challenging with the delegation.
I tried out some suggestions locally, let me know what you think.

The CI had a hick-up when you pushed, I restarted it and it failed in more expected ways.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -463,6 +429,15 @@ import scala.util.control.NonFatal
     }
   }
 
+  private def ensureProgressTracker(): ConsumerProgressTracking = {
+    // we need to not only set the progress tracker, if it isn't set yet, as well as return it to the caller
+    progressTracker = Some(progressTracker match {
+      case Some(tracker) => tracker
+      case None => new ConsumerProgressTrackerImpl()
+    })
+    progressTracker.get
```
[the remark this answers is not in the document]
jyates: good point.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
jyates: My preference was to the looser coupling, rather than having some sort of wonky inheritance thing. Open to suggestions though!

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
jyates: I went back on forth on that API - didn't have a good feel on style for the actor refs, can do!

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
jyates: The constraints I came up with for the tracking were:
 * only a single object should track the requested/committed/assigned state for the consumer
 * other objects can query that state, but should not be making updates (outside of calls by the ConsumerActor)
 * commit refreshing needs to know about the changes in consumer state so it can update its internal state.

That's how i ended up with a parent tracker that calls listeners for calls to those update methods. Wanted to avoid another type in the hierarchy for the base (which listeners would implement) and then a separate one for the tracker that lets objects access the consumer progress state.

But yeah, its a bit weird that CommitRefreshing$Impl has committedOffsets/requestedOffsets. My expectation was that it is hidden away by the interface, so its not actually that big of a deal.

jyates:
Thank you for taking a look @ennru!

> I wonder how we can document it so that users that might need it can find and understand it.

Hopefully we can use the description in the original ticket as a basis for explanation and then the fact that its a simple field in the config for "try to fix it if it goes over this value" should be easy enough to grok. But something we can definitely run by folks with a "blind" understanding...at the same time, its a pretty "deep" issue and i have a hard time figuring out if we can create a sane "always on" default that will work for all setups.

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
```diff
@@ -0,0 +1,135 @@
...
+        case None =>
+          // it's a partition that we have no information on, so assume it's safe and continue because it's likely
+          // due to a rebalance, in which we have already reset to the committed offset, which is safe
+          Some((tp, partitionRecords))
+      }
+    }
+
+    /**
+     * We found that we have previously requested the [[OffsetAndMetadata]] for the [[TopicPartition]], check it
+     * against the partition records top ensure that we haven't exceeded the threshold.  If the records are within the
```
ennru, line 75:
```suggestion
     * against the partition records to ensure that we haven't exceeded the threshold.  If the records are within the
```

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
```diff
@@ -0,0 +1,135 @@
...
+     * threshold, we just return the given records. If there are records found outside the threshold:
+     * 1. Drop the records returned for the partition (they are outside the threshold) by returning None)
```
ennru, line 77:
```suggestion
     * 1. Drop the records returned for the partition (they are outside the threshold) by returning `None`
```

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
```diff
@@ -0,0 +1,135 @@
...
+     * 2. Request a seek to the latest committed offset of the partition (known to be "safe").
+     * 3. New records for the partition arrive with the next poll, which should be within the threshold.
+     *
+     */
+    private[internal] def protectPartition[K, V](
```
ennru, line 82:
```suggestion
    private def protectPartition[K, V](
```

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Check to see if the records can be safely passed to the consumer (no reset found). If not, then tells the
+     * consumer to seek back to the last safe location (last committed offset) and try pulling again. Note that if
+     * the consumer commits less frequently than the configured protection interval, then they will be constantly
+     * seeking back to the last committed offset.
+     * @return [[Some]] records if they can be safely passed to the consumer, returns [[None]] otherwise.
+     */
+    private[internal] def maybeProtectRecords[K, V](
```
ennru, line 58:
```suggestion
    private def maybeProtectRecords[K, V](
```

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
ennru: As we implement two interfaces with similar things it becomes blurry which method is here for wich reason.

Maybe an anonymous class make that more clear?
```scala
      progress.addProgressTrackingCallback(new ConsumerProgressTracking {
        // required by the progress tracking trait, but not actually used or exposed as this type should only be
        //accessed by its CommitRefreshing interface
        override def committedOffsets: Map[TopicPartition, OffsetAndMetadata] = null
        override def requestedOffsets: Map[TopicPartition, OffsetAndMetadata] = null

        override def revoke(revokedTps: Set[TopicPartition]): Unit = {
          refreshDeadlines = refreshDeadlines -- revokedTps
        }

        override def assignedPositions(assignedTps: Set[TopicPartition],
                                       assignedOffsets: Map[TopicPartition, Long]): Unit = {
          // assigned the partitions, so update all the of deadlines
          refreshDeadlines = refreshDeadlines ++ assignedTps.map(_ -> commitRefreshInterval.fromNow)
        }
      })
```

ennru (commented):
Thanks for the updates.

> Hopefully we can use the description in the original ticket as a basis for explanation and then the fact that its a simple field in the config for "try to fix it if it goes over this value" should be easy enough to grok.

Please add a section to the docs (maybe in "Error handling"?), the longer log warnings could refer to the docs section.

jyates:
Appologies for the delay, got busy at the office. Latest version is squashed of last changes addressing comments, onto latest master.

It also includes support for using timestamp deltas as well as offsets (the former being much easier to understand); did require adding another hook for tracking 'received' message high-water marks, but fits in the same model.

jyates:
looks like flaky build system for many of the test failures - failure to setup clusters and such.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
seglo: I'm having similar trouble understanding this type hierarchy.  I think it would make sense to have a separate type for the listener that is implemented outside this source file and seal the rest of traits.

jyates: Which is sounds similar to what I have in the latest. But im the first to admit the weakness in my scala-fu.

seglo (commented):
Looking good @jyates. I find the type hierarchy a little confusing, maybe we can separate that out a little. More types might be more readable in this case.

Apologies for all the nits. There's a lot of code!

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
seglo: What I meant was in this type:

```
trait ConsumerProgressTrackingListener {
  def requested(offsets: Map[TopicPartition, OffsetAndMetadata]): Unit = {}
  def received[K, V](records: ConsumerRecords[K, V]): Unit = {}
  def committed(offsets: java.util.Map[TopicPartition, OffsetAndMetadata]): Unit = {}
  def revoke(revokedTps: Set[TopicPartition]): Unit = {}
  def assignedPositions(assignedTps: Set[TopicPartition], assignedOffsets: Map[TopicPartition, Long]): Unit = {}
}
```

We have default impls for the first 4 methods, but they're only ever implemented by `ConsumerProgressTracking`. Do they belong there instead?

seglo:
I'd like to include this in the `2.1.0-RC1` milestone.  I don't have a date, but I'm aiming for some time this month.

BTW, if you rebase on master you'll pull in changes that should reduce the number of client timeouts in the build.

jyates:
@seglo thanks for taking a look! Been swamped with work, but planning to take some time this week to address your comments.

seglo:
Thanks @jyates.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
jyates: My thought was that if we are going to have a listener for progress, it should be able to listen to all the progress events, if it wants. Felt "lame" to only allow listening for a subset of events. That said, unless you need it YAGNI... i'll drop them. Easy enough to add later :)

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
```diff
@@ -0,0 +1,173 @@
...
+        log.warning(
+          s"""
+             |Dropping offsets for partition $tp - received an offset which is less than allowed $threshold from the
+             |last requested offset (threshold: $threshold). Seeking to the latest known safe
+             |(committed or assigned) offset: $committed. See
+             |https://doc.akka.io/docs/alpakka-kafka/current/errorhandling.html#unexpected-consumer-offset-reset
+             |for more information.
+             |""".stripMargin
+        )
+        consumer ! Seek(Map(tp -> committed.offset()))
```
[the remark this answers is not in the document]
jyates: I think the actor is still handling "this" request, so all the seeks back up in the actor queue. If we get another poll we would get the next batch of records from the partition start. Either its still very old or its within the range. Former would seek again and we make progress on the next poll. Latter allows us to make progress.

Worst case is that we re-process a couple polls of data, but that should be expected to happen.

Happy to add the batched seek logic, wanted to keep this initial one small while everyone gets their heads around it - know it took me a while :)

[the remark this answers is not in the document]
seglo: > I think the actor is still handling "this" request, so all the seeks back up in the actor queue

I think it's _possible_ a timer tick message (Poll) could be interleaved with these Seek messages, but it's not very likely.  

Your explanation that it should all work out in the end anyway SGTM, so I'm fine keeping it as-is :+1:

seglo:
Thanks for the updates @jyates. I think you may have missed some of my review comments. GitHub was kind enough to hide some of them so you'll need to expand the hidden review comments to see them (thanks GitHub! :)

One of the CI checks failed too.

https://github.com/akka/alpakka-kafka/pull/1299/checks?check_run_id=1883909741#step:6:98

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
```diff
@@ -0,0 +1,173 @@
...
+import akka.actor.ActorRef
+import akka.event.LoggingAdapter
+import akka.kafka.ResetProtectionSettings
+import akka.kafka.internal.KafkaConsumerActor.Internal.Seek
+import org.apache.kafka.clients.consumer.{ConsumerRecord, ConsumerRecords, OffsetAndMetadata, OffsetAndTimestamp}
+import org.apache.kafka.common.TopicPartition
+
+import scala.jdk.CollectionConverters._
+
+trait ConsumerResetProtection {
```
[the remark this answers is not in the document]
seglo: I think this one got missed.

core/src/main/resources/reference.conf
```diff
@@ -143,6 +143,14 @@ akka.kafka.consumer {
     backoff-factor = 2.0
   }
 
+  # If we get an offset back from Kafka that is greater than this number of offsets or amount back in time before the
+  # requested offset for the partition, we identify it as a broker-side error and attempt to reset the consumer to the
+  # latest committed offset.
+  reset-protection {
```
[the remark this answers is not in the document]
jyates: i like `consumer-reset-protection` as the latest version also considers timestamps.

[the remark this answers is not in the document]
jyates: That said, setting this config alone would be `akka.kafka.consumer.consumer-reset-protection` which feels weird. WDYT about keeping the name as is?

core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -266,9 +268,54 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     val enrichAsync: Option[ConsumerSettings[K, V] => Future[ConsumerSettings[K, V]]],
     val consumerFactory: ConsumerSettings[K, V] => Consumer[K, V],
     val connectionCheckerSettings: ConnectionCheckerSettings,
-    val partitionHandlerWarning: FiniteDuration
+    val partitionHandlerWarning: FiniteDuration,
+    val resetProtectionSettings: ResetProtectionSettings
 ) {
 
+  def this(
```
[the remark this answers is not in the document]
jyates: It breaks the CI otherwise with
```
[error] akka-stream-kafka: Failed binary compatibility check against com.typesafe.akka:akka-stream-kafka_2.13:2.1.0-M1! Found 1 potential problems (filtered 24)
[error]  * method this(scala.collection.immutable.Map,scala.Option,scala.Option,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.Duration,java.lang.String,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.concurrent.duration.FiniteDuration,scala.Option,scala.Function1,akka.kafka.ConnectionCheckerSettings,scala.concurrent.duration.FiniteDuration)Unit in class akka.kafka.ConsumerSettings does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.kafka.ConsumerSettings.this")
[error] java.lang.RuntimeException: Failed binary compatibility check against com.typesafe.akka:akka-stream-kafka_2.13:2.1.0-M1! Found 1 potential problems (filtered 24)
[error] 	at scala.sys.package$.error(package.scala:30)
```

core/src/main/scala/akka/kafka/ResetProtectionSettings.scala
```diff
@@ -0,0 +1,73 @@
...
+
+package akka.kafka
+import java.time.{Duration => JDuration}
+
+import akka.util.JavaDurationConverters._
+import com.typesafe.config.Config
+
+import scala.concurrent.duration._
+
+class ResetProtectionSettings private[kafka] (val enable: Boolean,
```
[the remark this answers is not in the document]
jyates: I expected users to use this API to configure it by hand, similar to the `ConnectionCheckerSettings`. Is that one just not tagged correctly @seglo ?

docs/src/main/paradox/errorhandling.md
```diff
@@ -35,3 +35,63 @@ If reading from Kafka failure is caused by other reasons, like **deserialization
 consuming raw byte arrays and deserializing in a subsequent `map` stage where you can use supervision to skip failed elements. See also @ref:[Serialization](serialization.md) and @ref:["At least once"](atleastonce.md) pages for more suggestions.
 
 @@@
+
+## Unexpected consumer offset reset
+
+Sometimes, due to various Kafka server bugs ([KAFKA-7414](https://issues.apache.org/jira/browse/KAFKA-7414),
+[KAFKA-7447](https://issues.apache.org/jira/browse/KAFKA-7447),
+[KAFKA-8896](https://issues.apache.org/jira/browse/KAFKA-8896),
+[KAFKA-8764](https://issues.apache.org/jira/browse/KAFKA-8746),
```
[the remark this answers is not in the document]
jyates: Hmm, doesn't look right to me...ill just remove it. lots of other examples already :)

docs/src/main/paradox/errorhandling.md
```diff
@@ -35,3 +35,63 @@ If reading from Kafka failure is caused by other reasons, like **deserialization
...
+impact on consumer performance, but as always, be sure to benchmark for your use case before enabling. 
+
+### Setting offset threshold appropriately
+
+If you set the offset threshold to less than the the frequency at which you commit, any reset protection that takes
+place will likely cause you reprocess more data than necessary. For instance, assume you commit every 1000 records, but
+have `reset-protection.offset-threshold` set to 500 records. A reset could then cause you to-reprocess up to 999
+records in the worst case. 
+
+You should set the reset protection to the number of offsets near the topic's configured. Alternatively, you can
```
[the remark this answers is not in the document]
jyates: its like the hook the nerdiest rap song...fixed :)

tests/src/test/scala/akka/kafka/internal/ConsumerProgressTrackingSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+    tracker.requested(Map(tp -> offset(0)))
+    tracker.received(records)
+
+    tracker.requestedOffsets(tp).offset() should be(0)
+    tracker.receivedMessages should have size 1
+    tracker.receivedMessages(tp).offset should be(10L)
+    tracker.receivedMessages(tp).timestamp should be(ConsumerRecord.NO_TIMESTAMP)
+  }
+
+  it should "skip tracking received records that are not assigned" in {
```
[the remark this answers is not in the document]
jyates: Looking at the code, i found the use of 'requested' confusing, so updating that to be 'commitRequested' (which is what is actually happening). then a little finagling on this test should make it clear what is going on.

core/src/main/scala/akka/kafka/ResetProtectionSettings.scala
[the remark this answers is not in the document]
seglo: Yes, someone forgot to include `@InternalApi` on that constructor too.

core/src/main/scala/akka/kafka/ConsumerSettings.scala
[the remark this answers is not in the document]
seglo: Ah, that's ok. Add a MiMa exception to squash the error.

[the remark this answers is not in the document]
seglo: Just an issue with MiMa.

docs/src/main/paradox/errorhandling.md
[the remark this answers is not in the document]
seglo: lol

core/src/main/scala/akka/kafka/internal/ConsumerProgressTracking.scala
```diff
@@ -0,0 +1,130 @@
...
+@InternalApi
+object ConsumerProgressTrackerNoop extends ConsumerProgressTracking {}
+
+/**
+ * Track the progress/state of the consumer. We generally try to be 'fast' with handling the partitions we track;
+ * most of the smarts are expected to be handled outside this class. For example, we will update any offsets to
+ * commit - [[commitRequested]] - without regard for what had previously been assigned
+ * or revoked from the consumer. Thus, care should be taken when managing state of the consumer and making updates.
+ *
+ * The only case we try and be "smart" is during [[received]], where we will filter out offsets that have not been
```
jyates, line 58: @seglo when going through the behavior in tests, this stood out as a bit strange to me. Either feels like we should be smart for all entry to this class or not smart at all (none of this filter-sometimes behavior). I prefer to be 'dumb' here if we can and push the logic into the KafkaConsumerActor if we can. If so, then we could do the "is-assigned" filtering where we also do the reset protection (https://github.com/akka/alpakka-kafka/pull/1299/files#diff-c05b458cf33e1bf97dadc6ea4b01c56daf01a6ddb72bd207c62a7d8331441d60R617 - line 617), and drop the SourceLogicBuffer. 

However, it is quite hard to _not_ do the filtering either here or in the actor as we want to we might receive offsets that are not assigned anymore due to the nature of poll results. In that case, we would want to just filter out everything that is not assigned for all the methods here.

WDYT?

seglo: > @seglo when going through the behavior in tests, this stood out as a bit strange to me. Either feels like we should be smart for all entry to this class or not smart at all (none of this filter-sometimes behavior). I prefer to be 'dumb' here if we can and push the logic into the KafkaConsumerActor if we can.

Good point. I'm leaning towards the "dumb" behaviour as well. What would you move to `KafkaConsumerActor` if we went this route?

> If so, then we could do the "is-assigned" filtering where we also do the reset protection (#1299 (files) - line 617), and drop the SourceLogicBuffer.

The `SourceLogicBuffer` filtering is important because a buffer exists between the `KafkaConsumerActor` and the first user stage. If partitions are revoked for messages already in the buffer then we'll send late messages, so it would be preferred to keep that logic as well. I'm not sure I understood your point correctly though.

jyates: Apologies, was on vacation last week....

I also seem to have forgotten the cleverness i was considering when i wrote that comment. However, im realizing that there might be a subtle bug here - if we haven't committed the partition yet, we won't have it in `requestedOffsetsImpl` (which tracks the commits that have been requested...and clearly needs a slight rename) which means we won't track it yet. That is a little unexpected behavior from the outside - you would think received would just track things we receive. 

All of this to say, we should probably _not_ try and be clever here (clearly its hard) and just eat the miniscule amount of overhead of tracking some more partitions; its a topic name, partition ID and a long value for each, not really that bad in terms of memory when compared with all the other moving parts here. If its does become bad (too much memory) we could have a chaining map that does topic -> partition -> offset...but that also seems overkill at the moment.

jyates: And now that i look more closely, if we just track the currently assigned partitions, we can filter on that instead. Since revokes and assignments are synchronous, we should be able to use that as our filtering mechanism; which to me makes at least logical sense from the outside.

tests/src/test/scala/akka/kafka/internal/ConsumerProgressTrackingSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+    tracker.commitRequested shouldBe empty
+    tracker.receivedMessages shouldBe empty
+    tracker.committedOffsets shouldBe empty
+
+    // once is is assigned, we should start tracking it
+    tracker.assignedPositions(Set(tp), Map(tp -> 0L))
+    // haven't received any records yet, stays empty
+    tracker.receivedMessages shouldBe empty
+    tracker.commitRequested.map(extractOffset) should be(Map(tp -> 0L))
+    tracker.commitRequested.map(extractOffset) should be(Map(tp -> 0L))
```
seglo, line 65: I think meant to be:

```suggestion
    tracker.committedOffsets.map(extractOffset) should be(Map(tp -> 0L))
```

core/src/main/resources/reference.conf
```diff
@@ -143,6 +143,21 @@ akka.kafka.consumer {
     backoff-factor = 2.0
   }
 
+  # Protect against server -ide bugs that cause Kafka to temporarily "lose" the latest offset for a consumer, which
```
seglo, line 146:
```suggestion
  # Protect against server-side bugs that cause Kafka to temporarily "lose" the latest offset for a consumer, which
```

core/src/main/resources/reference.conf
```diff
@@ -143,6 +143,21 @@ akka.kafka.consumer {
...
+  # then causes the Kafka consumer to follow is normal 'auto.offset.reset' behavior. For 'earliest', these settings
```
seglo, line 147:
```suggestion
  # then causes the Kafka consumer to follow its normal 'auto.offset.reset' behavior. For 'earliest', these settings
```

core/src/main/resources/reference.conf
```diff
@@ -143,6 +143,21 @@ akka.kafka.consumer {
...
+  # allow the client to detect and attempt to recover from this issue. For 'none' and 'latest', these settings will
+  # only add overhead. See
+  # https://doc.akka.io/docs/alpakka-kafka/current/errorhandling.html#unexpected-consumer-offset-reset
+  # for moreinformation
```
seglo, line 151:
```suggestion
  # for more information
```

core/src/main/scala/akka/kafka/ConsumerSettings.scala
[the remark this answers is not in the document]
seglo: Bump. Let's drop `this` overload and add the MiMa exception.

core/src/main/scala/akka/kafka/OffsetResetProtectionSettings.scala
```diff
@@ -0,0 +1,103 @@
...
+import akka.annotation.InternalApi
+import akka.util.JavaDurationConverters._
+import com.typesafe.config.Config
+
+import scala.concurrent.duration._
+
+class OffsetResetProtectionSettings @InternalApi private[kafka] (val enable: Boolean,
+                                                                 val offsetThreshold: Long,
+                                                                 val timeThreshold: FiniteDuration) {
+  require(offsetThreshold > 0, "A offset threshold must be greater than 0")
```
seglo, line 18:
```suggestion
  require(offsetThreshold > 0, "An offset threshold must be greater than 0")
```

core/src/main/scala/akka/kafka/internal/ConsumerResetProtection.scala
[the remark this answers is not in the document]
seglo: `sealed`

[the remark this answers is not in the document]
seglo: Can you add a link to #1286 here so it's easy to look up.

docs/src/main/paradox/errorhandling.md
```diff
@@ -35,3 +35,67 @@ If reading from Kafka failure is caused by other reasons, like **deserialization
...
+ exists. In this case, the consumer has three approaches to handling the missing offset:
+ 
+  * No action: consumer fails and stops attempting to make progress.
+  * Latest: consumer skips to the end of the partition and starts reading from there. 
+    - **NOTE**: consumer can skip processing some data in the topic by going to the latest data
+  * Earliest: consumer restarts from the beginning of the partition.
+    - **NOTE**: consumer will never skip data, but may reprocess many days of data, up to the topic's configured
+   retention
+
+Alpakka-kafka cannot do anything for the first two approaches. However, the `offset-reset-protection` configuration in
```
seglo, line 51:
```suggestion
Alpakka Kafka cannot do anything for the first two approaches. However, the `offset-reset-protection` configuration in
```

seglo (commented):
Looking good. I think everything major has been addressed. Just some small nits.

core/src/main/scala/akka/kafka/internal/ConsumerProgressTracking.scala
seglo: Looks ok to me.

seglo (approved):
LGTM. Thanks a lot for adding this feature @jyates !

PR #1307, opened 2021-01-11
Migrate build to GitHub actions

seglo:
Continuation from #1289


seglo:
I tested the release (snapshot) to bintray and gustav.

seglo:
PR won't go green because travis seems stuck trying to build because no `.travis.yml` exists.

.github/workflows/check-build-test.yml
```diff
@@ -0,0 +1,220 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [ master ]
+    tags-ignore: [ v.* ]
+  schedule:
+    - cron:  '0 0 1 * *' # monthly
```
ennru, line 9: What's you thinking here?

.github/workflows/check-build-test.yml
```diff
@@ -0,0 +1,220 @@
...
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+          fetch-depth: 0
```
ennru, line 24: Would be interesting to see if there is middle ground. In other repos we try to fetch the tags explicitly. Not sure what is more efficient, but some repos are quite huge and would benefit from not fetching all history.

.github/workflows/check-build-test.yml
```diff
@@ -0,0 +1,220 @@
...
+      - name: Compile all code with fatal warnings for Scala 2.13
+        # Run locally with: env CI=true sbt 'clean ; Test/compile ; It/compile'
+        run: sbt "; Test/compile; It/compile"
+
+      - name: Create all API docs for artifacts/website and all reference docs
+        # Run locally with: sbt verifyDocs
+        run: sbt verifyDocs
+
+  test-java8-scala212:
+    name: Build and Test (Java 8, Scala 2.12)
```
ennru, line 54: Did you consider a matrix for these?

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+      - main
+      - release-*
+    tags: [ v.* ]
```
ennru, line 9: There is even a trigger on GitHub release. But this is what we did in Travis.

ennru (commented):
Looks good. A pity that the basic setup needs to be repeated over and over.

.github/workflows/check-build-test.yml
seglo: I thought there was a monthly cron on Travis, but that's just the default line item to add a cron.

There's a line in `build.sbt` that toggles `Nightly` when the travis event type is `cron`, but if we don't have a travis cron then I don't know its purpose.

.github/workflows/check-build-test.yml
seglo: I didn't experiment with this much.  I tried a depth of 1000 and the dyn ver stuff still failed.

.github/workflows/check-build-test.yml
seglo: I didn't. I'll look into it.

seglo:
> A pity that the basic setup needs to be repeated over and over.

I suppose we could make our own action if we get consensus on these patterns in the akka builds.

.github/workflows/check-build-test.yml
andreaTP: We have all kind of workarounds in place for this, e.g.:
https://github.com/lightbend/cloudflow/blob/master/.github/workflows/push.yml#L21

.github/workflows/check-build-test.yml
ennru: We used to have Akka 2.6 builds as a Travis cron job.

.github/workflows/release.yml
ennru: Will this trigger for regular pushes to those branches so that it builds snapshots? Or does it build only if there is a v-tag?

seglo: It will build for tags and the branches defined.

seglo: I haven't actually tested the tag build though.

.github/workflows/check-build-test.yml
seglo: Ok. I'll remove this cron.

.github/workflows/check-build-test.yml
seglo: I'll see if we need to optimize this later.  Seems fine for now.

.github/workflows/check-build-test.yml
seglo: Very slick :+1:

.github/workflows/release.yml
seglo: I'll keep the trigger on tag for now.

patriknw (commented):
looking good, but I'll leave the final reviews to the others

octonato (approved):

seglo:
I'm going to merge this for now. Feel free to add any late reviews and I can address them in a follow up PR.

PR #1310, opened 2021-01-30
Use restart script

seglo:
This was meant to be a part of the testcontainer refresh for 2.0.6, but it was omitted.  The restart script lets the user control instance of Kafka in a container, without managing containers directly.


ennru (approved):
LGTM. That explains my question in #1311

seglo:
Build failures caused by timeouts that should be reduced with #1311

PR #1328, opened 2021-02-09
Update scaladex badge

seglo:
The default scaladex artifact doesn't exist any more (`reactive-kafka`) so it was always pointing to an old version `0.7.2`. Including the main dependency in the badge URL gets the latest non-pre-release (`2.0.7`). Since it's not showing the latest pre-release it's out of sync with the maven central badge which is `2.1.0-M1`. I'm not sure if this is something we really need to address.


ennru (approved):
👍

Issue #1126, opened 2020-06-11
Failed: A consume-transform-produce cycle must rebalance safely using transactional partitioned flow, timeout


seglo:
From an unrelated PR build

https://travis-ci.org/github/akka/alpakka-kafka/jobs/696759321#L465	

```
Failed(java.lang.Error: Timeout while waiting for desired consumer group state. Tried [100] times, slept [100 milliseconds] in between.)
```

seglo:
Possibly fixed by #1109.

seglo:
This is actually a timeout of a `waitUntilConsumerSummary` call (10 second timeout).

https://travis-ci.com/github/akka/alpakka-kafka/jobs/480442181#L3229

PR #1329, opened 2021-02-09
Increase consumer group summary timeout in build

seglo:
Hopefully reduces this will reduce the occurrence of #1126.


ennru (approved):
LGTM.

PR #1354, opened 2021-04-12
Pin version of protobuf used

seglo:
Disable scala steward PRs for `protobuf-java`.


seglo:
Not necessary for Alpakka Kakfa

PR #1355, opened 2021-04-12
Upgrade Akka 2.6.14 and Kafka 2.7.0

seglo:
Upgrade core deps for upcoming release

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,19 +10,20 @@ val Scala212 = "2.12.13"
 val Scala213 = "2.13.4"
 
 val AkkaBinaryVersionForDocs = "2.6"
-val KafkaVersionForDocs = "26"
+val KafkaVersionForDocs = "27"
 
-val akkaVersion = "2.6.10"
-val kafkaVersion = "2.6.0"
-// TODO Jackson is now a provided dependency of kafka-clients
-// https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/2.6.0
-val jacksonVersion = "2.10.5.1"
+val akkaVersion = "2.6.14"
+val kafkaVersion = "2.7.0"
+// Jackson is now a provided dependency of kafka-clients
+// This should align with the Jackson minor version used in Akka 2.6.x
+// https://github.com/akka/akka/blob/master/project/Dependencies.scala#L23
+val jacksonVersion = "2.11.4"
 val scalatestVersion = "3.1.4"
 val testcontainersVersion = "1.15.1"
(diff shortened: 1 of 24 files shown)


build.sbt
```diff
@@ -10,32 +10,28 @@ val Scala212 = "2.12.13"
...
-val kafkaVersion = "2.6.0"
-// TODO Jackson is now a provided dependency of kafka-clients
-// https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/2.6.0
-val jacksonVersion = "2.10.5.1"
+val akkaVersion = "2.6.14"
+val kafkaVersion = "2.7.0"
+// Jackson is now a provided dependency of kafka-clients
+// This should align with the Jackson minor version used in Akka 2.6.x
+// https://github.com/akka/akka/blob/master/project/Dependencies.scala#L23
+val jacksonVersion = "2.11.4"
```
ennru, line 20: Add a reminder for the release notes as kafka-clients 2.7.0 is still on Jackson 2.10.5.

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
```diff
@@ -373,9 +373,10 @@ class IntegrationSpec extends SpecBase with TestcontainersKafkaLike with Inside
         .run()
 
       // Wait a tiny bit to avoid a race on "not yet initialized: only setHandler is allowed in GraphStageLogic constructor"
```
ennru, line 375: I believe you need to keep the sleep to avoid the race (it rarely shows).

ennru (approved):
LGTM.

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
seglo: Something was triggering the race condition frequently after the upgrade. I played with different intervals, but low numbers didn't seem to work consistently. I'll take another look.

build.sbt
seglo: I noticed that as well. Since the tests seem to work with this pulled up version I think it's fine to use this. What would include in the release notes? "Akka uses 2.11, Kafka refs (provided) 2.10. Akka uses 2.11 because ____"

tests/src/test/scala/akka/kafka/scaladsl/IntegrationSpec.scala
seglo: I'll keep both.

PR #1395, opened 2021-07-22
Fix context propagation in CommitCollectorStage

ygree:
References https://github.com/lightbend/cinnamon/issues/2874

Lightbend Telemetry tests: https://github.com/lightbend/cinnamon/pull/2875

+ support context propagation in CommitCollectorStage for Lightbend Telemetry
+ bump Akka to 2.6.15
+ fixed MatchError resulting in the stage failure

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
@@ -42,7 +42,12 @@ private final class CommitCollectorStageLogic(
   import CommitCollectorStage._
   import CommitTrigger._
 
-  final val settings: CommitterSettings = stage.committerSettings
+  val settings: CommitterSettings = stage.committerSettings
+
+  // Context propagation is needed to notify Lightbend Telemetry to keep the context in case of a deferred downstream
+  // push call that might not happen during onPush but later onTimer, onPull, or only during the next onPush call.
+  private val contextPropagation = akka.stream.impl.ContextPropagation()
+  private var contextSuspended = false
 
   override protected def logSource: Class[_] = classOf[CommitCollectorStageLogic]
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
```diff
@@ -79,18 +91,24 @@ private final class CommitCollectorStageLogic(
...
         if (updateBatch(offset)) {
           // Push only of the outlet is available, a commit on interval might have taken the pending demand.
           // This is very hard to get tested consistently, so it gets this big comment instead.
-          if (isAvailable(stage.out)) pushDownStream(BatchSize)
+          if (isAvailable(stage.out)) {
+            pushDownStream(BatchSize)
+            pushed = true
+          }
         } else tryPull(stage.in) // accumulating the batch
+        if (!pushed) suspendContext()
```
octonato, line 105: minor suggestion: 

Since here we have only one else branch to care about, we could have: 

```suggestion
          if (isAvailable(stage.out)) pushDownStream(BatchSize)
        } else {
          tryPull(stage.in) // accumulating the batch
          suspendContext()
        }
```

That removes the need for `var pushed = false` above.

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
```diff
@@ -51,7 +52,14 @@ import scala.concurrent.{Future, Promise}
     if (!isClosed(shape.out)) {
       complete(shape.out)
     }
-    sourceActor.become(shuttingDownReceive)
+    sourceActor.become(shuttingDownReceive.orElse {
+      case (_, Messages(requestId, messages)) =>
+        // Prevent stage failure during shutdown by ignoring Messages
+        if (messages.hasNext)
+          log.debug("Unexpected message `Messages` received with requestId={} and a non-empty message iterator: {}",
```
octonato, line 59: Reads weird 'message `Messages`'. We can maybe drop 'message'.

```suggestion
          log.debug("Unexpected `Messages` received with requestId={} and a non-empty message iterator: {}",
```

I didn't understand why this is related with the changes for telemetry though. Have run into tests errors that required this?

octonato (approved):
LGTM with some minor suggestions (and question)

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
octonato: About my question above, please ignore it, you mentioned it in the description.

> fixed MatchError resulting in the stage failure

ignasi35: I like to use the `Hide` feature to reduce noise. :-) 

Screenshot 2021-07-23 at 08 52 34

octonato: This is not off-topic. There is a code suggestion.

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
ignasi35: @octonato I think your code should be:

```
        if (updateBatch(offset)) {
          if (isAvailable(stage.out)) pushDownStream(BatchSize)
          else suspendContext()
        } else {
          tryPull(stage.in) // accumulating the batch
          suspendContext()
        }
```

then it's not much simpler, is it?

core/src/main/scala/akka/kafka/internal/SingleSourceLogic.scala
ignasi35: Apologies, I don't know what the hell happened there. I think I read a PR with comments of yours where you added corrections to yourself and I used the `Hide` feature on the wrong PR.

ignasi35 (approved):

core/src/main/scala/akka/kafka/internal/CommitCollectorStage.scala
ygree: I was trying to avoid repeating suspendContext in each branch other than those calling pushDownStream. I think this way it's less probable to forget to call suspendContext.

PR #1432, opened 2021-11-08
Upgrade scalatest to 3.2.10

mrooding:
Upgrades scalatest to 3.2.10. 

Please let me know if you'd prefer a split between 3.1 and 3.2 depending on the scala version


ennru:
We'd preferably sync this with the version used in Akka's testkit. 
That's why it is excluded from Scala Steward: https://github.com/akka/alpakka-kafka/blob/master/.scala-steward.conf#L11

ennru:
Just looking at this again. ScalaTest 3.1.4 is a `provided` dependency of our test kit, so you should be able to easily choose a later version.

ennru:
I assume the note about it being `provided` solved your need. Please reopen if not.

PR #1443, opened 2021-12-13
Update sbt plugins; replace silencer with nowarn

ennru:
- sbt plugin updates (required scalafmtAll)
- replace silencer with nowarn (no change in Scala versions)


ennru:
The pending checks will not complete, as they got updated to show Scala version 2.13.4.

RayRoestenburg (approved):
LGTM

PR #1448, opened 2021-12-23
Use Scala 2.13 (drop Scala 2.12)

ennru:
* Scala 2.13 only
* Fix more compiler warnings
* Update readme


core/src/main/scala/akka/kafka/internal/CommittingProducerSinkStage.scala
```diff
@@ -167,6 +167,7 @@ private final class CommittingProducerSinkStageLogic[K, V, IN <: Envelope[K, V,
 
   override protected def onTimer(timerKey: Any): Unit = timerKey match {
     case CommittingProducerSinkStage.CommitNow => commit(Interval)
+    case _ =>
```
johanandren, line 170: would be good to throw to avoid potential future foot-shooting/head-scratching

johanandren (approved):
LGTM if CI says so

core/src/main/scala/akka/kafka/internal/CommittingProducerSinkStage.scala
ennru: Right, I added more logging/errors for unexpected matches.

PR #1469, opened 2022-02-11
Add getProperty to ProducerSettings

mdedetrich:
This PR adds the equivalent `getProperty` from `ConsumerSettings` (see https://github.com/akka/alpakka-kafka/blob/eb5246d3e03992ce60a1f33fefe75dadc58bc866/core/src/main/scala/akka/kafka/ConsumerSettings.scala#L333-L336) into `ProducerSettings`


ennru:
Thank you!

PR #1492, opened 2022-04-09
testkit: fail better if cluster hasn't started

### LINKED ISSUE #1491: Calling methods which use `cluster` under the hood can cause `NullPointerExecption`
### Versions used 
Akka version: `latest` (code on master)

### Description
Consider the following code:
```scala
abstract class SpecBase(kafkaPort: Int) extends ScalatestKafkaSpec(kafkaPort) {
  protected def this() = this(kafkaPort = -1)
}

class FooTest extends SpecBase with TestcontainersKafkaLike {
  val bar = brokerContainers.head.getBrokerNum

}
```
Running this suite will result in the following exception:
```
An exception or error caused a run to abort: Cannot invoke "akka.kafka.testkit.internal.KafkaContainerCluster.getBrokers()" because the return value of "akka.kafka.testkit.internal.TestcontainersKafka$Spec.akka$kafka$testkit$internal$TestcontainersKafka$Spec$$cluster()" is null 
...
```
Of course, the issue here is that I'm trying to get information regarding the cluster when before I even started it.

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.
A proper exception, saying the cluster hasn't been started, will be thrown.

### Actual Behavior
The method returns a `NullPointerExecption`

### Suggested Solution
Check by default if the cluster has been started, like other methods do:

```

yuvalshi0:
Closes #1491 
*Not sure if it's the best solution

### DIFF
--- a/testkit/src/main/scala/akka/kafka/testkit/internal/TestcontainersKafka.scala
+++ b/testkit/src/main/scala/akka/kafka/testkit/internal/TestcontainersKafka.scala
@@ -16,16 +16,12 @@ import scala.jdk.CollectionConverters._
 
 object TestcontainersKafka {
   trait Spec extends KafkaSpec {
+    private var cluster: KafkaContainerCluster = _
     private var kafkaBootstrapServersInternal: String = _
     private var kafkaPortInternal: Int = -1
 
     private def requireStarted(): Unit =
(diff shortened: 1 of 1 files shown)


ennru (approved):
LGTM.

ennru:
Thank you!

PR #1493, opened 2022-04-09
Reduce kafka docker version constraint

### LINKED ISSUE #1424: Reduce kafka docker version constraint
### Short description

It would be nice to have the ability to run kafka testcontainers that don't fit the constraints enforced by the `Version` class.  It assumes a semantic versioning scheme that is not used by all docker tags

### Details

In order to test with redpanda or other kafka compatible containers, it would be good to loosen the tag constraints implemented in `Version`.  They seem to be there primarily to control switching between options that would be better made more explicit, ie `--zookeeper` etc....

yuvalshi0:
Closes #1424 

I don't know if this is the best solution, but since some people might want to use a custom tag (like building a `cp-kafka` image on M1), a good idea might be to just use the 5.2.0 and above connect param, tested at: https://github.com/akka/alpakka-kafka/blob/9419b5d9cb420026759fc0b11e669c2266934874/testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java#L310

So, if the tag is not "comparable", we will just use the newer connect param

### DIFF
--- a/testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
+++ b/testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
@@ -341,7 +341,7 @@ public class KafkaContainerCluster implements Startable {
 class Version implements Comparable<Version> {
 
   private String version;
-  private boolean comparable = false;
+  private boolean comparable = true;
 
   public final String get() {
     return this.version;
--- a/testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
+++ b/testkit/src/main/java/akka/kafka/testkit/internal/KafkaContainerCluster.java
@@ -347,6 +347,10 @@ class Version implements Comparable<Version> {
     return this.version;
   }
 
+  public final boolean isComparable() {
+    return this.comparable;
+  }
+
   public Version(String version) {
     if (version == null) throw new IllegalArgumentException("Version can not be null");
     if (!version.matches("[0-9]+(\\.[0-9]+)*")) this.comparable = false;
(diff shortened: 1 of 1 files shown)


ennru:
Wouldn't it be much better to allow setting the value to be returned from `kafkaTopicConnectParam`?

yuvalshi0:
> Wouldn't it be much better to allow setting the value to be returned from `kafkaTopicConnectParam`?

And saving the version as a simple string? Couldn't it break backward compatibility?

ennru:
Ok, I finally looked a bit into what you are trying to solve here. Your problem boils down which flags to pass on the command lines for zookeeper or non-zookeeper.
My suggestion is to move the comparison into the `Version` class as `usesZookeeper` and add some logic that works for the version schemes you notice with Redpanda and others.

patriknw:
@yuvalshi0 What do you think about @ennru 's suggestion, and would you be able to continue on this PR?

yuvalshi0:
> Ok, I finally looked a bit into what you are trying to solve here. Your problem boils down which flags to pass on the command lines for zookeeper or non-zookeeper. My suggestion is to move the comparison into the `Version` class as `usesZookeeper` and add some logic that works for the version schemes you notice with Redpanda and others.

I personally don't use red panda, rather I build a completely custom tags (mainly because there are no official arm docker images for Kafka), how do you think we should handle a tag which does not follow any schema?

yuvalshi0:
> @yuvalshi0 What do you think about @ennru 's suggestion, and would you be able to continue on this PR?

Missed that PR because Gmail moved it to my spam 😅, will do now

PR #1502, opened 2022-05-13
docs: Better match JUnit tests with version

leviramsey:
The linked docs page gets things right, but testing.md swaps things around


ennru (approved):
LGTM.

PR #1518, opened 2022-08-10
bump: run validations on Ubuntu 22.04 (was 18.04)

ennru:
Other actions use ubuntu-latest, but these are convenient to have on a fixed version.


efgpinto (approved):
LGTM

PR #1519, opened 2022-08-12
build: align Scala 2.13 version for releasing

ennru:
The releasing on every build has been broken for some time now.
https://github.com/akka/alpakka-kafka/runs/7783726593?check_suite_focus=true#step:7:74


efgpinto (approved):
LGTM

PR #1537, opened 2022-09-19
Update protobuf dependency in tests

patriknw:
CVE-2021-22569, but only in test code.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -251,7 +251,7 @@ lazy val tests = project
     name := "akka-stream-kafka-tests",
     libraryDependencies ++= Seq(
         "com.typesafe.akka" %% "akka-discovery" % akkaVersion,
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -269,7 +269,7 @@ lazy val tests = project
         // Schema registry uses Glassfish which uses java.util.logging
         "org.slf4j" % "jul-to-slf4j" % slf4jVersion % Test,
         "org.mockito" % "mockito-core" % "4.7.0" % Test,
-        "com.thesamet.scalapb" %% "scalapb-runtime" % "0.10.11" % Test
+        "com.thesamet.scalapb" %% "scalapb-runtime" % "0.11.11" % Test
```
patriknw, line 272: this depends on protobuf-java 3.19.2, but latest patch 3.19.5 should be fine

PR #1538, opened 2022-09-19
Add back Scala 2.12

patriknw:
* because it has ripple effect to Alpakka and then Cassandra plugin
* we need to coordinate this decision across all Akka projects, later

This reverts some parts, as little as possible, of https://github.com/akka/alpakka-kafka/pull/1448

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,6 +9,7 @@ val Nightly = sys.env.get("EVENT_NAME").contains("schedule")
 // align ignore-prefixes in scripts/link-validator.conf
 // align in release.yml
 val Scala213 = "2.13.8"
+val Scala212 = "2.12.16"
 
 val AkkaBinaryVersionForDocs = "2.6"
 val akkaVersion = "2.6.19"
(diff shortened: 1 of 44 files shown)


build.sbt
```diff
@@ -81,7 +82,7 @@ val commonSettings = Def.settings(
   scalacOptions ++= Seq(
       "-encoding",
       "UTF-8", // yes, this is 2 args
-      "-Wconf:cat=feature:w,cat=deprecation:w,cat=unchecked:w,cat=lint:w,cat=unused:w,cat=w-flag:w"
+      "-Wconf:cat=feature:w,cat=deprecation&msg=.*JavaConverters.*:s,cat=unchecked:w,cat=lint:w,cat=unused:w,cat=w-flag:w"
```
patriknw, line 85: to avoid all warnings for JavaConverters, but not silence everything
https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html

ennru (approved):
LGTM.

PR #1539, opened 2022-09-20
Akka 2.7.0-M1 dependency

patriknw:


patriknw:
a symlink for 2.7 was wrong, but I have fixed that now

wonder why it's not reporting back on two of the jobs, which also looks like they are duplicates

PR #1540, opened 2022-09-20
Temporary remove Alpakka csv dependency

patriknw:
* problematic with the cycle
* missing Scala 2.12


patriknw:
404 https://dist.apache.org/repos/dist/release/kafka/3.0.1/RELEASE_NOTES.html because they have released 3.0.2?

PR #1545, opened 2022-09-28
Update protobuf-java to 3.21.1, in tests

patriknw:
https://github.com/akka/alpakka-kafka/pull/1537 wasn't enough, must also override the protobuf-java dependency
This is same version that is currently in use in akka-grpc.


ennru (approved):
LGTM.

PR #1547, opened 2022-10-13
bump: kafka-clients 3.3.1

ennru:
A new method `Consumer.enforceRebalance(reason)` got introduced, so the client library is not compatible with earlier versions.

https://kafka.apache.org/33/documentation.html
https://archive.apache.org/dist/kafka/3.3.1/RELEASE_NOTES.html


patriknw (approved):
LGTM

PR #1548, opened 2022-10-13
bump: Scala 2.13.10

ennru:


johanandren (approved):

build.sbt
```diff
@@ -8,7 +8,7 @@ val Nightly = sys.env.get("EVENT_NAME").contains("schedule")
 
 // align ignore-prefixes in scripts/link-validator.conf
 // align in release.yml
-val Scala213 = "2.13.8"
+val Scala213 = "2.13.10"
 val Scala212 = "2.12.16"
```
patriknw, line 12: Would be good if we bump to 2.12.17 at the same time so that we use 2.13.10 and 2.12.17 all over the place

PR #1564, opened 2022-10-14
bump: Akka M5; sbt 1.7.2 and test library versions

ennru:
Supersedes #1563, #1562, #1561, #1560, #1559, #1558, #1557, #1556, #1555, #1554, #1553


patriknw (approved):
LGTM

PR #1574, opened 2022-11-07
chore: current GH actions; coursier/setup-action

ennru:
GitHub workflow housekeeping (to not hit Node 12 EOL):
- checkout v3
- coursier/setup-action instead of olafurpg/setup-scala

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -12,16 +12,15 @@ permissions:
 
 jobs:
   update_release_draft:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     if: github.repository == 'akka/alpakka-kafka'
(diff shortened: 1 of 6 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -7,10 +7,19 @@ on:
...
 jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     if: github.repository == 'akka/alpakka-kafka'
+    permissions:
+      # write permission is required to create a github release
+      contents: write
+      # write permission is required for autolabeler
+      # otherwise, read permission is required at least
+      pull-requests: write
```
ennru, line 22: See https://github.com/release-drafter/release-drafter#usage

johanandren (approved):

Issue #266, opened 2017-01-19
AskTimeoutException when using commitScaladsl


YunLiaw:
Hello,

I sometimes receive this error when running reactive kafka and I do not have any clue.

```
akka.pattern.AskTimeoutException: Ask timed out on [Actor[akka://default/system/kafka-consumer-1#1749876186]] after [15000 ms]. Sender[null] sent message of type "akka.kafka.KafkaConsumerActor$Internal$Commit".
	at akka.pattern.PromiseActorRef$$anonfun$1.apply$mcV$sp(AskSupport.scala:604)
	at akka.actor.Scheduler$$anon$4.run(Scheduler.scala:126)
	at scala.concurrent.Future$InternalCallbackExecutor$.unbatchedExecute(Future.scala:601)
	at scala.concurrent.BatchingExecutor$class.execute(BatchingExecutor.scala:109)
	at scala.concurrent.Future$InternalCallbackExecutor$.execute(Future.scala:599)
	at akka.actor.LightArrayRevolverScheduler$TaskHolder.executeTask(LightArrayRevolverScheduler.scala:329)
	at akka.actor.LightArrayRevolverScheduler$$anon$4.executeBucket$1(LightArrayRevolverScheduler.scala:280)
	at akka.actor.LightArrayRevolverScheduler$$anon$4.nextTick(LightArrayRevolverScheduler.scala:284)
	at akka.actor.LightArrayRevolverScheduler$$anon$4.run(LightArrayRevolverScheduler.scala:236)
	at java.lang.Thread.run(Thread.java:745)
```

And below is the code for the stream:

```scala
    Consumer.committablePartitionedSource(consumerSettings, Subscriptions.topics("consumerTopic"))
    .map { case (_, source) =>
      source
      .map { m =>
        
        // business logic
        toProducerMessage(...)
      }
      .via(Producer.flow(producerSettings, producerSettings.createKafkaProducer()))
      .map(_.message.passThrough)
      .batch(20, CommittableOffsetBatch.empty.updated) { (batch, elem) => batch.updated(elem) }
      .mapAsync(producerSettings.parallelism)(_.commitScaladsl()) // I believe here's the place for the exception
      .toMat(Sink.ignore)(Keep.right)
      .run()
    }
    .mapAsyncUnordered(consumerMaxPartition)(identity)
    .to(Sink.ignore)
  }.run()
```

This exception will sometimes happen. For now I have two ways to avoid this exception:
[...]

PR #1584, opened 2022-11-18
retry committing for transient exceptions

omeraha:
### Background
Offset commits might fail due to timeout on the Kafka broker, in which case the broker returns an error stating the [operation timed out](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L1383). Timeouts are transient in nature, so it might be worth to retry the commit.
### Changes 

- Attempt to retry a commit that failed due to `TimeoutException` exception.
- Add timeout exception scenario to the commit retry test.

References #1111

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
+++ b/core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
@@ -27,7 +27,11 @@ import akka.kafka.KafkaConsumerActor.{StopLike, StoppingException}
 import akka.kafka._
 import akka.kafka.scaladsl.PartitionAssignmentHandler
 import org.apache.kafka.clients.consumer._
-import org.apache.kafka.common.errors.{RebalanceInProgressException, TimeoutException}
+import org.apache.kafka.common.errors.{
+  CoordinatorLoadInProgressException,
+  RebalanceInProgressException,
+  TimeoutException
+}
 import org.apache.kafka.common.{Metric, MetricName, TopicPartition}
 
 import scala.annotation.nowarn
@@ -588,7 +592,8 @@ import scala.util.control.NonFatal
               progressTracker.committed(offsets)
               replyTo.foreach(_ ! Done)
 
-            case e @ (_: RebalanceInProgressException | _: TimeoutException) => retryCommits(duration, e)
+            case e @ (_: RebalanceInProgressException | _: TimeoutException | _: CoordinatorLoadInProgressException) =>
+              retryCommits(duration, e)
             case e: RetriableCommitFailedException => retryCommits(duration, e.getCause)
 
             case commitException =>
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @omeraha,

[https://www.lightbend.com/contribute/cla](https://www.lightbend.com/contribute/cla)

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
```diff
@@ -588,7 +588,7 @@ import scala.util.control.NonFatal
               progressTracker.committed(offsets)
               replyTo.foreach(_ ! Done)
 
-            case e: RebalanceInProgressException => retryCommits(duration, e)
+            case e @ (_: RebalanceInProgressException | _: TimeoutException) => retryCommits(duration, e)
```
ennru, line 591: I agree with your comment in https://github.com/akka/alpakka-kafka/issues/266#issuecomment-1320935836 that this could handle all kinds of `RetriableException` here instead. That would even catch `RetriableCommitFailedException` below. 
I assume most of the subclasses to `RetriableException` can not happen just here.

omeraha: @ennru I'll fix it and commit. I'll also take a look at the classes that (currently) extend `RetriableException`. But it should be safe to retry all of them inside a callback on the commit result (as this means they must have been thrown during the commit itself)

omeraha: @ennru The following (specific) exceptions will be handled if we decide to go for `RetriableException`:

1. CoordinatorLoadInProgressException
2. UnknownTopicOrPartitionException
3. CoordinatorNotAvailableException
4. NotCoordinatorException
5. TimeoutException

While it makes sense to retry a timed-out operation and the `CoordinatorLoadInProgressException` are clearly safe to retry (we've seen `CoordinatorLoadInProgressException` in our services a couple of times), the 3 other are not entirely clear to me, for instance: the `UnknownTopicOrPartitionException` comment states that 
```
This exception is used in contexts where a topic doesn't seem to exist based on possibly stale metadata.
 This exception is retriable because the topic or partition might subsequently be created.
```
Which feels that this operation might be retried endlessly (as currently `retryCommits()` doesn't have a limit on the number of retries, if I understand the code correctly)

ennru: Thank you for looking closely at those sub-exceptions. It was the endless retrying I was afraid of, stale metadata sounds like a reason to try, but if the topic or partition is never created we it will loop.

ennru (approved):
LGTM.

core/src/main/scala/akka/kafka/internal/KafkaConsumerActor.scala
omeraha: @ennru Thanks. Anything else on my side?

ennru:
Thank you for improving Alpakka Kafka!

PR #1585, opened 2022-12-08
ci: cover JDK 17 via scheduled build

ennru:
Full matrix to cover even JDK17 via a scheduled build.
It looks as if Scala 2.12 wasn't actually covered in any CI build.

Follow-up to #1574

### DIFF
--- a/.github/workflows/check-build-test.yml
+++ b/.github/workflows/check-build-test.yml
@@ -107,13 +107,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # other combinations are covered in `scheduled-builds.yml`
-          - { scalaVersion: "2.12", jdkVersion: "1.8.0",  jvmName: "temurin:1.8.0",  extraOpts: '' }
-          # { scalaVersion: "2.12", jdkVersion: "1.11.0", jvmName: "temurin:1.11.0", extraOpts: '-J-XX:+UnlockExperimentalVMOptions -J-XX:+UseJVMCICompiler' }
-          # { scalaVersion: "2.12", jdkVersion: "1.17.0", jvmName: "temurin:1.17.0", extraOpts: '' }
-          # { scalaVersion: "2.13", jdkVersion: "1.8.0",  jvmName: "temurin:1.8.0", extraOpts: '' }
-          - { scalaVersion: "2.13", jdkVersion: "1.11.0", jvmName: "temurin:1.11.0", extraOpts: '-J-XX:+UnlockExperimentalVMOptions -J-XX:+UseJVMCICompiler' }
-          # { scalaVersion: "2.13", jdkVersion: "1.17.0", jvmName: "temurin:1.17.0", extraOpts: '' }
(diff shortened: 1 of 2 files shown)


johanandren (approved):
LGTM

.github/workflows/check-build-test.yml
```diff
@@ -124,13 +129,13 @@ jobs:
...
+      - name: Set up JDK ${{ matrix.jdkVersion }}
         uses: coursier/setup-action@v1.3.0
         with:
-          jvm: ${{ matrix.java-version }}
+          jvm: ${{ matrix.jvmName }}
 
-      - name: Run tests with Scala ${{ matrix.scala-version }} and Java ${{ matrix.java-version }}
-        run: sbt ${{ matrix.sbt-opts }} "test"
+      - name: sbt test
+        run: sbt ++${{ matrix.scalaVersion }} test ${{ matrix.extraOpts }}
```
patriknw, line 138: Note that if we have mixed cross Scala versions in the sbt projects this will not work (for Scala 3).

ennru: Thank you. That was added after I started on this.

.github/workflows/check-build-test.yml
```diff
@@ -107,8 +107,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { java-version: 'temurin:1.8',  sbt-opts: '' }
-          - { java-version: 'temurin:1.11', sbt-opts: '-J-XX:+UnlockExperimentalVMOptions -J-XX:+UseJVMCICompiler' }
+          # other JDKs are covered in `scheduled-builds.yml`
+          - { jdkVersion: "1.8.0",  jvmName: "temurin:1.8.0",  extraOpts: '' }
+          - { jdkVersion: "1.11.0", jvmName: "temurin:1.11.0", extraOpts: '-J-XX:+UnlockExperimentalVMOptions -J-XX:+UseJVMCICompiler' }
+          # { jdkVersion: "1.17.0", jvmName: "temurin:1.17.0", extraOpts: '' }
```
johanandren, line 113: I think this covers the JDK versions well enough but what about the Scala versions?

ennru: Right I guess we need a `+` somewhere to run tests for all three Scala versions in the build.

patriknw: In several other projects we use `+Test/compile` in PR validation and only run tests with one Scala and JDK. I don't think it's worth running too many combinations in PR validation.

patriknw: That `+Test/compile` should be with JDK 8 (lowest supported), and that is not the case in this build.

ennru:
The intention of this PR got lost underway. I'll close it and may open a new PR for JDK17 in CI.

.github/workflows/check-build-test.yml
ennru: For the record "Check Code Compilation" currently runs `sbt +Test/compile; +It/compile` on JDK 11.

Issue #943, opened 2019-10-21
Investigate different consumer group rebalance strategies


seglo:
### Short description

It's been observed that every time an Alpakka Kafka source app is run that all partitions are first revoked before assignments take place.  We should investigate why this is the case and if this is normal behaviour.

### Details

Explore the default consumer group rebalance strategies.  According to Kafka docs the default strategy is defined by the consumer property `partition.assignment.strategy`, with a default of `class org.apache.kafka.clients.consumer.RangeAssignor` (https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/consumer/RangeAssignor.html).  See what happens when try using another provided assignor, such as the `StickyAssignor` (https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/consumer/StickyAssignor.html).

jsbrucker:
This seems to be the expected "normal behavior" for legacy compatibility reasons (if you have an always on system, there is a procedure to rollout the change in strategy).

I have found that you can use:
```
.withProperty(
   // Equivalent to "partition.assignment.strategy"
  org.apache.kafka.clients.consumer.ConsumerConfig.PARTITION_ASSIGNMENT_STRATEGY_CONFIG,
  // We want the name of the org.apache.kafka.clients.consumer.CooperativeStickyAssignor class
  // either using the the String directly or using getName on the type.
  "CooperativeStickyAssignor"
)
```
when configuring `akka.kafka.ConsumerSettings`. NOTE: it is probably possible to set this value in the typesafe Config.

After making the change above, when new consumers are added/removed, the rebalance only seems to impact partitions that need to be moved while the others continue processing.

ennru:
Thank you, @jsbrucker for updating this issue with these details.
I guess it is time for Alpakka Kafka to default to the `CooperativeStickyAssignor` in the next release.

patriknw:
@jsbrucker Do you have any insight to how such change would work during a rolling update when old nodes have the other (default) strategy? Any risk of wrong partition assignments (multiple consumers of same partition)?

Seems like compatibility has been the main reason for not changing the default in Kafka, see https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=177048248

If that is the case we should not change the default, but we can add a recommendation to the documentation with caveat about rolling updates.

jsbrucker:
@patriknw

[...]

PR #1598, opened 2023-02-03
feat: simpler access to partition assignment strategies

ennru:
Add direct access to control partition assignment strategies and specifically the `CooperativeStickyAssignor`.

References #943

### DIFF
--- a/core/src/main/scala/akka/kafka/ConsumerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ConsumerSettings.scala
@@ -300,6 +300,26 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
     withProperty(ConsumerConfig.GROUP_INSTANCE_ID_CONFIG, groupInstanceId)
 
   /**
+   * A list of class names or class types, ordered by preference, of supported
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/kafka/ConsumerSettings.scala
```diff
@@ -299,6 +299,26 @@ class ConsumerSettings[K, V] @InternalApi private[kafka] (
...
     withProperty(ConsumerConfig.GROUP_INSTANCE_ID_CONFIG, groupInstanceId)
 
+  /**
+   * A list of class names or class types, ordered by preference, of supported
+   * partition assignment strategies that the client will use to distribute
+   * partition ownership amongst consumer instances when group management is used.
+   *
+   * See https://kafka.apache.org/documentation/#consumerconfigs_partition.assignment.strategy
+   */
+  def withPartitionAssignmentStrategy(strategy: String): ConsumerSettings[K, V] =
```
leviramsey, line 309: Might be nice to have a convenience method which takes a collection (maybe an array?) of strings and builds the property string to pass here?

patriknw (approved):
LGTM

PR #1622, opened 2023-04-17
bump: Downwards, support Scala 3.1 and later

johanandren:


build.sbt
```diff
@@ -10,7 +10,7 @@ val Nightly = sys.env.get("EVENT_NAME").contains("schedule")
 // align in release.yml
 val Scala213 = "2.13.10"
 val Scala212 = "2.12.17"
-val Scala3 = "3.2.2"
+val Scala3 = "3.1.3"
```
patriknw, line 13: Good, let's make that decision from akka/akka. For reference regarding the Scala 3 ecosystem approach to bumping https://github.com/akka/alpakka-kafka/pull/1603#discussion_r1117403493
Means that we at some point might have to follow them, even though I don't like it.

patriknw (approved):
LGTM

build.sbt
johanandren: Yeah, I already saw problems where upstream libs had bumped to 3.2, stopping us from coming along (Akka gRPC)

patriknw: ok, shall we make that call right away then? We just bump to latest Scala 3 in whatever Akka patch we are about to release.

Meaning that we can leave this unchanged, and bump in projections instead, then starting from akka/akka, but release whenever there are other reasons for releasing?

johanandren: Ok

Issue #1642, opened 2023-06-16
DrainingControl.isShutdown memory leak


ilinandrii:
### Versions used 
2.0.5
master

Akka version: 
2.5.32

### Expected Behavior
Calls to `DrainingControl.isShutdown` method don't create additional future callbacks that occupy memory and in a long run for infinite streams lead to OOM

### Actual Behavior
I'm using `DrainingControl.isShutdown().isCompleted()` for consumer health checks. 
Health check calls lead to corresponding `isShutdown` calls on all running consumers, which in turn create [new future callbacks](https://github.com/akka/alpakka-kafka/blob/8670cf9f2f767bad34e09ade7836a2e76b7f2243/core/src/main/scala/akka/kafka/scaladsl/Consumer.scala#L122) with `.flatMap`.  
The number of callbacks constantly growth and because all of my consumers are infinite streams those callbacks are never resolved and occupy heap until OOM is reached after a very long run.
I've workaround this by caching first call on `isShutdown` and calling `isCompleted` on returned future on each health check call, however it would be nice to have it fixed in a library.

ennru:
Thank you for reporting this. Recreating the future should not be necessary and must be avoided.
Would you be in a position to suggest a fix?

PR #1654, opened 2023-06-27
fix: only attach callbacks in DrainingControl.isShutdown once

leviramsey:
References #1642.

Also brings `javadsl.DrainingControl.isShutdown` behavior into line with `scaladsl` by not completing until both the source and overall stream have completed.

### DIFF
--- a/core/src/main/scala/akka/kafka/javadsl/Consumer.scala
+++ b/core/src/main/scala/akka/kafka/javadsl/Consumer.scala
@@ -92,11 +92,11 @@ object Consumer {
     def drainAndShutdown(ec: Executor): CompletionStage[T] =
       control.drainAndShutdown(streamCompletion, ec)
 
-    override def isShutdown: CompletionStage[Done] = _isShutdown
-
-    private[this] lazy val _isShutdown =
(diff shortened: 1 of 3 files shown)


core/src/main/scala/akka/kafka/javadsl/Consumer.scala
```diff
@@ -92,7 +92,12 @@ object Consumer {
     def drainAndShutdown(ec: Executor): CompletionStage[T] =
       control.drainAndShutdown(streamCompletion, ec)
 
-    override def isShutdown: CompletionStage[Done] = control.isShutdown
+    override def isShutdown: CompletionStage[Done] = _isShutdown
+
+    private[this] lazy val _isShutdown =
```
leviramsey, line 97: Not totally sure if this is needed for Java API bincompat or if we can make `isShutdown` itself lazy...

johanandren: I think you should skip the lazy and just make it a val on both sides, a single future or a future + completion stage alloc per DrainingControl does not seem like a problem.

johanandren (approved):
LGTM

PR #1717, opened 2024-03-02
chore: repository gardening

ennru:
- current GH actions
- fossa exclusions
- wording


johanandren (approved):
LGTM (but I wonder a bit how useful that gh release labeler categorization really is, what would seem useful to me is rather to pick up the conventional commit and list "New features" (all feat:), "Bugfixes" (all fix:), "Other improvements" (doc: and chore: and all other)

PR #1727, opened 2024-03-25
Add Scala 3 in tests

ennru:
Added Scala 3 even for tests and testing. Scala 3 artefacts were published even before.


SethTisue:
the CI failure:

```
[error] -- Error: /home/runner/work/alpakka-kafka/alpakka-kafka/tests/src/test/java/docs/javadsl/TestkitTestcontainersTest.java:47:0 
[error] 47 |
[error]    |^^^^^^^^^
[error]    |')' expected but eof found.
```

it's reproducible locally with `++3.3.3; tests/Test/compile`

it's the Scala 3 compiler that's complaining, not javac. but I tried `CompileOrder.JavaThenScala` and `CompileOrder.ScalaThenJava` but neither worked, so we have to make mixed compilation work (or restructure the sources, which we'd rather not have to do)

the problematic line seems to be:

```
@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```

if I comment that line out, compilation succeeds.

I don't see anything obviously relevant in https://github.com/scala/scala3/issues (but maybe I missed something)

not sure what to think yet. just writing down what I have so far. need to work on something else now — I'll come back to this but anyone else should feel free to take a look in the meantime

SethTisue:
> not sure what to think yet

well, I mean, it's clearly a bug in Scala 3's Java parser. it's just a matter of:

1. creating a self-contained, no-unnecessary-code-or-dependencies reproducer and filing it at https://github.com/scala/scala3/issues
2. trying to find a workaround we can use for the time being

SethTisue:
minimizes to:

```
% cat S.scala
class S
% cat J.java
//> using scala 3.4
//> using dep org.junit.jupiter:junit-jupiter-api:5.10.2
@TestInstance(org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS)
class J { }
% scala-cli compile .
Compiling project (Scala 3.4.0, JVM (17))
[error] ./J.java:6:1
[error] ')' expected but eof found.
```

SethTisue:
I minimized it and filed https://github.com/scala/scala3/issues/20026

I will still try to find a way to do without the external dependency (on junit-jupiter-api)

as for workarounds, I don't see any wiggle room, can't even think of something to try

johanandren:
Now I think it is some actual difference in behavior between Scala 2 and 3:

```
Logging finished for test [akka.kafka.internal.ConsumerProgressTrackingSpec: should lookup positions from the consumer when assigned] that [Failed(org.mockito.exceptions.misusing.WrongTypeOfReturnValue: 
Integer cannot be returned by position()
position() should return long
```

johanandren (approved):

PR #1728, opened 2024-04-03
feat: kip447 transactions

johanandren:
This uses the new APIs added for KIP447 to pass the right consumer group metadata along when committing the transaction. This should avoid the need to use any specific scheme for the transaction id other than making sure it is unique per consumer-producer stream (so could be a UUID for example)

### DIFF
--- a/core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
+++ b/core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
@@ -113,8 +113,8 @@ object Transactional {
 
   /**
    * Publish records to Kafka topics and then continue the flow. The flow can only be used with a [[Transactional.source]] that
-   * emits a [[ConsumerMessage.TransactionalMessage]]. The flow requires a unique `transactional.id` across all app
-   * instances.  The flow will override producer properties to enable Kafka exactly-once transactional support.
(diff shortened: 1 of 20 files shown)


core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -104,6 +116,18 @@ object Transactional {
    * emits a [[ConsumerMessage.TransactionalMessage]]. The flow requires a unique `transactional.id` across all app
    * instances.  The flow will override producer properties to enable Kafka exactly-once transactional support.
```
ennru, line 117:
```suggestion
   * emits a [[ConsumerMessage.TransactionalMessage]].
   * The flow will override producer properties to enable Kafka exactly-once transactional support.
```

ennru (approved):
Let's see if I get this correctly:
1. The source/flow auto-generates the `transactionalId`
1. When consuming, the `TransactionalMessageBuilder` injects `ConsumerGroupMetadata` with the offset
1. That metadata is used in `TransactionalProducerStage.commitTransaction`
1. Safeguards avoid pre-emptive stage shutdown

Other changes improve thread safety and make use of improved stuff in Scala and Akka.

LGTM.

PR #1733, opened 2024-04-11
feat: Add transaction prefix option

JustinPihony:
An add-on to #1728 for a bit more configurability

### DIFF
--- a/core/src/main/scala/akka/kafka/ProducerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ProducerSettings.scala
@@ -343,7 +343,6 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
     copy(eosCommitInterval = eosCommitInterval.asScala)
 
   /**
-   * Java API:
    * The prefix to append to the generated transaction id when using the `Transactional.sink` or `Transactional.flow`.
    */
   def withTransactionIdPrefix(transactionIdPrefix: String): ProducerSettings[K, V] =
--- a/core/src/main/scala/akka/kafka/ProducerSettings.scala
+++ b/core/src/main/scala/akka/kafka/ProducerSettings.scala
@@ -342,6 +342,10 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
   def withEosCommitInterval(eosCommitInterval: java.time.Duration): ProducerSettings[K, V] =
     copy(eosCommitInterval = eosCommitInterval.asScala)
 
+  /**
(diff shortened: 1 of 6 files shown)


johanandren:
Added some fixes, instead of waiting for a review cycle.

sebastian-alfers:
Should we think about adding a spec for this?

johanandren:
I don't think we can observe the transaction id in any reasonable way in a test without a lot of effort, so probably not worth it.

johanandren:
We can try running one of the existing test cases with a prefix specified though.

core/src/main/scala/akka/kafka/ProducerSettings.scala
```diff
@@ -436,7 +447,8 @@ class ProducerSettings[K, V] @InternalApi private[kafka] (
     s"dispatcher=$dispatcher," +
     s"eosCommitInterval=${eosCommitInterval.toCoarsest}," +
     s"enrichAsync=${enrichAsync.map(_ => "needs to be applied")}," +
-    s"producerFactorySync=${producerFactorySync.map(_ => "is defined").getOrElse("is undefined")})"
+    s"producerFactorySync=${producerFactorySync.map(_ => "is defined").getOrElse("is undefined")})" +
+    s"transactionIdPrefix=$transactionIdPrefix"
```
ennru, line 451:
```suggestion
    s",transactionIdPrefix=$transactionIdPrefix"
```

ennru (approved):
LGTM.

core/src/main/scala/akka/kafka/scaladsl/Transactional.scala
```diff
@@ -137,7 +138,7 @@ object Transactional {
 
     val flow = Flow
       .fromGraph(
-        new TransactionalProducerStage[K, V, ConsumerMessage.PartitionOffset](settings, settings.transactionIdPrefix + transactionalId)
+        new TransactionalProducerStage[K, V, ConsumerMessage.PartitionOffset](settings, transactionalId)
```
JustinPihony: I had considered adding it to the above method, but wouldn't it be better to keep it at the underlying one? Maybe you chose the top level since this one is deprecated? I had assumed it might be made private is all?

johanandren: the deprecated one expects the user to "design" their entire transaction id, so we shouldn't mess with it, they can (and do) already have whatever prefix they want

PR #1739, opened 2024-05-06
fix: push based consumer group metadata

johanandren:
When rebalance triggers draining, it blocks the consumer actor so that it could not reply to the request for consumer group metadata, this would often if not always cause the stream to crash on rebalance (new consumer added, consumer leaving).

### DIFF
--- a/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
+++ b/core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
@@ -186,22 +186,22 @@ private final class TransactionalProducerStageLogic[K, V, P](
     }
 
   private def commitTransaction(beginNewTransaction: Boolean = true,
-                                     abortEmptyTransactionOnComplete: Boolean = false): Unit = {
+                                abortEmptyTransactionOnComplete: Boolean = false): Unit = {
     val awaitingConf = awaitingConfirmationValue
     batchOffsets match {
       case batch: NonemptyTransactionBatch if awaitingConf == 0 =>
         try {
           log.debug("Committing transaction for transactional id '{}' consumer group '{}' with offsets: {}",
-            transactionalId,
-            latestSeenConsumerGroupMetadata,
-            batch.offsets)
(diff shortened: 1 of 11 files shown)


core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
```diff
@@ -117,27 +121,32 @@ private final class TransactionalProducerStageLogic[K, V, P](
...
   private var batchOffsets = TransactionBatch.empty
   private var demandSuspended = false
-  private var commitInProgress = false
+  private var latestSeenConsumerGroupMetadata: ConsumerGroupMetadata = null
 
   private var firstMessage: Option[Envelope[K, V, P]] = None
-
-  private val commitTransactionCB: Try[CommitTransaction] => Unit =
-    createAsyncCallback[Try[CommitTransaction]](commitTransaction _).invoke _
+  private var waitingForLastCommitAck = false
```
ennru, line 127: I can't see any change to this field.

ennru (commented):
Looks like a viable solution to have the group metadata at hand when needed.
The unused new field makes me wonder if you had a thought that didn't make it into the code.

core/src/main/scala/akka/kafka/internal/TransactionalProducerStage.scala
johanandren: Ah, leftover that didn’t end up being needed, I’ll remove

johanandren: Remove in [97ed384](https://github.com/akka/alpakka-kafka/pull/1739/commits/97ed38492f2f8504ad1dd81fa5b5c167cf381c01)

ennru (commented):
I notice the `CommitTransaction` class in `object TransactionalProducerStage` can now be removed.

ennru (approved):
Getting rid of the extra callback in the committing logic makes it much better.

LGTM.

PR #1757, opened 2024-07-05
chore: remove warnings filters

ennru:
The Scalac filters are no longer needed as Scala 2.12 has been dropped a while ago.


ennru:
Renames in GH workflows lets required builds show as missing, I will adapt once this is approved.

efgpinto (approved):

PR #1843, opened 2025-05-14
chore: bump kafka-clients to 4.0.0

sebastian-alfers:


build.sbt
```diff
@@ -22,8 +22,8 @@ val AkkaBinaryVersionForDocs = VersionNumber(akkaVersion).numbers match {
 }
 
 // Keep .scala-steward.conf pin in sync
-val kafkaVersion = "3.7.1"
-val KafkaVersionForDocs = "37"
+val kafkaVersion = "4.0.0"
+val KafkaVersionForDocs = "40"
```
johanandren, line 26: No need to bump test containers because client 4.0.0 also works with old brokers?

sebastian-alfers: test containers is still the version we have in `main`, isn't it?

sebastian-alfers: I think I misread your questions, which was more of a "if it should be bumped or not". I think not, since it works?

johanandren:
Let's hold this off if bumping to 3.9 is enough for now.

PR #1866, opened 2025-09-27
chore: tests with java 25

sebastian-alfers:


johanandren:
Some mockito upgrade needed to support running tests on JDK 25 here.

PR #1869, opened 2025-10-06
chore: Re-generate proto test message with newer protoc

johanandren:
Also, move proto file to where you'd expect it


ennru (approved):
LGTM.

